PR #1, opened 2025-07-25
Kicking off the SDK based on Symfony's MCP SDK

chr-hertel:
Kicking off the SDK with Symfony one to merge in features of PHP-MCP on top.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+/.git* export-ignore
+/examples export-ignore
+/tests export-ignore
+/.php-cs-fixer.dist.php export-ignore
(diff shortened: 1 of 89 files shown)


Nyholm:
Good. But do the initial code dump by pushing to main directly. This is to keep the git history.

chr-hertel:
We only have unsubstancial changes in that subtree split for now, see https://github.com/symfony/mcp-sdk/commits/main/

chr-hertel:
Lowered to PHP 8.1 for supporting all current PHP versions and Symfony 6.4

PR #2, opened 2025-08-01
Introducing schema component

chr-hertel:
Moving over [php-mcp/schema](https://github.com/php-mcp/schema) which is based on [2025-03-26/schema.ts](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-03-26/schema.ts)

Would still apply changes:
* using library specific exception classes instead of native base classes
* ~switching to `readonly` classes~
* adding phpstan types for array structures
* [still questionable] - do we need always need all five? `construct`, `make`, `toArray`, `fromArray` and `jsonSerialize`
  my2cents: the array structure is needed for `json_encode` and therefore the `jsonSerialize` that's the main use case for converting into array - and we have three different ways of constructing those value objects, why?
  my proposal would be to reduce the amount of methods per class to three:
  * `__construct(...)` for developers to instantiate
  * `make(array $data): static` for creation from array data
  * `jsonSerialize(): array` (this would reduce the need to array_map/toArray calls on subobjects since jsonSerialize works recursively)

WDYT @CodeWithKyrian - am I missing something?

**Still WIP**

### DIFF
--- a/src/Schema/JsonRpc/Parser.php
+++ b/src/Schema/JsonRpc/Parser.php
@@ -1,0 +1,80 @@
+<?php
+
+/*
+ * This file is part of the official PHP MCP SDK.
+ *
+ * A collaboration between Symfony and the PHP Foundation.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Mcp\Schema\JsonRpc;
+
+use Mcp\Exception\InvalidArgumentException;
+use Mcp\Schema\Constants;
+
+/**
+ * @author Kyrian Obikwelu <koshnawaza@gmail.com>
+ */
+class Parser
+{
+    /**
+     * Parses a raw JSON string into a JSON-RPC Message object (Request, Notification, Response, Error, or Batch variants).
+     *
+     * This method determines if the incoming message is a request-like message (Request, Notification, BatchRequest)
+     * or a response-like message (Response, Error, BatchResponse) based on the presence of 'method' vs 'result'/'error'.
+     *
+     * @param string $json the raw JSON string to parse
+     *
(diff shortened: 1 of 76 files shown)


CodeWithKyrian:
Yep, I agree with this direction. `toArray` is mostly just serving `jsonSerialize` so no point keeping both.

On `make(array $data)`, I’d lean toward sticking with `fromArray()` as the static factory for array data. `make()` works better as a convenient constructor alias for devs who prefer `Response::make(...)` over `new Response(...)`. Of course, with PHP 8.4 now supporting `new Response()->method()` without the extra parentheses, that pattern is slowly becoming less necessary. 

I’m fine with keeping `make()` if we see it as syntactic sugar, but I’m happy to drop it too if we want to slim things down.

chr-hertel:
When thinking about updates over time, I feel keeping in sync `__construct` and `make` is a bit annoying - also because we're only downstream to that official schema here.

If you give me a :+1: I would update the PR like this:
- [x] Lib specific exceptions
- [x] Drop `toArray` in favor of `jsonSerialize`
- [x] No changes to `fromArray`
- [x] Drop `make` in favor of slimming down
- [x] Add PHPStan array shapes

src/Schema/JsonRpc/Parser.php
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Mcp\Schema\JsonRpc;
+
+use Mcp\Exception\InvalidArgumentException;
+use Mcp\Schema\Constants;
+
+/**
+ * @author Kyrian Obikwelu <koshnawaza@gmail.com>
+ */
+class Parser
```
chr-hertel: we have this twice still, and also the method string is handled in multiple places, but that's to sort with #3

chr-hertel:
Got merged into main with 6025f890f913fe6538c0ce0593e1bfe5398af445

PR #3, opened 2025-08-03
Adopt Schema component in server handlers

chr-hertel:
~Okay, this was freaky easy.~ Celebrated too early.

Refactored the relation between `Notification` and `Request` base classes and their children
was quite implicit coupling but serving to concern - moved the schema upcasting (e.g. `fromNotification()`) and the method check to `Mcp\Message\Factory`

Reworked the relation between handler, schema and what will become the registries again. next step.


src/Schema/JsonRpc/Response.php
```diff
@@ -11,76 +11,53 @@
...
             $data['result']
         );
     }
+
+    public function jsonSerialize(): array
+    {
+        return [
+            'jsonrpc' => Constants::JSONRPC_VERSION,
+            'id' => $this->id,
+            'result' => \is_array($this->result) ? $this->result : $this->result->jsonSerialize(),
```
CodeWithKyrian: Since `jsonSerialize` is recursive and `Result` already implements JsonSerializable, this could be cleaner by simply returning `$this->result`. Just realized while going through : )

chr-hertel: yup, you're right - missed this one :+1:

chr-hertel:
there's another push incoming for replacing the VOs in `Mcp\Capability\`

CodeWithKyrian:
This is solid overall. 

I noticed the move to class-level readonly and was going to comment, but then I checked the minimum supported PHP version and it's 8.2 so that's fine. 

Curious though, is there any particular reason for going with 8.2 as the minimum and not, say... 8.1?

chr-hertel:
PHP 8.1 is still officially supported, that's true - it's only another three months or so and packagist.org claims it is only 12% right now. If you care, we can keep 8.1, but I didn't since this is quite new code :speak_no_evil:

CodeWithKyrian:
Fair enough 😅 I totally get that.

For me, I just like to support as many PHP setups as possible (especially within the same major version). But yeah, we could also decide to make this one of those "you-must-upgrade" packages and help push people off 8.1 entirely. So it’s really a matter of vibe - do we want to be accommodators or enforcers? 😂 (What's Symfony's in this case?)

Either way works for me tbh.

chr-hertel:
Well, Symfony has usually quite long support ranges with the LTS versions, but higher versions on recent versions:

* Legacy LTS v5.4 => Support until beginning of 2029 => PHP 7.2
* Current LTS v6.4 => Support until end of 2027 => PHP 8.1 
* Current Minor 7.3 => Support until beginning of 2026 => PHP 8.2

So, I'd say you have a good point here :laughing:
Let's keep Symfony 6.4 compatibility basically - can't do PHP 7.2 without crying

CodeWithKyrian:
Nahhh. 7.2 is off the table!!!!

chr-hertel:
Got merged into main with 09707b4798df239776f7415143e140ccab473aa4

PR #21, opened 2025-09-04
Add note about minimum stability to README

chr-hertel:
Like discussed, no tag for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,6 @@ Since this package has no tagged releases yet, it is required to extend your `co
     "prefer-stable": true,
 ```
 
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,6 +51,13 @@ Examples working
 composer require mcp/sdk
 ```
 
+Since this package has no tagged releases yet, it is required to extend your `composer.json`:
+```json
+    "minimum-stability": "dev",
+    "prefer-stable": true,
+```
+
+
```
chr-hertel:
```suggestion

```

PR #22, opened 2025-09-04
Fix JSON formatting

OskarStark:


README.md
```diff
@@ -66,8 +66,6 @@ This example demonstrates the most common usage pattern - a `stdio` server using
 Create `src/CalculatorElements.php`:
 
 ```php
-<?php
-
```
chr-hertel: this example is about the entire file, not only the class, so this is correct.

README.md
```diff
@@ -66,8 +66,6 @@ This example demonstrates the most common usage pattern - a `stdio` server using
 Create `src/CalculatorElements.php`:
 
 ```php
```
OskarStark:
```suggestion
```php
<?php

```

PR #24, opened 2025-09-04
[CI] Update `actions/checkout` from v4 to v5

OskarStark:


chr-hertel:
Looks like your commits here also lack some signing @OskarStark

OskarStark:
Fixed, for all PRs @chr-hertel

PR #33, opened 2025-09-04
Remove unused development dependencies

OskarStark:
Remove symfony/dependency-injection and symfony/event-dispatcher from require-dev as they are not used in the codebase. The project uses psr/event-dispatcher interface instead of Symfony's implementation.


chr-hertel (approved):
Planned to use them while reviving the examples, but anyhow.

PR #34, opened 2025-09-04
Sort `suggest` section in `composer.json` alphabetically

OskarStark:


chr-hertel (changes requested):
They can actually be removed instead

OskarStark:
On it...

OskarStark:
Ready from my side

chr-hertel (approved):

PR #36, opened 2025-09-05
feat: implement variadic parameter handling in ReferenceHandler

monayemislam:
- Add support for variadic parameters (...) in prepareArguments method
- Collect all numeric-indexed arguments starting from parameter position
- Add proper error handling for variadic parameter processing
- Remove TODO comment and complete the implementation
- Maintain backward compatibility with existing parameter handling

Fixes the TODO item in ReferenceHandler.php line 88 and enables proper handling of functions with variadic parameters in MCP tools and resources.

### DIFF
--- a/src/Capability/Registry/ReferenceHandler.php
+++ b/src/Capability/Registry/ReferenceHandler.php
@@ -85,9 +85,27 @@ class ReferenceHandler
         $finalArgs = [];
 
         foreach ($reflection->getParameters() as $parameter) {
-            // TODO: Handle variadic parameters.
             $paramName = $parameter->getName();
             $paramPosition = $parameter->getPosition();
+
+            // Handle variadic parameters
+            if ($parameter->isVariadic()) {
+                // For variadic parameters, collect all remaining arguments
+                $variadicArgs = [];
+                foreach ($arguments as $key => $value) {
+                    if (is_numeric($key) && $key >= $paramPosition) {
+                        try {
+                            $variadicArgs[] = $this->castArgumentType($value, $parameter);
+                        } catch (InvalidArgumentException $e) {
+                            throw RegistryException::invalidParams($e->getMessage(), $e);
+                        } catch (\Throwable $e) {
+                            throw RegistryException::internalError("Error processing variadic parameter `{$paramName}`: {$e->getMessage()}", $e);
+                        }
+                    }
+                }
+                $finalArgs[$paramPosition] = $variadicArgs;
+                continue;
+            }
 
             if (isset($arguments[$paramName])) {
                 $argument = $arguments[$paramName];


src/Capability/Registry/ReferenceHandler.php
```diff
@@ -85,10 +85,28 @@ private function prepareArguments(\ReflectionFunctionAbstract $reflection, array
         $finalArgs = [];
 
         foreach ($reflection->getParameters() as $parameter) {
-            // TODO: Handle variadic parameters.
             $paramName = $parameter->getName();
             $paramPosition = $parameter->getPosition();
 
+            // Handle variadic parameters
```
chr-hertel, line 91: i think this is clear with the next line of code already and we can remove this comment
```suggestion
```

src/Capability/Registry/ReferenceHandler.php
```diff
@@ -85,10 +85,28 @@ private function prepareArguments(\ReflectionFunctionAbstract $reflection, array
...
+            if ($parameter->isVariadic()) {
+                // For variadic parameters, collect all remaining arguments
+                $variadicArgs = [];
+                foreach ($arguments as $key => $value) {
+                    if (is_numeric($key) && $key >= $paramPosition) {
```
chr-hertel, line 96: this reads like `array_slice` could be of help :+1:

src/Capability/Registry/ReferenceHandler.php
```diff
@@ -85,10 +85,28 @@ private function prepareArguments(\ReflectionFunctionAbstract $reflection, array
...
+                        try {
+                            $variadicArgs[] = $this->castArgumentType($value, $parameter);
+                        } catch (InvalidArgumentException $e) {
+                            throw RegistryException::invalidParams($e->getMessage(), $e);
+                        } catch (\Throwable $e) {
+                            throw RegistryException::internalError("Error processing variadic parameter `{$paramName}`: {$e->getMessage()}", $e);
+                        }
```
chr-hertel, line 103: this is basically the same like in line 112-118 - does it make sense to shift the `throw RegistryException` into the `castArgumentType` method instead and skip the `try-catch` in both cases?

or somehow merge this block into lower if-else part as well?

chr-hertel (changes requested):
Hey @monayemislam, good catch that todo comment - thanks for picking up on that!

I made some smaller comments, but can you share how you tested this?

chr-hertel:
Hi @monayemislam, do you want to continue here?

src/Capability/Registry/ReferenceHandler.php
```diff
@@ -85,10 +85,28 @@ private function prepareArguments(\ReflectionFunctionAbstract $reflection, array
...
+                    }
+                }
+                $finalArgs[$paramPosition] = $variadicArgs;
+                continue;
+            }
```
soyuka, line 108:
```suggestion
            if ($parameter->isVariadic()) {
                return $arguments;
            }
```

This is sufficient for a use case where the handler has `...$arguments`:

```
                $definition['handler'] = fn (...$arguments) => ($this->operationHandler)($mcpName, $arguments);
```

Could you enlighten me with your use case where you gather the type of the variadic argument? What's the use case?

chr-hertel:
Closing here for now - please feel free to reopen and continue the work if it still matters to you.

PR #42, opened 2025-09-06
PoC - Support server notifications

### LINKED ISSUE #12: [Server] Emit Notification on Capability Change
The specification lists server emitting notifications, whenever a list of capabilities for the client changes:

For this we would need support from dispatching events in the Registry to creating notifications for
- [ ] ResourceListChangedNotification
- [ ] PromptListChangedNotification
- [ ] ToolListChangedNotification

(There are more server notification, but less connected to this scenario.)

---

Please announce yourself before working on this - so we can avoid duplicate effort.

Bellangelo:
Based on #12, the server needs to notify the client about different events. To make this work, I have created a `NotificationPublisher` which listens on the events that we dispatch and stores them until the `Server` utilizes them.
Since the `NotificationPublisher` needs to listen on the events, I have installed Symfony Dispatcher for internal usage.

## Motivation and Context

Closes #12

## How Has This Been Tested?
-

## Breaking Changes
-

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally


Bellangelo:
Hi @chr-hertel, I created a PoC of a solution. There are a few things that do not work 100% correct but I wanted a quick validation around the idea. I had to implement a few new things so I wasn't sure if I understood correctly the issue.

chr-hertel:
Nice one @Bellangelo - thanks for working on this! :)

So, i just gave it a quick read, maybe i missed something. so using an event dispatcher in the registry was an idea to publish events, but i wonder if we'd go with your `NotificationPublisher` idea - and i like it btw :+1: - if we'd need that dispatcher at all.

with the publisher there is a central stack of notifications that get's emptied by the server - who else would listen to those events? and if there's only one consumer of events why not push from the emitting service (registry) to the publisher directly?

Bellangelo:
> with the publisher there is a central stack of notifications that get's emptied by the server - who else would listen to those events? and if there's only one consumer of events why not push from the emitting service (registry) to the publisher directly?

Good question @chr-hertel, my initial thought for the PSR event dispatcher was that the developer building the server might also want to listen to these events. That’s why I’m dispatching through the developer’s event dispatcher here:
https://github.com/modelcontextprotocol/php-sdk/pull/42/files#diff-1758bd2c4379f6f88e9048d7b437c38d17a64e3d71dd2e9493322c7e61d4a87fR224-R229

If my understanding is off, then you’re right and we could call the `NotificationPublisher` directly instead of going through the dispatcher. The trade-off is that this would couple components that are otherwise unrelated. However, it might be unnecessary complexity at this stage of the project.

chr-hertel:
yup, valid point, server application code might listen to that as well - but on the other hand i'd expect that the impulse for change is rather coming from the application side anyways, since the server itself doesn't have inner logic to change the lists - as of now.
currently i'm more leaning to keeping it simple for now - like having the direct coupling and removing the EventDispatcher for now, yea

huangdijia:
Transitionally depending on symfony is not a good thing; it lacks flexibility

Bellangelo:
hi @chr-hertel, it should be ready now.

butschster:
Hi @Bellangelo 👋

I noticed that in addition to the `MessageFactory` instantiation https://github.com/Bellangelo/php-sdk/blob/support-server-notifications/src/JsonRpc/Handler.php#L58, the `NotificationPublisher` class also has a constructor that takes just one argument. I was wondering why you decided to introduce the extra static `make()` method for creating it, instead of relying directly on the constructor. (https://github.com/modelcontextprotocol/php-sdk/pull/42/files#diff-1758bd2c4379f6f88e9048d7b437c38d17a64e3d71dd2e9493322c7e61d4a87fR211)

Is there a specific reason or design goal behind this approach? I’d love to better understand the intention here. 🙂

Bellangelo:
> Hi @Bellangelo 👋
> 
> Is there a specific reason or design goal behind this approach? I’d love to better understand the intention here. 🙂

Hey @butschster, I did this to minimize how many things the dependents know about their dependencies. If the `NotificationPublisher` grows more, you won't have to change the `ServerBuilder`.

src/Capability/Registry.php
```diff
@@ -102,7 +101,7 @@ public function registerTool(Tool $tool, callable|array|string $handler, bool $i
 
         $this->tools[$toolName] = new ToolReference($tool, $handler, $isManual);
 
-        $this->eventDispatcher?->dispatch(new ToolListChangedEvent());
+        $this->notificationPublisher->enqueue(ToolListChangedNotification::class);
```
chr-hertel: wouldn't it be a bit easier to just go with
```suggestion
        $this->notificationPublisher->enqueue(new ToolListChangedNotification());
```

Bellangelo: @chr-hertel Hmm.. I thought that they would need to pass through the `MessageFactory`. If they don't need to pass, should I also remove these notifications from the `MessageFactory`? Because they won't be send as a response to a prompt.

Bellangelo: `MessageFactory` removed from the `NotificationPublisher` in this commit https://github.com/modelcontextprotocol/php-sdk/pull/42/commits/3019972eaf56ada49753394e13d2302d16209f1c

src/JsonRpc/MessageFactory.php
```diff
@@ -106,7 +111,7 @@ public function create(string $input): iterable
     /**
      * @return class-string<HasMethodInterface>
      */
-    private function getType(string $method): string
+    private function getType(string $method/**/): string
```
chr-hertel:
```suggestion
    private function getType(string $method): string
```

src/Server/NotificationPublisher.php
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * @return \Generator<int, Notification, void, void>
+     */
+    public function flush(): iterable
+    {
+        $out = $this->queue;
+        $this->queue = [];
+
+        yield from $out;
```
chr-hertel: should work as well
```suggestion
        yield from $this->queue;

        $this->queue = [];
```

src/Capability/Registry.php
```diff
@@ -61,8 +60,8 @@ class Registry
     private array $resourceTemplates = [];
 
     public function __construct(
+        private readonly NotificationPublisher $notificationPublisher,
```
chr-hertel:
```suggestion
        private readonly NotificationPublisher $notificationPublisher = new NotificationPublisher(),
```

src/Capability/Registry.php
chr-hertel: the `MessageFactory` is intended to handle incoming json/array payload - in this case it's outgoing

src/JsonRpc/MessageFactory.php
Bellangelo: Fixed in https://github.com/modelcontextprotocol/php-sdk/pull/42/commits/759a379ec02f9ee475bcc23ae6d939552399137c

src/Server/NotificationPublisher.php
Bellangelo: Fixed in https://github.com/modelcontextprotocol/php-sdk/pull/42/commits/f226b2b82186dc57e635596e7868fe0099079a8a

src/Capability/Registry.php
Bellangelo: Fixed in https://github.com/modelcontextprotocol/php-sdk/pull/42/commits/4c2415b7d4a916f41a5e8245db667a78fe1ea649

src/Capability/Registry.php
Bellangelo: I know but it seems that it contains outgoing notifications https://github.com/modelcontextprotocol/php-sdk/blob/main/src/JsonRpc/MessageFactory.php#L39. So we should remove these from the `MessageFactory`, right?

chr-hertel (approved):
Ready for merging from my point of view - thanks @Bellangelo!

@CodeWithKyrian any thoughts? :)

chr-hertel:
Needs a rebase to solve the conflicts

Bellangelo:
> Needs a rebase to solve the conflicts

@chr-hertel Done.

CodeWithKyrian:
@Bellangelo please rebase and adjust to fit the new architecture.

I’m also wondering if there’s a possibility of using the session as a message queue, since it can be persistent (depending on the store) and is unique to each client connection. Just rough thoughts at this point.

@chr-hertel, any thoughts?

chr-hertel:
@CodeWithKyrian that would be even required for making this work on the http transport, right?
@Bellangelo can you have a look into this? changed with merging #49

chr-hertel:
@Bellangelo still up for finishing here?

chr-hertel:
Closing here for now, we merged quite a bunch of changes - if you still want to continue - feel free to reopen & rebase.
Cheers!

PR #55, opened 2025-09-09
BugFix - Pagination of List requests

bigdevlarry:
Implements MCP pagination for tools/list, prompts/list, and resources/list endpoints.

- Added pagination support to ListToolsHandler, ListPromptsHandler, ListResourcesHandler
- Enhanced Registry with pagination methods and cursor calculation
- Added comprehensive unit tests for all pagination scenarios
- Added proper exception handling with @throws annotations

## Motivation and Context

The change was raised as an issue to enable the list request to have pagination support.

## How Has This Been Tested?

Yes, this has been tested. 

## Breaking Changes

No, users don't need to make any updates

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/Capability/Registry.php
+++ b/src/Capability/Registry.php
@@ -22,6 +22,7 @@ use Mcp\Event\ResourceListChangedEvent;
 use Mcp\Event\ResourceTemplateListChangedEvent;
 use Mcp\Event\ToolListChangedEvent;
 use Mcp\Exception\InvalidArgumentException;
+use Mcp\Exception\InvalidCursorException;
 use Mcp\Schema\Content\PromptMessage;
 use Mcp\Schema\Content\ResourceContents;
 use Mcp\Schema\Prompt;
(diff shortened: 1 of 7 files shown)


bigdevlarry:
Hi @chr-hertel,I have opened a PR to address the issue related to pagination.

chr-hertel:
Thanks @bigdevlarry for jumping on that :+1: 

I think this should wait for #46 as well - you could basically use that as a base branch already

chr-hertel:
#46 got merged, needs a rebase

src/Server/RequestHandler/ListPromptsHandler.php
```diff
@@ -34,13 +35,22 @@ public function supports(HasMethodInterface $message): bool
...
+        $allPrompts = $this->registry->getPrompts();
+
         $prompts = $this->registry->getPrompts($this->pageSize, $message->cursor);
-        $nextCursor = (null !== $cursor && \count($prompts) === $this->pageSize) ? $cursor : null;
+
+        $nextCursor = $this->registry->calculateNextCursor(
+            $allPrompts,
+            $message->cursor,
+            \count($prompts)
+        );
```
chr-hertel: we're only getting `$allPrompts` from the registry to hand it over to the registry again, right?
i guess the API would be slimmer if the registry would handle that internally, and we'd only need to go with:
```php
$nextCursor = $this->registry->calculatePromptCursor($message->cursor);
```

chr-hertel: or even better, we could return a more sophisticated DTO from the `getPrompts(...): ReferencePage`

```php
final class ReferencePage
{
    public function __construct(
        public readonly array $references,
        public readonly ?string $nextCursor,
    ) {
    }
}
```

WDYT?

src/Capability/Registry.php
```diff
@@ -238,30 +241,173 @@ public function getResourceTemplate(string $uriTemplate): ?ResourceTemplateRefer
...
+
+        if (null === $reference) {
+            throw new InvalidArgumentException(\sprintf('Prompt "%s" is not registered.', $name));
+        }
+
+        return $reference->formatResult(
+            $this->referenceHandler->handle($reference, $arguments)
+        );
+    }
+
```
chr-hertel: this is a mix up due to the rebase maybe?

src/Capability/Registry.php
```diff
@@ -238,30 +241,173 @@ public function getResourceTemplate(string $uriTemplate): ?ResourceTemplateRefer
...
 
-    public function getResourceTemplates(): array
+    /**
+     * Calculate next cursor for pagination.
+     *
+     * @param list<mixed> $allItems      The complete array of items
+     * @param string|null $currentCursor Current cursor position
+     * @param int         $returnedCount Number of items actually returned
+     */
+    public function calculateNextCursor(array $allItems, ?string $currentCursor, int $returnedCount): ?string
```
chr-hertel: if we're only doing a `count()` on `$allItems` it should be the count as `int` parameter right away

chr-hertel (changes requested):

src/Capability/Registry.php
bigdevlarry: Ah ! I see what you're saying. Will just count within the getTools and parse. Make sense

bigdevlarry:
@chr-hertel this is back up for review again

src/Server/RequestHandler/Reference/Page.php
```diff
@@ -0,0 +1,52 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Mcp\Server\RequestHandler\Reference;
+
+/**
+ * @implements \ArrayAccess<int|string, mixed>
+ */
+final class Page implements \Countable, \ArrayAccess
```
chr-hertel, line 17: what do you think about extending `ArrayObject` instead? would maybe ease some stuff

chr-hertel (changes requested):
I was testing this with adding more tools to example no1 and added the pagination limit on the builder - this still needs to be taken into account here - or do we better split into a follow-up PR?

anyhow, this needs a rebase after #49 and #46 

Thanks already @bigdevlarry!

chr-hertel:
Oh, and can you please make sure to squash this into a single, signed commit please

bigdevlarry:
@chr-hertel I think we should split it into another PR once this one gets merged.

PR #72, opened 2025-09-17
[Server] Handle empty message queue in CachePoolStore::pop()

MartkCz:
Previously, pop() would call array_shift() even if the cached queue was empty,
leading to unnecessary operations (caching an empty array repeatedly). This change ensures that pop() returns null
early when no messages are available.


chr-hertel (approved):
Thanks @MartkCz!

chr-hertel:
Not needed anymore after #49 which includes the removal of SSE transport.

PR #73, opened 2025-09-17
Add ToolExecutionExceptionInterface for custom tool error handling

MartkCz:
## Motivation and Context
The current tool execution system lacks a standardized way for handling tool-specific errors. This change provides a consistent way for tools to report errors.

## How Has This Been Tested?
Added unit tests in ToolCallerTest.php

## Breaking Changes
No breaking changes.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
The implementation includes modifications to:
- ToolCaller and ToolCallerInterface for exception handling integration
- ToolChain for proper error propagation
- New ToolExecutionExceptionInterface defining the contract for custom tool error handling

## Usage
```php
class McpElementsException extends RuntimException implements ToolExecutionExceptionInterface { ... }

class McpElements
{
	#[McpTool(name: 'calculate')]
	public function calculate(float $a, float $b, string $operation): float|string
	{
		$op = strtolower($operation);

		switch ($op) {
			case 'add':
				$result = $a + $b;
				break;
			case 'subtract':
				$result = $a - $b;
				break;
			case 'multiply':
				$result = $a * $b;
				break;
			case 'divide':
				if (0 == $b) {
					throw new McpElementsException(['Division by zero.']);
				}
				$result = $a / $b;
				break;
			default:
				throw new McpElementsException(["Unknown operation '{$operation}'. Supported: add, subtract, multiply, divide."]);
		}

		if (!$this->config['allow_negative'] && $result < 0) {
			throw new McpElementsException(['Negative results are disabled.']);
		}

		return round($result, $this->config['precision']);
	}
}
```

### DIFF
--- a/src/Capability/Registry/ReferenceHandlerInterface.php
+++ b/src/Capability/Registry/ReferenceHandlerInterface.php
@@ -27,8 +27,9 @@ interface ReferenceHandlerInterface
      *
      * @return mixed the result of the element execution
      *
-     * @throws \Mcp\Exception\InvalidArgumentException if the handler is invalid
-     * @throws \Mcp\Exception\RegistryException        if execution fails
+     * @throws \Mcp\Exception\InvalidArgumentException        if the handler is invalid
+     * @throws \Mcp\Exception\ToolExecutionExceptionInterface if the tool reports an error during its execution
+     * @throws \Mcp\Exception\RegistryException               if execution fails
      */
     public function handle(ElementReference $reference, array $arguments): mixed;
 }
--- a/src/Capability/Tool/ToolCaller.php
+++ b/src/Capability/Tool/ToolCaller.php
@@ -14,6 +14,7 @@ namespace Mcp\Capability\Tool;
 use Mcp\Capability\Registry\ReferenceHandlerInterface;
 use Mcp\Capability\Registry\ReferenceProviderInterface;
 use Mcp\Exception\ToolCallException;
+use Mcp\Exception\ToolExecutionExceptionInterface;
 use Mcp\Exception\ToolNotFoundException;
 use Mcp\Schema\Content\AudioContent;
 use Mcp\Schema\Content\EmbeddedResource;
@@ -68,6 +69,11 @@ final class ToolCaller implements ToolCallerInterface
             ]);
 
             return new CallToolResult($formattedResult);
+        } catch (ToolExecutionExceptionInterface $e) {
+            return CallToolResult::error(array_map(
+                fn (string $message): TextContent => new TextContent($message),
+                $e->getErrorMessages(),
+            ));
         } catch (\Throwable $e) {
             $this->logger->error('Tool execution failed', [
                 'name' => $toolName,
--- a/src/Capability/Tool/ToolCallerInterface.php
+++ b/src/Capability/Tool/ToolCallerInterface.php
@@ -12,6 +12,7 @@
 namespace Mcp\Capability\Tool;
 
 use Mcp\Exception\ToolCallException;
+use Mcp\Exception\ToolExecutionExceptionInterface;
 use Mcp\Exception\ToolNotFoundException;
 use Mcp\Schema\Request\CallToolRequest;
 use Mcp\Schema\Result\CallToolResult;
@@ -22,8 +23,9 @@ use Mcp\Schema\Result\CallToolResult;
 interface ToolCallerInterface
 {
     /**
-     * @throws ToolCallException     if the tool execution fails
-     * @throws ToolNotFoundException if the tool is not found
+     * @throws ToolCallException               if the tool execution fails
+     * @throws ToolNotFoundException           if the tool is not found
+     * @throws ToolExecutionExceptionInterface if the tool reports an error during its execution
      */
     public function call(CallToolRequest $request): CallToolResult;
 }
--- a/src/Capability/ToolChain.php
+++ b/src/Capability/ToolChain.php
@@ -17,6 +17,7 @@ use Mcp\Capability\Tool\MetadataInterface;
 use Mcp\Capability\Tool\ToolCallerInterface;
 use Mcp\Exception\InvalidCursorException;
 use Mcp\Exception\ToolCallException;
+use Mcp\Exception\ToolExecutionExceptionInterface;
 use Mcp\Exception\ToolNotFoundException;
 use Mcp\Schema\Request\CallToolRequest;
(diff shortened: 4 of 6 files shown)


chr-hertel:
Hi @MartkCz,
I think you're right, we should at least document how users should deal with that kind of situation, but I disagree on the naming here.

The `ReferenceHandlerInterface` is rather generic and does/should not care/know about the concrete element implementation of tools or prompts, etc. In your example you're implementing the exception even in a generic way again with an `McpElementsException` - a bit confusing tbh.

Currently the `ReferenceHandlerInterface` states already, that you should throw a `RegistryException` "if execution fails". So the use case is covered, only the exception is not tailored.

I would prefer, instead of creating a new interface to implement, you create a `ReferenceExecutionException extends RegistryException` or `ElementsExecutionException` or similar - which is more generic than tool specific - and that new one replacing the `RegistryException` on the interface. WDYT?

MartkCz:
@chr-hertel Thanks for your time.

1. I removed duplicate logging in the CallToolHandler (all logging is already here: https://github.com/modelcontextprotocol/php-sdk/blob/main/src/Capability/Tool/ToolCaller.php#L51-L76
).
2. I fixed https://github.com/modelcontextprotocol/php-sdk/blob/main/tests/Server/RequestHandler/CallToolHandlerTest.php#L117-L145
 — when a tool is not found, the code should not be -32603 (INTERNAL_ERROR), but -32602 (INVALID_PARAMS), see https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling
.
 3. I think it creates a mess because RegistryException and ReferenceExecutionException must be wrapped into ToolCallException and then unwrapped to determine whether it’s a protocol error or an execution error, since each is handled differently. On top of that, there’s a separate protocol error ToolNotFoundException. In my opinion, the best approach would be to split them into protocol and execution exceptions and throw those instead of ToolCallException and ToolNotFoundException.
 
 ```php
interface ReferenceExceptionInterface {}

final class ExecutionReferenceException extends \RuntimException implements ReferenceExceptionInterface {} // business errors

abstract class ProtocolReferenceException extends \LogicException implements ReferenceExceptionInterface {} // unknown tools, invalid arguments, server errors

final class ToolNotFoundReferenceException extends ProtocolReferenceException {}

final class InvalidParamsReferenceException extends ProtocolReferenceException {}

final class InternalErrorReferenceException extends ProtocolReferenceException {}
```
 
 This is just a proposal — there wouldn’t necessarily have to be so many exception classes inheriting from ProtocolReferenceException, but I wanted to preserve throwing ToolNotFoundException. The naming is also not final.
 
 and then this is simplified from this:
 
 ```php
try {
    $result = $this->referenceHandler->handle($toolReference, $arguments);
    /** @var TextContent[]|ImageContent[]|EmbeddedResource[]|AudioContent[] $formattedResult */
    $formattedResult = $toolReference->formatResult($result);

        'name' => $toolName,
    ]);

    return new CallToolResult($formattedResult);
} catch (RegistryException $e) {
    throw new ToolCallException($request, $e);
} catch (\Throwable $e) {
    if ($e instanceof ToolCallException) {
        throw $e;
    }

    $this->logger->error('Tool execution failed', [
        'name' => $toolName,
        'exception' => $e->getMessage(),
        'trace' => $e->getTraceAsString(),
    ]);

    throw new ToolCallException($request, RegistryException::internalError('Error while executing tool', $e));
}
```

to this:

```php
try {
	$result = $this->referenceHandler->handle($toolReference, $arguments);
	/** @var TextContent[]|ImageContent[]|EmbeddedResource[]|AudioContent[] $formattedResult */
	$formattedResult = $toolReference->formatResult($result);

		'name' => $toolName,
	]);

	return new CallToolResult($formattedResult);
} catch (ReferenceExceptionInterface $e) {
	throw $e;
} catch (\Throwable $e) {
	$this->logger->error('Tool execution failed', [
		'name' => $toolName,
		'exception' => $e->getMessage(),
		'trace' => $e->getTraceAsString(),
	]);

	throw new InternalErrorReferenceException($request, $e);
}
```

and this:
```php
try {
    $content = $this->toolCaller->call($message);
} catch (ToolNotFoundException $exception) {
    return Error::forInvalidParams($exception->getMessage(), $message->getId());
} catch (ToolCallException $exception) {
    $registryException = $exception->registryException;

    if ($registryException instanceof ReferenceExecutionException) {
        return new Response($message->getId(), CallToolResult::error(array_map(
            fn (string $message): TextContent => new TextContent($message),
            $registryException->messages,
        )));
    }

		return new Error($message->getId(), $registryException->getCode(), $registryException->getMessage());
}
```

to this:
```php
try {
	$content = $this->toolCaller->call($message);
} catch (ProtocolReferenceException $exception) {
	return new Error($message->getId(), $exception->getCode(), $exception->getMessage());
} catch (ExecutionReferenceException $exception) {
	return new Response($message->getId(), CallToolResult::error(array_map(
		fn (string $message): TextContent => new TextContent($message),
		$exception->messages,
	)));
}
```

chr-hertel:
@MartkCz Thanks for the follow-up and the detailed explanation! :pray: 
I think your proposal sounds & looks reasonable to me :+1: 

btw, this needs a rebase - and please make sure to have a squashed, signed commit please

chr-hertel:
Hi @MartkCz - do you see your concerns also addressed with #124?

MartkCz:
Hi @chr-hertel , sorry, but I didn't have time to finish the PR.

At a quick glance it looks like a sufficient solution, the only question is whether it's really necessary to log errors from ToolCallException. In my experience, they were never needed, and in the worst case you can log directly in the McpTool method.

If I take an example where the AI/human calls divideNumbers with $b as 0 and McpTool responds that you can't divide by zero, I would expect it to handle that and provide a different number or ask for a different one. Otherwise, I would end up with Sentry full of logs that I can’t and don’t want to deal with, because it’s not an application error.

Logging is probably fine during debugging, but I would keep it at most at debug/info severity.

Thanks for mention.

chr-hertel:
Thanks @MartkCz for the fast feedack - referring to this one, right?
https://github.com/modelcontextprotocol/php-sdk/blob/7a3b473c405411aca8be14c0c6e12d2493752d61/src/Server/Handler/Request/CallToolHandler.php#L77-L85

MartkCz:
Yes, that's the right one

chr-hertel:
Opened #157 as follow up - closing here for now - thanks for raising the topic!

PR #75, opened 2025-09-20
Add getRegistry() method to Server class for debugging access

### LINKED ISSUE #74: [Server] Expose Server Registry for debugging and profiling tools
Hello! I'm working on a Symfony Web Profiler integration for MCP servers and need access to the discovered capabilities for debugging purposes.

## Current Issue

  1. ServerBuilder::build() creates a Registry and populates it with discovered capabilities
  2. This Registry is passed to Handler::make() via interfaces (ReferenceRegistryInterface, ReferenceProviderInterface)
  3. Request handlers like ListToolsHandler can access capabilities via $registry->getTools()
  4. But the Registry is never exposed publicly from the Server class

  The Registry class has all the needed methods:
  - getTools(): array
  - getPrompts(): array
  - getResources(): array
  - getResourceTemplates(): array

  But they're only accessible internally through the MCP protocol handlers.

## Proposed Solution
Add a method to expose the registry

camilleislasse:
## Summary
Expose the Registry through the Server class to enable debugging and monitoring tools like Symfony Web Profiler to access discovered capabilities.

## Changes
- Add `ReferenceProviderInterface $registry` parameter to `Server` constructor
- Add `getRegistry(): ReferenceProviderInterface` method to `Server` class  
- Update `ServerBuilder` to pass registry to `Server` constructor
- Update standalone example to use new `Server` constructor signature
- Add test coverage for `getRegistry()` method

## Use Case
This enables integration with debugging tools like Symfony Web Profiler by providing access to:
- `getTools()` - All registered tools
- `getResources()` - All registered resources  
- `getPrompts()` - All registered prompts
- `getResourceTemplates()` - All registered resource templates

## Test Plan
- [x] All existing tests pass
- [x] New test added for `getRegistry()` method
- [x] PHP CS Fixer validation passes
- [x] PHPStan static analysis passes
- [x] Manual testing with various MCP capabilities confirmed working

Resolves #74

### DIFF
--- a/examples/09-standalone-cli/index.php
+++ b/examples/09-standalone-cli/index.php
@@ -13,12 +13,16 @@ require __DIR__.'/vendor/autoload.php';
 
 use Symfony\Component\Console as SymfonyConsole;
 use Symfony\Component\Console\Output\OutputInterface;
+use Mcp\Capability\Registry;
 
 $debug = (bool) ($_SERVER['DEBUG'] ?? false);
 
 // Setup input, output and logger
 $output = new SymfonyConsole\Output\ConsoleOutput($debug ? OutputInterface::VERBOSITY_VERY_VERBOSE : OutputInterface::VERBOSITY_NORMAL);
 $logger = new SymfonyConsole\Logger\ConsoleLogger($output);
+
+// Configure the registry (empty for this standalone example)
+$registry = new Registry();
 
 // Configure the JsonRpcHandler and build the functionality
 $jsonRpcHandler = new Mcp\JsonRpc\Handler(
@@ -28,7 +32,7 @@ $jsonRpcHandler = new Mcp\JsonRpc\Handler(
 );
 
 // Set up the server
-$sever = new Mcp\Server($jsonRpcHandler, $logger);
+$sever = new Mcp\Server($jsonRpcHandler, $registry, $logger);
 
 // Create the transport layer using Stdio
 $transport = new Mcp\Server\Transport\StdioTransport(logger: $logger);
--- a/src/Server.php
+++ b/src/Server.php
@@ -11,6 +11,7 @@
 
 namespace Mcp;
 
+use Mcp\Capability\Registry\ReferenceProviderInterface;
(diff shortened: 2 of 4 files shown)


chr-hertel:
If we make services injectable, it shouldn't be needed to use getters for them since the runtime (or in your case the Symfony DIC) should know them already. Having getters for service dependencies on services leads to misusage for using them to access that service instead of having the explicit dependency. and if the use case is only the profiler toolbar, i'm against adding it. if it is too hard to get the registry instance in a different way, we usually use `Traceable*` decorator to gather insights for the profiler.

soyuka:
This makes quite hard to reuse the code of the registry (in #108 I'd like to use my own registry but probably that #108 need a proper fix instead of trying to hack around it).  Can we have a setter maybe?

soyuka:
My PR was @camilleislasse could you check if https://github.com/modelcontextprotocol/php-sdk/pull/111 helps with your use case? We may close this then.

camilleislasse:
Thanks for the feedback @soyuka and @chr-hertel!

  I checked #111 and I see it adds addLoaders(), but I'm not sure how to use it properly for my use case.

  @chr-hertel I understand the concerns about getters and using Traceable decorators instead. However, I'm struggling to see how to apply that pattern here since the Registry seems to only exist inside Builder::build() as a local variable and isn't passed to the Server.

 Looking at the code after #111, I could technically create a custom Loader that captures the Registry reference when load() is called, but that feels wrong - like I'd be misusing the Loader mechanism just to get access to something.

  I might be missing something obvious though. What's the recommended way to access the Registry for profiling purposes? Should I:

  1. Create a custom "capture" loader (feels hacky to me)
  2. Use a different approach I'm not seeing?
  3. Is there a plan to expose the Registry in a future update?

soyuka:
Loaders give you the ability to fill the registry, I'm not sure why you need access to the Registry? Could you expose exactly what you try to do?

chr-hertel:
Closed in favor of #150

PR #77, opened 2025-09-21
[Server] Fix: ensure JSON-RPC error responses include message IDs

CodeWithKyrian:
This PR fixes a critical issue where JSON-RPC error responses were missing message IDs, causing clients to timeout waiting for responses.

**Problem:**
When errors occurred (method not found, invalid params, etc.), the error responses were sent without the original message ID. This caused clients to wait indefinitely for a response that would never come, leading to timeouts.

**Solution:**
- Modified `Handler.php` to pass message IDs to all error responses using the existing Error factory methods
- Updated `encodeResponse()` to handle JSON encoding errors gracefully with fallback error responses
- Removed unnecessary try-catch in `Server.php` since `encodeResponse()` no longer throws exceptions

### DIFF
--- a/src/JsonRpc/Handler.php
+++ b/src/JsonRpc/Handler.php
@@ -94,9 +94,6 @@ class Handler
 
     /**
      * @return iterable<array{string|null, array<string, mixed>}>
-     *
-     * @throws ExceptionInterface When a handler throws an exception during message processing
-     * @throws \JsonException     When JSON encoding of the response fails
      */
     public function process(string $input, ?Uuid $sessionId): iterable
     {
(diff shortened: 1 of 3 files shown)


chr-hertel:
I like that you cleaned up the exception handling - missed that one :+1: 

I wonder tho if we could ease the `Handler` with just passing the `$message` to the error and the error checks for `$message instance of Request` (it's only than that we should attach an ID, right?) - wdyt?

CodeWithKyrian:
Oh sure. That's even better. We shouldn't send a response for notifications, even if there's an error

CodeWithKyrian:
Oh wait. Seems I misunderstood what you meant @chr-hertel . You're suggesting we modify those factory methods in the Error class itself to accept Message instances instead of the id, yeah?

chr-hertel:
What got me thinking was the `$messageId` handling and the `method_exists`  - maybe there's a different way to make that cleaner, but handing over the message to the error was just one idea to try

CodeWithKyrian:
Yeah, I get it. The `method_exists` check is clunky. I think instead of passing the whole message to `Error` (which already expects just a message string), we can check if `$message instanceof Request` in the Handler and extract the ID there, otherwise use 0.

PR #81, opened 2025-09-25
docs: add more about building servers and updating front readme

chr-hertel:
Cleaning up front readme, and adding more docs for building servers

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -32,13 +32,13 @@ STDIO examples use standard input/output for communication:
 
 ```bash
 # Interactive testing with MCP Inspector
-npx @modelcontextprotocol/inspector php examples/01-discovery-stdio-calculator/server.php
+npx @modelcontextprotocol/inspector php examples/stdio-discovery-calculator/server.php
 
 # Run with debugging enabled
-npx @modelcontextprotocol/inspector -e DEBUG=1 -e FILE_LOG=1 php examples/01-discovery-stdio-calculator/server.php
+npx @modelcontextprotocol/inspector -e DEBUG=1 -e FILE_LOG=1 php examples/stdio-discovery-calculator/server.php
 
 # Or configure the script path in your MCP client
-# Path: php examples/01-discovery-stdio-calculator/server.php
+# Path: php examples/stdio-discovery-calculator/server.php
 ```
 
 ### HTTP Examples
@@ -47,7 +47,7 @@ HTTP examples run as web servers:
 
 ```bash
 # Start the server
-php -S localhost:8000 examples/02-discovery-http-userprofile/server.php
+php -S localhost:8000 examples/http-discovery-userprofile/server.php
 
 # Test with MCP Inspector
 npx @modelcontextprotocol/inspector http://localhost:8000
@@ -61,9 +61,9 @@ curl -X POST http://localhost:8000 \
 
 ## STDIO Examples
 
-### 01. Discovery STDIO Calculator
+### Discovery Calculator
 
-**File**: `examples/01-discovery-stdio-calculator/`
+**File**: `examples/stdio-discovery-calculator/`
 
 **What it demonstrates:**
 - Attribute-based discovery using `#[McpTool]` and `#[McpResource]`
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -71,59 +98,163 @@ require_once __DIR__ . '/vendor/autoload.php';
...
+    ->setDiscovery(
+        basePath: __DIR__,
+        scanDirs: ['.', 'src'],      // Default: ['.', 'src']
+        excludeDirs: ['vendor'],     // Default: ['vendor', 'node_modules']
+        cache: $cache
+    )
+    ->build();
+```
+
+## 📚 Documentation
```
chr-hertel: It would be more consistent to remove the Icon/emoji here as well

```suggestion
## Documentation
```

CodeWithKyrian: Sure. I thought I got them all 🙃

docs/examples.md
```diff
@@ -0,0 +1,273 @@
...
+# Test with curl
+curl -X POST http://localhost:8000 \
+  -H "Content-Type: application/json" \
+  -H "Accept: application/json, text/event-stream" \
+  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","clientInfo":{"name":"test","version":"1.0.0"},"capabilities":{}}}'
+```
+
+## STDIO Examples
+
+### 01. Discovery STDIO Calculator
```
chr-hertel: While Reading this i wonder if
* examples should have numbers, but only meaningful names
* transport should be configurable in examples via env or Option

Different PR tho, just wanted to get your opinion

CodeWithKyrian: For the naming, what do you mean by "should have numbers, but only meaningful names"? The examples already have both numbers and descriptive names.... are you suggesting removing the numbers, or that the current names aren't meaningful enough?

For the transport configuration, I get the flexibility appeal, but since STDIO and HTTP have fundamentally different usage patterns (single process vs web server), I still think separate examples are more appropriate

docs/examples.md
```diff
@@ -0,0 +1,273 @@
...
+
+**Usage:**
+```bash
+# Interactive testing
+npx @modelcontextprotocol/inspector php examples/01-discovery-stdio-calculator/server.php
+
+# Or configure in MCP client: php examples/01-discovery-stdio-calculator/server.php
+```
+
+### 03. Manual Registration STDIO
```
chr-hertel: somehow i struggle with the term "manual" registration - i can see where it's coming from, but it's basically just a set of adder methods. what do you think about using "explicit" instead?

docs/examples.md
chr-hertel: more like #87 - WDYT?

docs/examples.md
```diff
@@ -0,0 +1,273 @@
...
+All examples are located in the `examples/` directory and use the SDK dependencies from the root project. Most examples can be run directly without additional setup.
+
+### Prerequisites
+
+```bash
+# Install dependencies (from project root)
+composer install
+
+# Optional: Install HTTP-specific dependencies for web examples
+composer install --dev
```
chr-hertel: dev dependencies are installed by default so that doesn't make sense to me

docs/transports.md
```diff
@@ -0,0 +1,362 @@
...
+# Connect with MCP Inspector
+npx @modelcontextprotocol/inspector http://localhost:8000
+```
+
+## Choosing a Transport
+
+### Use STDIO When:
+
+- Building command-line tools
+- Integrating with MCP clients (Claude Desktop, etc.)
```
chr-hertel: that's confusing to me, Claude Desktop supports both transports, right?

docs/transports.md
```diff
@@ -0,0 +1,362 @@
...
+### Use HTTP When:
+
+- Building web services
+- Need browser compatibility
+- Integration with web frameworks
+- Multiple concurrent sessions
+- Web-based MCP clients
+- RESTful API patterns
+
+The key difference is that STDIO is process-based (one session per process) while HTTP is request-based (multiple sessions via headers).
```
chr-hertel: not sure i agree here - for me the key difference is connectivity. are you able to connect via STDIO since it's local? no, go HTTP.

docs/transports.md
CodeWithKyrian: Yeah, my bad!

chr-hertel:
This is def quite an improvement - won't be done forever anyways - should we merge @CodeWithKyrian?

CodeWithKyrian:
Yeah sure. We can always improve later on

CodeWithKyrian (approved):

PR #85, opened 2025-09-28
refactor: rename ServerBuilder to Builder

CodeWithKyrian:
Renames `ServerBuilder` to `Builder` to remove redundant naming within the `Mcp\Server` namespace.

## Motivation and Context

The "Server" prefix is redundant since the class is already in the Server namespace. This follows common PHP patterns and makes imports cleaner.

## Breaking Changes
- Import statements need to be updated from `ServerBuilder` to `Builder`
- Type hints need to be updated accordingly


chr-hertel (approved):
Did that once before was well, but was not sure - but looks like you're confirming that it makes sense :)

PR #86, opened 2025-09-28
[Server] Reorganize server handlers and transport structure for better organization

CodeWithKyrian:
Reorganizes the internal server structure to improve code organization and maintainability.

### Handler Reorganization
- **Moved**: `JsonRpc/Handler` → `Server/Handler/JsonRpcHandler`
- **Reorganized**: `Server/RequestHandler/` → `Server/Handler/Request/`
- **Reorganized**: `Server/NotificationHandler/` → `Server/Handler/Notification/`
- **Moved**: `MethodHandlerInterface` → `Server/Handler/` namespace

### Transport Structure
- **Moved**: `TransportInterface` → `Server/Transport/` namespace
- **Consolidated**: All transport-related code in one location

### Data Structure
- **Moved**: `Page.php` → `Schema/` namespace
- **Rationale**: It's a data structure concern, not server logic

### Better Organization
- **Logical grouping**: Related code is now co-located
- **Predictable structure**: Easier to find and understand code
- **Cleaner namespaces**: More intuitive import statements

### Separation of Concerns
- **Server-specific**: `JsonRpcHandler` is server implementation, not global protocol
- **Protocol-agnostic**: `MessageFactory` stays in `JsonRpc/` for reusability
- **Data vs Logic**: `Page` is a data structure, not server logic

## Breaking Changes

- **Internal only**: No public API changes
- **Import updates**: All internal references updated
- **Namespace changes**: Internal classes moved to new namespaces


chr-hertel:
in general i'm fine with this change, couple of thoughts tho:
* the `JsonRpcHandler` is quite a central service - even if currently internal to the server i was expecting it to be useful with the client as well, but we can tackle that then anyways.
* moving the `Page` to the schema def makes sense - next step might be to have that PaginatedRequest base class :D
* the `MessageFactory` feels lonely, but maybe also something to revisit when working on the client.

chr-hertel (approved):

CodeWithKyrian:
- Yeah true, its internal to the server at the moment that's why I moved it. For the client, the handling process will be very much different I think, but as you said, we'll cross that bridge when we get there. .
- You read my mind!! +1 on that
- Sure!

PR #87, opened 2025-09-28
[Examples] Rename example folders

chr-hertel:
Not sure, but at least i like that they now consistently start with the transport - that's easier to navigate


CodeWithKyrian:
Sure. This works!

Issue #68, opened 2025-09-14
[Server] No Ability to set outputSchema


chr-hertel:
https://modelcontextprotocol.io/specification/2025-06-18/server/tools#data-types
Tools are missing output schema support, which would be a very useful feature.

bigdevlarry:
Hi, @chr-hertel, can I pick this one up?

chr-hertel:
Sure @bigdevlarry , go ahead! :)

bigdevlarry:
Linking back the updated PR - https://github.com/modelcontextprotocol/php-sdk/pull/153

PR #88, opened 2025-09-28
issues-68 No Ability to set outputSchema

bigdevlarry:
The changes add comprehensive support for outputSchema in the Tools definition, enabling MCP servers to specify the expected structure and types of tool responses. 

This includes:
 - Added outputSchema parameter to Builder::addTool() method
-  Updated Tool class constructor and JSON serialization to support outputSchema
- Added validation for outputSchema structure and type requirements
- Enhanced existing tests to verify outputSchema functionality works correctly
- Fixed related type annotation issues and code style problems

This allows MCP clients to validate tool responses against the defined schema and provides better type safety and documentation for tool interactions.

## Motivation and Context

This was raised as an issue for the output schema to be added

## Types of changes

- [ ] Documentation update

## Checklist


bigdevlarry:
@chr-hertel PR to address https://github.com/modelcontextprotocol/php-sdk/issues/68 is now up for review

.gitignore
```diff
@@ -5,3 +5,4 @@ coverage
 vendor
 examples/**/dev.log
 examples/**/sessions
+.idea
```
chr-hertel: please revert this, it's specific to your setup

a common practice to add OS/IDE specific things to gitignore, is using a [global one](https://stackoverflow.com/questions/7335420/global-git-ignore)

chr-hertel:
Hi @bigdevlarry, thanks for working on this - look great already. was thinking tho if it makes sense to extend maybe one of our examples - or bring in a new one to demo this?

WDYT & cc @CodeWithKyrian

bigdevlarry:
I didn’t know how to extend but I’m happy with either extending or bringing a new example tho …

.gitignore
bigdevlarry: Brill, thanks for flagging will push an update

bigdevlarry:
Okay, that make sense. You can have a look again

chr-hertel:
Pipeline issue related to https://github.com/modelcontextprotocol/inspector/issues/834

CodeWithKyrian:
This is solid work, @bigdevlarry ! Appreciate the effort you put into this!

On a side note, I was wondering if there’s a way we could also infer the `outputSchema` when it’s not explicitly defined, similar to how we handle `inputSchema`. Also, I'm thinking of the `Schema` attribute (or even a dedicated `OutputSchema` attribute), if and how that could come into play here. 

Just a thought though...curious what you all think.

ineersa:
https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content

While working on my MCP this one is not seems to be working too.

It looks like if outputSchema provided server must also provide structuredContent alongside content. 

I will try this PR in my fork and will try to implement structured content in CallToolResult.

Because of this I wouldn't add auto schema definition, developer should be aware that by adding `outputSchema` they should also add `structuredContent`.

bigdevlarry:
@CodeWithKyrian, thanks for the shout and feedback.  I've included support to help infer the outputSchema when it’s not explicitly defined, similar to how we handle inputSchema. Also on this one - 

> Also, I'm thinking of the Schema attribute (or even a dedicated OutputSchema attribute), if and how that could come into play here.

I'm not sure how you mean here. If you create a separate issue with more description, then I can pick it up. 

@chr-hertel Could you give the PR another look? I think should be good

bigdevlarry:
hi @chr-hertel Still need a review/feedback here

src/Capability/Discovery/DocBlockParser.php
```diff
@@ -136,4 +136,51 @@ public function getParamTypeString(?Param $paramTag): ?string
 
         return null;
     }
+
+    /**
+     * Gets the return type string from a Return tag.
+     */
+    public function getReturnTypeString(?DocBlock $docBlock): ?string
+    {
+        if (!$docBlock) {
```
chr-hertel: always like to be explicit
```suggestion
        if (null === $docBlock) {
```

src/Capability/Discovery/DocBlockParser.php
```diff
@@ -136,4 +136,51 @@ public function getParamTypeString(?Param $paramTag): ?string
...
+            return null;
+        }
+
+        $returnTags = $docBlock->getTagsByName('return');
+        if (empty($returnTags)) {
```
chr-hertel: and [try to avoid empty](https://www.beberlei.de/post/when_to_use_empty_in_php_i_say_never):
```suggestion
        if ([] === $returnTags) {
```

src/Capability/Discovery/DocBlockParser.php
```diff
@@ -136,4 +136,51 @@ public function getParamTypeString(?Param $paramTag): ?string
...
+            return null;
+        }
+
+        $returnTag = $returnTags[0];
+        if (method_exists($returnTag, 'getType') && $returnTag->getType()) {
```
chr-hertel: can we check for `TagWithType` instead of using `method_exists`?

chr-hertel: and should be inverted to:
```php
if (!$returnTag instanceof TagWithType) {
    return null;
}
```

it's better to continue with the style of early exits - like you did in the beginning of the method

src/Capability/Discovery/SchemaGenerator.php
```diff
@@ -79,6 +79,36 @@ public function generate(\ReflectionMethod|\ReflectionFunction $reflection): arr
...
+            ? $docBlockReturnType
+            : ($reflectionReturnTypeString ?: $docBlockReturnType);
+
+        if (!$returnTypeString || 'void' === strtolower($returnTypeString)) {
+            return null;
+        }
+
+        return $this->buildOutputSchemaFromType($returnTypeString, $returnDescription);
+    }
+
```
chr-hertel, line 113: please add tests for this method

src/Capability/Discovery/DocBlockParser.php
```diff
@@ -136,4 +136,51 @@ public function getParamTypeString(?Param $paramTag): ?string
...
+            return null;
+        }
+
+        $returnTag = $returnTags[0];
+        $description = method_exists($returnTag, 'getDescription')
+            ? trim((string) $returnTag->getDescription())
+            : '';
+
+        return $description ?: null;
+    }
```
chr-hertel, line 187: similar comments would apply to this method like with `getReturnTypeString`

PR #90, opened 2025-09-29
[Server] Fix missing instructions from builder to handler

chr-hertel:
Missed to wire up the config correctly.

### DIFF
--- a/examples/stdio-discovery-calculator/server.php
+++ b/examples/stdio-discovery-calculator/server.php
@@ -20,6 +20,7 @@ logger()->info('Starting MCP Stdio Calculator Server...');
 
 $server = Server::builder()
     ->setServerInfo('Stdio Calculator', '1.1.0', 'Basic Calculator over STDIO transport.')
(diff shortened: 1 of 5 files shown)


CodeWithKyrian:
Perfect. I was actually wondering if we still needed that `Configuration` DTO before this.

Though I’m thinking… what’s its role now, really? I first added it because there were too many parameters being passed around (server builder → server → handler), and I just wanted a typed object to hold all the configured services/settings in one place.

Right now it only has `serverInfo`, `capabilities`, `paginationLimit`, and instructions , but the `JsonRpcHandler::make` method still takes `paginationLimit` and the config object, which kinda defeats the point.

Also, that make method has quite a few arguments. Some of them (like `toolCaller`, `resourceReader`, `promptGetter`, maybe even the session store/factory...or not?) could fit nicely inside `Configuration`. Might be overkill anyways 😂, but since we already have it, maybe it’s worth thinking about.

Anyway, for now, removing the `paginationLimit` is a good start.

chr-hertel:
I like that it bundles some arguments into a structure - but it's not that semantically strong - that's true.

CodeWithKyrian (approved):

PR #91, opened 2025-09-29
[Server] Add happy cases for functional tests of example 1

chr-hertel:
It's only happy cases yet, but test by test, step by step, we'll cover more.

### DIFF
--- a/tests/Inspector/InspectorSnapshotTestCase.php
+++ b/tests/Inspector/InspectorSnapshotTestCase.php
@@ -19,20 +19,42 @@ abstract class InspectorSnapshotTestCase extends TestCase
 {
     private const INSPECTOR_VERSION = '0.16.8';
 
+    /**
+     * @param array<string, mixed> $toolArgs
+     */
     #[DataProvider('provideMethods')]
-    public function testResourcesListOutputMatchesSnapshot(string $method): void
-    {
-        $process = (new Process([
-            'npx',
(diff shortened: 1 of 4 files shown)


tests/Inspector/InspectorSnapshotTestCase.php
```diff
@@ -17,14 +17,41 @@
...
+    /**
+     * @param array<string, mixed> $toolArgs
+     */
     #[DataProvider('provideMethods')]
-    public function testResourcesListOutputMatchesSnapshot(string $method): void
-    {
-        $process = Process::fromShellCommandline(
-            \sprintf('npx @modelcontextprotocol/inspector --cli php %s --method %s', $this->getServerScript(), $method)
-        )->mustRun();
+    public function testResourcesListOutputMatchesSnapshot(
```
CodeWithKyrian: The name’s a little confusing since the test covers multiple methods (not just `resources/list`). Maybe we could rename it to something more neutral like `testInspectorOutputMatchesSnapshot` or `testMethodOutputMatchesSnapshot` to better reflect what’s happening?

CodeWithKyrian: Thoughts?

chr-hertel: nooo, you're completely right, that's a mess - and a left over i guess

PR #93, opened 2025-10-05
[Server] Structured content return and tool error

ineersa:
Added optional `structuredContent` for `CallToolResult` which enables return of structured content when we provide output schema for a tool.

Added ability to return `CallToolResult` from tool directly, which enables proper return of tool errors to allow LLM to self-correct tool call.

## Motivation and Context

While testing my MCP server I've noticed inconsistencies between output with Python MCP server.
1. There was no possibility and no implementation for `structuredContent`
2. There was no possibility to return `isError=true`

Also while `outputSchema` being added in https://github.com/modelcontextprotocol/php-sdk/pull/88 , with `outputSchema` MCP server **MUST** provide structured content alongside text representation.

## How Has This Been Tested?

Tested on my MCP server to ensure same response between PHP and Python versions.
Added unit tests.

## Breaking Changes

No breaking changes

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/Schema/Result/CallToolResult.php
+++ b/src/Schema/Result/CallToolResult.php
@@ -38,12 +38,14 @@ class CallToolResult implements ResultInterface
     /**
      * Create a new CallToolResult.
      *
-     * @param Content[] $content The content of the tool result
-     * @param bool      $isError Whether the tool execution resulted in an error.  If not set, this is assumed to be false (the call was successful).
+     * @param Content[] $content           The content of the tool result
+     * @param bool      $isError           Whether the tool execution resulted in an error.  If not set, this is assumed to be false (the call was successful).
+     * @param mixed[]   $structuredContent JSON content for `structuredContent`
      */
     public function __construct(
         public readonly array $content,
         public readonly bool $isError = false,
+        public readonly ?array $structuredContent = null,
     ) {
         foreach ($this->content as $item) {
             if (!$item instanceof Content) {
@@ -107,9 +109,15 @@ class CallToolResult implements ResultInterface
      */
     public function jsonSerialize(): array
     {
-        return [
+        $result = [
             'content' => $this->content,
             'isError' => $this->isError,
         ];
+
+        if ($this->structuredContent) {
+            $result['structuredContent'] = $this->structuredContent;
+        }
+
+        return $result;
     }
 }
--- a/src/Server/Handler/Request/CallToolHandler.php
+++ b/src/Server/Handler/Request/CallToolHandler.php
@@ -59,14 +59,17 @@ final class CallToolHandler implements RequestHandlerInterface
             }
 
             $result = $this->referenceHandler->handle($reference, $arguments);
-            $formatted = $reference->formatResult($result);
+
+            if (!$result instanceof CallToolResult) {
+                $result = new CallToolResult($reference->formatResult($result));
+            }
 
             $this->logger->debug('Tool executed successfully', [
                 'name' => $toolName,
                 'result_type' => \gettype($result),
             ]);
 
-            return new Response($request->getId(), new CallToolResult($formatted));
+            return new Response($request->getId(), $result);
         } catch (ToolNotFoundException $e) {
             $this->logger->error('Tool not found', ['name' => $toolName]);
 
--- a/tests/Unit/Server/Handler/Request/CallToolHandlerTest.php
+++ b/tests/Unit/Server/Handler/Request/CallToolHandlerTest.php
@@ -342,6 +342,64 @@ class CallToolHandlerTest extends TestCase
         $this->assertEquals($expectedResult, $response->result);
     }
 
+    public function testHandleReturnsStructuredContentResult(): void
+    {
+        $request = $this->createCallToolRequest('structured_tool', ['query' => 'php']);
+        $toolReference = $this->createMock(ToolReference::class);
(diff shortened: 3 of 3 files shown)


src/Capability/Tool/ToolCaller.php
```diff
@@ -59,15 +60,24 @@ public function call(CallToolRequest $request): CallToolResult
...
+            /** @var array<int, TextContent|ImageContent|EmbeddedResource|AudioContent|StructuredContent> $formattedResult */
             $formattedResult = $toolReference->formatResult($result);
 
             $this->logger->debug('Tool executed successfully', [
                 'name' => $toolName,
                 'result_type' => \gettype($result),
             ]);
 
-            return new CallToolResult($formattedResult);
+            return CallToolResult::fromArray($formattedResult);
```
soyuka: I'm not fond of this change, I think it'd be better to have some kind of Result that has both contents and structured content:
- we could still return a collection of content's (and add structured content if the result has some)
- at some point we will probably only return structured content. 

For example:

```
StructuredContentResult {
  public StructuredContent
}
```

ineersa: You mean something like 
```
StructuredContentResult implements CallToolResult {
  public StructuredContent
  public array $contents
}

ContentResult implements CallToolResult {
  public array $contents
}

interface CallToolResult extends \JsonSerializable{
}

```

Then we can expect instance of `CallToolResult` interface as return and maybe skip 
```php
/** @var array<int, TextContent|ImageContent|EmbeddedResource|AudioContent|StructuredContent> $formattedResult */
$formattedResult = $toolReference->formatResult($result);
```
and expect user to return proper response from tool?

> at some point we will probably only return structured content.

Structured content is not required while `outputSchema` is not set, and I can argue that for some tasks text response works better.  
At least for now content response [required](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content) even if `structuredContent` is set. 

This implementation is far from best, so if you could steer me in right direction would be great.

I used it in my [browser-mcp](https://github.com/ineersa/browser-mcp/blob/main/src/Tools/SearchTool.php) tools, most crucial part was ability to return proper tool usage errors with hints to help model correct tool calls. Structured content provided no improvements for tool calling and quality so I just removed it from final version.

soyuka: I'm using this: 

```php
<?php

namespace ApiPlatform\Mcp\Schema\Result;

use Mcp\Schema\JsonRpc\ResultInterface;

class StructuredContentResult implements ResultInterface
{
    /**
     * Create a new StructuredContentResult.
     *
     * @param array     $structuredContent The JSON content
     * @param ResultInterface $result A traditional result
     */
    public function __construct(
        public array $structuredContent,
        public readonly ?ResultInterface $result = null
    ) {
    }

    public function jsonSerialize(): mixed
    {
        return ['structuredContent' => $this->structuredContent] + ($this->result?->jsonSerialize() ?? []);
    }
}
``` 

This is merely a suggestion :).

ineersa: I changed implementation, removed some changes.

Now we have `CallToolResultInterface` and 2 classes implementing it:
 - `CallToolResult` as our general result class
 - `CallToolStructuredContentResult` as SturcturedContent result which pretty much decorates `CallToolResult`
 
 Tested on my [Browser MCP](https://github.com/ineersa/browser-mcp/commit/b7f86e5e82d2979fb146f65abf0d920a4a4602c5) it's working as expected.
 
 Also added Unit tests to cover behavior. 
 
 As for example of usage, maybe we can add one after https://github.com/modelcontextprotocol/php-sdk/pull/88 merged as they go hand by hand together.
 
 @chr-hertel @CodeWithKyrian could you take a look when will get time, it's ready for proper CR now.

chr-hertel: I'm a bit confused - the PR description doesn't match the code or do I miss something?

chr-hertel: I like the idea of enabling tools to return the `CallToolResult` optionally if they want to have more control, but let's stick with one implementation here.

If we patch it in the lib we don't need an additional decoration, but can extend the `CallToolResult` with an optional structure argument, that is omitted in case it's empty. That would be also more consistent with what we have in other data classes and makes it easier in change scenarios. currently spec changes are only additive for example and users relying on using those classes would need to change.

ineersa: > I'm a bit confused - the PR description doesn't match the code or do I miss something?

Yeah, I changed implementation a lot after previous comment and did rework, I can update it. 
Comments here describe main idea more.

> If we patch it in the lib we don't need an additional decoration, but can extend the `CallToolResult` with an optional structure argument, that is omitted in case it's empty. That would be also more consistent with what we have in other data classes and makes it easier in change scenarios. currently spec changes are only additive for example and users relying on using those classes would need to change.

Okay, gonna rework to keep single `CallToolResult` with optional `structuredContent`.

> I like the idea of enabling tools to return the `CallToolResult` optionally if they want to have more control, but let's stick with one implementation here.

That was actually even more beneficial for me to return proper `isError` to LLM in response with hints and error explanation so it could do correction, if `isError=false` always returned as it is now LLM don't always understand what happened and that it needs to correct tool call.

ineersa: @chr-hertel Moved `structuredContent` to `CallToolResult`. Fixed description. Updated unit tests accordingly. Tested responses in inspector, works as expected.

PR #99, opened 2025-10-06
Removing inspector tests, implemented application tests instead

ineersa:
<!-- Provide a brief summary of your changes -->
 Implemented application tests as proof of concept instead of inspector tests

## Motivation and Context
<!-- Why is this change needed? What problem does it solve? -->
Currently pipelines are failing
<img width="1449" height="696" alt="image" src="https://github.com/user-attachments/assets/89c9a4ca-4519-47ff-b9dd-213ee5ae1737" />
I believe that PHP version of tests fits better to PHP SDK and has a lot of benefits.

## How Has This Been Tested?
<!-- Have you tested this in a real application? Which scenarios were tested? -->

## Breaking Changes
<!-- Will users need to update their code or configurations? -->
No

## Types of changes
<!-- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
- [ ] Documentation update

## Checklist
<!-- Go over all the following points, and put an `x` in all the boxes that apply. -->

## Additional context
<!-- Add any other context, implementation notes, or design decisions -->

I couldn't test initialize response via Inspector tests, so as discussed in slack tried to implement Application tests as proof of concept.


| Category | Pros | Cons                                                                                                                                                                                                                                                                                                                                                                   |
|-----------|------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| **Inspector Tests** | - Automatically validate supported methods and error handling | - Run as global `npx` package, no version pin<br>- Require `npm`/`npx`, adds infrastructure dependency<br>- Generally slow<br>- Limited method coverage (`tools/list`, `tools/call`, `resources/*`, `prompts/*`, `logging/setLevel`)<br>- Cannot access `initialize` response, can’t test server capabilities<br>- Coding agents can’t run `npx` inside sandboxed envs |
| **Application Tests** | - Run directly against MCP server responses<br>- No `npm`/`npx` dependency<br>- Very fast (≈200 ms for `ManualStdioExampleTest`)<br>- Easy to implement<br>- Flexible — can extend with custom client logic (e.g. initialization, logger setup) | - Must implement client-side logic manually **once** <br>- Missing handlers won’t automatically raise MCP errors                                                                                                                                                                                                                                                       |

If we will add some client logic to ApplicationTestCase like initialization, logger setting etc. it will work same as Inspector but without overheads of npm. 

Test runs are very fast:
<img width="662" height="206" alt="image" src="https://github.com/user-attachments/assets/997568d4-6f4c-4e3e-becc-b0632571fe53" />


Testing is quite easy to implement, server responses indexed by ids and very easy to get for comparison.


chr-hertel:
Oh, the scope changed a bit from additional to replacing - can tell you right away that this is not on the table atm.

But I can share my motivation for using the Inspector
* have a high-level functional test suite
* use what I actually use to test the servers
* not re-implement the client side
* rely on something more neutral instead of the PHP SDK itself - external validation

If you have trouble running the inspector tests locally, I think it is fine to skip them and rely on the pipeline. that's actually also why we split them into separate pipeline. currently those tests take 6-8 seconds locally for me, so not a big deal. not blazing fast tho, true.

Regarding your pro/cons:
* Inspector Pro: you're missing the part i listed above & that it is ready - we can use it out of the box
  * you're missing the fact that we validate our implementation with an external tool - it's basically contract testing - 
* Inspector Con:
  * well, the first one get's addressed with #98 
  * I don't see the issue with the dependency, but i'm using npx regularly
  * I give you the limited method coverage, yes
  * Testing the initialize was not the purpose here - it is only one part of the pyramid, the goal is not to test all concerns and skip unit tests
 * Application Tests Pro
   * "Easy to implement" not from the LLOC compared to the other one
   * I feel like your inverting a few inspector cons here :D
 * Application Tests Con
   * I wonder if we can actually use a client implementation later when we have it - feels a bit like incest still

---

This only to share my thoughts and reasoning, but let's go back a few steps.

Which limitation did you face that you needed to act on this? or was is mostly the npx and inspector bug that threw you off?

ineersa:
```
This only to share my thoughts and reasoning, but let's go back a few steps.
Which limitation did you face that you needed to act on this? or was is mostly the npx and inspector bug that threw you off?
```

I was checking how to test MCP server I was working on and was looking at inspector tests, but went with separate implementation which worked really well for my case.

For tests here, yes it was mostly that npx inspector bug which threw me off, so I decided to make this concept into PR.

Also I agree that without client implementation this will be not complete tests.

Maybe we could get back to this when we will have reference MCP Client implementation, then inspector replacement will be super easy, we can close it for now.

Thank you for your feedback!

PR #103, opened 2025-10-08
[Server] Simplify server entrypoint with run-based transport lifecycle

CodeWithKyrian:
This PR replaces the old `connect()` + `listen()` flow with a single `run()` method that wires, executes, and tears down transports for every example, doc, and transport implementation.

## Motivation and Context

- Simplifies server usage so developers no longer have to remember to call both `connect() `and `listen()` and worry about cleanup.
- Keeps lifecycle responsibilities inside the SDK: transports expose the same interface (`initialize`, `onMessage`, `listen`, `close`) while the server coordinates their use and always closes them once `listen()` finishes...whether that’s a blocking STDIO loop,  a one-shot HTTP request, or blocking asynchronous loop.
 
## What’s Changed

- Introduced `Server::run()` which initializes the transport, hooks up callbacks, runs `listen()`, closes the transport, and returns whatever the transport returns.
- Updated STDIO transport to emit an integer status (`0` for success, `1` when `STDIN` reading fails) and cleared session state during close.
- Refreshed all docs and examples to call `run()` and, for STDIO scripts, exit with the returned status.
- Extended the HTTP samples to rely on the response produced by `run()` instead of invoking `listen()` manually.

## Breaking Changes

- `Server::connect()` has been removed. All consumers must call `Server::run($transport)` and drop their manual `$transport->listen()`/`$transport->close()` calls (if any).

### DIFF
--- a/tests/Unit/Capability/Discovery/DocBlockTestFixture.php
+++ b/tests/Unit/Capability/Discovery/DocBlockTestFixture.php
@@ -75,7 +75,7 @@ class DocBlockTestFixture
      * @deprecated use newMethod() instead
      * @see DocBlockTestFixture::newMethod()
      */
-    public function methodWithMultipleTags(float $value): bool /* @phpstan-ignore throws.unusedType */
+    public function methodWithMultipleTags(float $value): bool
     {
         return true;
     }


src/Server.php
```diff
@@ -35,7 +35,7 @@ public static function builder(): Builder
         return new Builder();
     }
 
-    public function connect(TransportInterface $transport): void
+    public function run(TransportInterface $transport): mixed
```
chr-hertel, line 45: which scenarios do we have instead of `int`?

CodeWithKyrian: Check the `StreamableHttpTransport` and the http examples. For that, It returns a PSR-7 response instead

PR #106, opened 2025-10-11
[Server] Refactor protocol pipeline to separate request and notification handling

CodeWithKyrian:
This PR introduces a new `Protocol` layer, reshapes message parsing, and redefines how custom handlers plug into the server. The `Protocol` exposes `handleRequest`, `handleNotification`, and a placeholder `handleResponse`, along with `sendRequest`, `sendResponse`, and `sendNotification`. Only response sending is wired for now, but the other direction-facing methods lay the foundation for server-initiated messaging.

## Motivation and Context

- The old `JsonRpcHandler`/`MethodHandlerInterface` blended request and notification semantics, making it impossible to enforce the JSON‑RPC contract (requests must yield `Response`|`Error`, notifications return **nothing**).
- Preparing for server‑initiated messaging (sampling, logging, custom notifications) required a bidirectional protocol layer that can send as well as receive.
- The existing message factory only understood requests/notifications; it could not parse responses or errors, blocking any future server‑initiated flows.
- To keep responsibilities clear, the server should manage the transport lifecycle (`initialize` → `connect` → `listen` → `close`) while the protocol owns all transport callbacks (`onMessage`, `onSessionEnd`, and any future hooks) and output routing.

## What’s Changed

- Renamed `JsonRpcHandler` to `Protocol` and made it the transport-facing façade that registers callbacks, parses all JSON‑RPC message types, dispatches request/notification handlers, and emits responses.
- Shifted `Server::run()` into a small lifecycle orchestrator: it initializes the transport, connects the protocol, runs `listen()`, and guarantees `close()`. No more callback wiring in Server.
- Replaced the coarse `MethodHandlerInterface` with `RequestHandlerInterface` and `NotificationHandlerInterface`, enforcing that requests always return `Response`|`Error` and notifications remain fire-and-forget.
- Refactored `MessageFactory` to emit typed `Request`, `Notification`, `Response`, and `Error` objects (or `InvalidInputMessageException`), laying the groundwork for bidirectional messaging.
- Added new builder hooks,  `addRequestHandler(s)` and `addNotificationHandler(s)`,  and updated docs, examples, and tests to use the new abstractions.

## Breaking Changes
- `Builder::addMethodHandler(s)` and `MethodHandlerInterface` have been removed.
- Use `addRequestHandler(s)`/`RequestHandlerInterface` for JSON‑RPC requests and
`addNotificationHandler(s)`/`NotificationHandlerInterface` for notifications.

### DIFF
--- a/src/JsonRpc/MessageFactory.php
+++ b/src/JsonRpc/MessageFactory.php
@@ -35,11 +35,11 @@ use Mcp\Schema\JsonRpc\Response;
 final class MessageFactory
 {
     /**
-     * Registry of all known notification classes.
+     * Registry of all known message classes that have methods.
      *
-     * @var array<int, class-string<Notification>>
+     * @var array<int, class-string<Request|Notification>>
      */
-    private const REGISTERED_NOTIFICATIONS = [
+    private const REGISTERED_MESSAGES = [
         Schema\Notification\CancelledNotification::class,
         Schema\Notification\InitializedNotification::class,
         Schema\Notification\LoggingMessageNotification::class,
@@ -49,14 +49,7 @@ final class MessageFactory
         Schema\Notification\ResourceUpdatedNotification::class,
         Schema\Notification\RootsListChangedNotification::class,
         Schema\Notification\ToolListChangedNotification::class,
-    ];
 
-    /**
-     * Registry of all known request classes.
-     *
-     * @var array<int, class-string<Request>>
-     */
-    private const REGISTERED_REQUESTS = [
         Schema\Request\CallToolRequest::class,
         Schema\Request\CompletionCompleteRequest::class,
         Schema\Request\CreateSamplingMessageRequest::class,
@@ -75,32 +68,24 @@ final class MessageFactory
     ];
 
     /**
-     * @param array<int, class-string<Notification>> $registeredNotifications
-     * @param array<int, class-string<Request>>      $registeredRequests
+     * @param array<int, class-string<Request|Notification>> $registeredMessages
      */
     public function __construct(
-        private readonly array $registeredNotifications,
-        private readonly array $registeredRequests,
+        private readonly array $registeredMessages,
     ) {
-        foreach ($this->registeredNotifications as $notification) {
-            if (!is_subclass_of($notification, Notification::class)) {
-                throw new InvalidArgumentException(\sprintf('Notification classes must extend %s.', Notification::class));
-            }
-        }
-
-        foreach ($this->registeredRequests as $request) {
-            if (!is_subclass_of($request, Request::class)) {
-                throw new InvalidArgumentException(\sprintf('Request classes must extend %s.', Request::class));
+        foreach ($this->registeredMessages as $messageClass) {
+            if (!is_subclass_of($messageClass, Request::class) && !is_subclass_of($messageClass, Notification::class)) {
+                throw new InvalidArgumentException(\sprintf('Message classes must extend %s or %s.', Request::class, Notification::class));
             }
         }
     }
 
     /**
-     * Creates a new Factory instance with all the protocol's default notifications and requests.
+     * Creates a new Factory instance with all the protocol's default messages.
      */
     public static function make(): self
     {
-        return new self(self::REGISTERED_NOTIFICATIONS, self::REGISTERED_REQUESTS);
+        return new self(self::REGISTERED_MESSAGES);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


tests/Unit/Server/ProtocolTest.php
```diff
@@ -0,0 +1,619 @@
...
+final class ProtocolTest extends TestCase
+{
+    /** @var MockObject&SessionFactoryInterface */
+    protected $sessionFactory;
+
+    /** @var MockObject&SessionStoreInterface */
+    protected $sessionStore;
+
+    /** @var MockObject&TransportInterface<mixed> */
+    protected $transport;
```
chr-hertel: i favor being most defensive and using `private` as default visibility for properties and method. only opting into promoting it to `protected` - also let's use typed properties

```suggestion
    private MockObject&SessionFactoryInterface $sessionFactory;
    private MockObject&SessionStoreInterface $sessionStore;
    private MockObject&TransportInterface $transport;
```

chr-hertel:
a bit funny to see that because having notification and request handlers separate was exactly how i started:
https://github.com/php-llm/mcp-sdk/tree/5bfd6bf757da756501b73242e11cf386de0c6ddc/src/Server

i treated both similar, that why i brought it up one level and unified. buuut, the context here is different - so i'm in for doing it :+1:

src/JsonRpc/MessageFactory.php
```diff
@@ -13,107 +13,227 @@
...
+     *
+     * Supports both single messages and batch requests. Returns an array containing
+     * MessageInterface objects or InvalidInputMessageException instances for invalid messages.
+     *
+     * @return array<MessageInterface|InvalidInputMessageException>
      *
      * @throws \JsonException When the input string is not valid JSON
      */
-    public function create(string $input): iterable
+    public function create(string $input): array
```
chr-hertel, line 101: sad to see that go, but it was broken anyway since we had that `iterator_to_array` and i can't come up with a concrete benefit right now

CodeWithKyrian: Yeah. Spec requires we review all messages and make sure initialize request can't be part of a batch so we can't go that route.

src/JsonRpc/MessageFactory.php
chr-hertel, line 1: yea, the redundancy here let me go one step higher since the logic only relies on the method - the factory knows nothing  about the difference of notifications or requests, and doesn't have to, it cares about methods.

not sure i see this contributing to bi-directional problem, but no strong opinion against it, just a bit bloated from my point of view.

CodeWithKyrian: I'm not sure I get what you're talking about here

chr-hertel: it's not a big deal anyhow, but rather a design perspective. the `MessageFactory` has the purpose of converting a json string into a data class. it does not really care about the data class. it basically does the routing to the correct `fromArray` method.
what does it need to do that job? does it need to know about the difference about notifications or requests? not really, it only needs methods, they are a commonality and unique.

CodeWithKyrian: Just to make sure I understand correctly: are you suggesting we could merge the `REGISTERED_NOTIFICATIONS` and `REGISTERED_REQUESTS` arrays back into a single array since the method field already uniquely identifies each message class (or rather just Notification and Request classes) regardless of whether it's a request or notification?

If so, yeah, I totally get that point. Since both are distinguished by their method names anyway, we could technically use a single registry. The separation was more of a semantic choice to mirror the JSON-RPC spec's distinction between requests (with id) and notifications (without id), but I agree it's not strictly necessary from a functional perspective.

The part contributing to the bi-directional problem was including Responses and Errors. I'm open to consolidating them if you think that's cleaner.

chr-hertel: well, it was like that before with the motivation of encapsulation of concerns, but do you feel that makes it harder to understand what's going on?

CodeWithKyrian: It doesn't really matter that much tbh. I'll just consolidate

CodeWithKyrian: Good now?

tests/Unit/JsonRpc/MessageFactoryTest.php
```diff
@@ -261,18 +257,20 @@ public function testUnknownNotificationMethod(): void
...
+        // The fromArray validation will handle any issues
+        $json = '{"jsonrpc": "2.0", "method": "notifications/initialized", "id": 1}';
 
         $results = $this->factory->create($json);
 
         $this->assertCount(1, $results);
+        // The notification class will reject the id in fromArray validation
         $this->assertInstanceOf(InvalidInputMessageException::class, $results[0]);
-        $this->assertStringContainsString('id', $results[0]->getMessage());
     }
```
chr-hertel, line 274: extra mile :sunglasses: :+1:

src/JsonRpc/MessageFactory.php
chr-hertel: Thanks Kyrian :)

PR #107, opened 2025-10-12
Removed redundant Builder.setServerCapabilities duplicate method

Adebayo120:
While trying to integrate logging with ServerCapabilities, I discovered that we have two methods and private properties on the Builder class that store ServerCapabilities. This PR removes the duplicate, so we maintain just one interface for setting ServerCapabilities
## How Has This Been Tested?

Yes
## Breaking Changes

No
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/Server/Builder.php
+++ b/src/Server/Builder.php
@@ -75,8 +75,6 @@ final class Builder
 
     private ?string $instructions = null;
 
-    private ?ServerCapabilities $explicitCapabilities = null;
-
     /**
      * @var array<int, MethodHandlerInterface>
      */
@@ -177,9 +175,9 @@ final class Builder
     /**
      * Explicitly set server capabilities. If set, this overrides automatic detection.
      */
-    public function setCapabilities(ServerCapabilities $capabilities): self
+    public function setCapabilities(ServerCapabilities $serverCapabilities): self
     {
-        $this->explicitCapabilities = $capabilities;
+        $this->serverCapabilities = $serverCapabilities;
 
         return $this;
     }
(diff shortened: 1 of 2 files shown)


src/Server/Builder.php
```diff
@@ -370,7 +361,7 @@ public function build(): Server
         $sessionStore = $this->sessionStore ?? new InMemorySessionStore($sessionTtl);
         $messageFactory = MessageFactory::make();
 
-        $capabilities = $this->explicitCapabilities ?? $registry->getCapabilities();
+        $capabilities = $registry->getCapabilities();
```
CodeWithKyrian, line 364: Please update to use the provided `serverCapabilities` and then default to the registry's if null, same as before, but with the updated property name

Adebayo120: I might be missing something, but from what I can see in the codebase — when a user explicitly sets the serverCapabilities, we store it as a private property in the Builder class. Then, during the build process, the first thing we do is create the registry and check whether the user provided a serverCapability; if so, we store it in the registry.

Given that, calling the registry to retrieve the serverCapability should already be sufficient, since it either returns the user-defined one or falls back to the internally built default. Am I missing something, or is there a specific reason we need to repeat what the registry already handles within the Builder class?

CodeWithKyrian:
Please [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits before we can merge

Adebayo120:
> Please [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits before we can merge

Done @CodeWithKyrian

PR #110, opened 2025-10-17
[Server] Missing Mcp-Protocol-Version to access control

soyuka:


src/Server/Transport/StreamableHttpTransport.php
```diff
@@ -44,7 +44,7 @@ class StreamableHttpTransport implements TransportInterface
     private array $corsHeaders = [
         'Access-Control-Allow-Origin' => '*',
         'Access-Control-Allow-Methods' => 'GET, POST, DELETE, OPTIONS',
-        'Access-Control-Allow-Headers' => 'Content-Type, Mcp-Session-Id, Last-Event-ID, Authorization, Accept',
+        'Access-Control-Allow-Headers' => 'Content-Type, Mcp-Session-Id, Mcp-Protocol-Version, Last-Event-ID, Authorization, Accept',
     ];
```
soyuka, line 48: Would it be possible to leave security features to the actual HTTP Proxy server that's running PHP (ie ngin/php/symfony)? Indeed this is hiding security settings and we've no ways of changing it.

chr-hertel: I agree that this is quite opinionated - at least we should have some config to shut it off or overwrite the actual values

PR #113, opened 2025-10-21
[Server] bugfix: Cache discovery is unavailable

okxaas:
## Changes

Moved discovery state registration logic from the `Discoverer` class to the `Builder` class.

### Specific Changes

1. **Discoverer.php**: Removed the `setDiscoveryState()` call, making the `discover()` method solely responsible for returning a `DiscoveryState` object.

2. **Builder.php**: Added handling and registration of the returned `DiscoveryState` object to the `performDiscovery()` method.

### Separation of Concerns

- **Before**: `Discoverer` could not return results in a cached state and did not call $registry->setDiscoveryState.
- **After**: `Discoverer` is solely responsible for discovering and returning results, while `Builder` coordinates the entire build process.

## Impact

- **Backward Compatibility**: No impact on external APIs.
- **Internal Improvements**: Improved code cohesion and maintainability.
- **Performance**: No performance impact, only code reorganization.

### DIFF
--- a/src/Capability/Registry/Loader/ArrayLoader.php
+++ b/src/Capability/Registry/Loader/ArrayLoader.php
@@ -1,0 +1,299 @@
+<?php
+
+/*
+ * This file is part of the official PHP MCP SDK.
+ *
+ * A collaboration between Symfony and the PHP Foundation.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Mcp\Capability\Registry\Loader;
+
+use Mcp\Capability\Attribute\CompletionProvider;
+use Mcp\Capability\Completion\EnumCompletionProvider;
+use Mcp\Capability\Completion\ListCompletionProvider;
+use Mcp\Capability\Completion\ProviderInterface;
+use Mcp\Capability\Discovery\DocBlockParser;
+use Mcp\Capability\Discovery\HandlerResolver;
+use Mcp\Capability\Discovery\SchemaGenerator;
+use Mcp\Capability\Registry\ElementReference;
+use Mcp\Capability\Registry\ReferenceRegistryInterface;
+use Mcp\Exception\ConfigurationException;
+use Mcp\Schema\Annotations;
+use Mcp\Schema\Prompt;
+use Mcp\Schema\PromptArgument;
+use Mcp\Schema\Resource;
+use Mcp\Schema\ResourceTemplate;
+use Mcp\Schema\Tool;
+use Mcp\Schema\ToolAnnotations;
+use Mcp\Server\Handler;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
+
+/**
+ * @author Antoine Bluchet <soyuka@gmail.com>
+ *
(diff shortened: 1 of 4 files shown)


chr-hertel:
Hey @okxaas, thanks for reaching out - not sure I understand how you ran into an error, but I still this change makes sense :+1: 

See #115 and especially #116 for what I changed to provide more insights.

Anyhow, like i wrote, makes sense still to me, but you need to address this one as well:
https://github.com/modelcontextprotocol/php-sdk/blob/6100ffcf3ff4b402b8f727844ce40e366f87fe9b/src/Capability/Discovery/Discoverer.php#L92-L101

cc @xentixar any thoughts on this?

i2mtams:
Hey @chr-hertel, with discovery caching enabled, the system was unable to utilize the cached content during actual runtime. Although the cache files were generated, the cached content couldn't be accessed. The issue was resolved after adding $registry->setDiscoveryState($discoveryState) to the performDiscovery method, which allowed the system to properly use the cached content.

i2mtams:
I read through #115 and #116 and the issue, I came up with was caused by cache generation and the discovery state of the cache not being registered.

xentixar:
Hey @chr-hertel, good catch! You're right - the PR is incomplete.

The refactoring removes `setDiscoveryState()` from the main path (line ~130) but leaves it at line 99 in the early-return path when no directories are found. For consistency, that call should also be removed:

**Line 92-101 should change from:**
```php
if (empty($absolutePaths)) {
    $this->logger->warning('No valid discovery directories found to scan.', [
        'configured_paths' => $directories,
        'base_path' => $basePath,
    ]);

    $emptyState = new DiscoveryState();
    $this->registry->setDiscoveryState($emptyState);

    return $emptyState;
}
```

**To:**
```php
if (empty($absolutePaths)) {
    $this->logger->warning('No valid discovery directories found to scan.', [
        'configured_paths' => $directories,
        'base_path' => $basePath,
    ]);

    return new DiscoveryState();
}
```

This way, `Builder::performDiscovery()` consistently handles state registration for all paths.

@okxaas - Can you update line 99 to complete the refactoring?

luoyue712:
hi, This issue occurs in `CachedDiscoverer.php` at lines 66-68, where there's an inconsistency with `Discoverer.php` lines 129-133. The problem is that `CachedDiscoverer.php` is missing the setDiscoveryState call.
```php
//  CachedDiscoverer.php
        $discoveryState = $this->discoverer->discover($basePath, $directories, $excludeDirs);

        $this->cache->set($cacheKey, $discoveryState);

        return $discoveryState;
```

```php
// Discoverer.php
        $discoveryState = new DiscoveryState($tools, $resources, $prompts, $resourceTemplates);

        $this->registry->setDiscoveryState($discoveryState);

        return $discoveryState;
```

i2mtams:
@lvluoyue Very correct expression @xentixar

chr-hertel (changes requested):
Understood & makes sense, still needs the patch here:
https://github.com/modelcontextprotocol/php-sdk/blob/6100ffcf3ff4b402b8f727844ce40e366f87fe9b/src/Capability/Discovery/Discoverer.php#L92-L101

Thanks already! :) :+1:

PR #115, opened 2025-10-23
[Server] Add more inspector tests for functionality tests of STDIO examples

chr-hertel:
More inspector based functional tests for STDIO examples - also renamed them to fit the new example names.


tests/Inspector/StdioCustomDependenciesTest.php
```diff
@@ -0,0 +1,29 @@
...
+    public static function provideMethods(): array
+    {
+        return [
+            ...parent::provideListMethods(),
+        ];
+    }
+
+    protected function getServerScript(): string
+    {
+        return \dirname(__DIR__, 2).'/examples/stdio-cached-discovery/server.php';
```
CodeWithKyrian: I'm sort of confused on why the same server script used for `StdioCachedDiscoveryTest` is also used here.

CodeWithKyrian: There's `examples/stdio-custom-dependencies/server.php` I think. That's what you meant to use right?

CodeWithKyrian (approved):

Issue #112, opened 2025-10-20
[Server] Make StreamableHttpTransport more configurable, e.g. for CORS headers


chr-hertel:
Currently the `StreamableHttpTransport` is explicit about the HTTP headers (e.g. for CORS) and lacks flexibility for configuring that behavior.

> Would it be possible to leave security features to the actual HTTP Proxy server that's running PHP (ie ngin/php/symfony)? Indeed this is hiding security settings and we've no ways of changing it.

_Originally posted by @soyuka in https://github.com/modelcontextprotocol/php-sdk/pull/110#discussion_r2439237545_

CodeWithKyrian:
Closed by #118

PR #118, opened 2025-10-25
[Server] Make CORS headers configurable in StreamableHttpTransport

CodeWithKyrian:
This PR introduces the ability to configure CORS headers in `StreamableHttpTransport`, addressing the opinionated and inflexible header implementation that was previously hardcoded.

## Motivation and Context
The current CORS implementation was explicitly set with fixed headers, limiting flexibility for different deployment scenarios. As raised in #112, this approach hid security settings and provided no way to adapt them when using reverse proxies like nginx or when specific CORS policies are required. 

## What's Changed
- Added optional `corsHeaders` parameter to `StreamableHttpTransport` constructor
- Secure default CORS headers are merged with any provided custom headers

## Breaking Changes

**Minor**: The logger parameter has shifted one position in the constructor signature.
If you’re passing it positionally, you’ll now need to include the new corsHeaders array parameter before it. Named arguments or cases where logger is omitted remain unaffected.

This change aligns the parameter order with how logger is handled consistently across the SDK.

### DIFF
--- a/examples/http-client-communication/server.php
+++ b/examples/http-client-communication/server.php
@@ -12,6 +12,7 @@
 require_once dirname(__DIR__).'/bootstrap.php';
 chdir(__DIR__);
 
+use Http\Discovery\Psr17Factory;
 use Laminas\HttpHandlerRunner\Emitter\SapiEmitter;
 use Mcp\Schema\Content\TextContent;
 use Mcp\Schema\Enum\LoggingLevel;
@@ -21,19 +22,16 @@ use Mcp\Server;
 use Mcp\Server\ClientGateway;
 use Mcp\Server\Session\FileSessionStore;
 use Mcp\Server\Transport\StreamableHttpTransport;
-use Nyholm\Psr7\Factory\Psr17Factory;
-use Nyholm\Psr7Server\ServerRequestCreator;
 
-$psr17Factory = new Psr17Factory();
-$creator = new ServerRequestCreator($psr17Factory, $psr17Factory, $psr17Factory, $psr17Factory);
-$request = $creator->fromGlobals();
+$request = (new Psr17Factory())->createServerRequestFromGlobals();
 
 $sessionDir = __DIR__.'/sessions';
 $capabilities = new ServerCapabilities(logging: true, tools: true);
+$logger = logger();
 
 $server = Server::builder()
     ->setServerInfo('HTTP Client Communication Demo', '1.0.0')
-    ->setLogger(logger())
+    ->setLogger($logger)
     ->setContainer(container())
     ->setSession(new FileSessionStore($sessionDir))
     ->setCapabilities($capabilities)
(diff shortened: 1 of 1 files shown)


chr-hertel:
Thanks for jumping on that @CodeWithKyrian :+1: 

@soyuka does that solve the issue for you?

src/Server/Transport/StreamableHttpTransport.php
```diff
@@ -41,20 +41,26 @@ class StreamableHttpTransport implements TransportInterface
...
         private readonly LoggerInterface $logger = new NullLogger(),
     ) {
         $sessionIdString = $this->request->getHeaderLine('Mcp-Session-Id');
         $this->sessionId = $sessionIdString ? Uuid::fromString($sessionIdString) : null;
+
+        $this->corsHeaders = array_merge([
+            'Access-Control-Allow-Origin' => '*',
+            'Access-Control-Allow-Methods' => 'GET, POST, DELETE, OPTIONS',
+            'Access-Control-Allow-Headers' => 'Content-Type, Mcp-Session-Id, Mcp-Protocol-Version, Last-Event-ID, Authorization, Accept',
+        ], $corsHeaders);
```
soyuka:
```suggestion
        $this->corsHeaders = $corsHeaders + [
            'Access-Control-Allow-Origin' => '*',
            'Access-Control-Allow-Methods' => 'GET, POST, DELETE, OPTIONS',
            'Access-Control-Allow-Headers' => 'Content-Type, Mcp-Session-Id, Mcp-Protocol-Version, Last-Event-ID, Authorization, Accept',
        ];
```

soyuka (commented):
minor comment

src/Server/Transport/StreamableHttpTransport.php
```diff
@@ -41,20 +41,26 @@ class StreamableHttpTransport implements TransportInterface
...
 
+    /**
+     * @param array<string, string> $corsHeaders
+     */
     public function __construct(
         private readonly ServerRequestInterface $request,
         private readonly ResponseFactoryInterface $responseFactory,
         private readonly StreamFactoryInterface $streamFactory,
+        array $corsHeaders = [],
         private readonly LoggerInterface $logger = new NullLogger(),
```
soyuka: any reason you don't set corsHeaders at the end to avoid breaking current implementations?

CodeWithKyrian: Yeah, wanted to keep things consistent with how the logger is passed across the SDK ie it’s always the last argument, and I think that consistency makes things more predictable overall. Since we haven’t tagged a release yet, this kind of breaking change is still fair game (for now).

chr-hertel: yup agree, we'd need to tag soon and care more than tho

chr-hertel (approved):
needs a rebase tho

src/Server/Transport/StreamableHttpTransport.php
CodeWithKyrian: Sure!

PR #111, opened 2025-10-18
[Server] refactor: registry loader

soyuka:
Refactor capability registration in Server Builder using dedicated loaders. A solution for https://github.com/modelcontextprotocol/php-sdk/issues/74

## Motivation and Context
The previous `Server/Builder.php` contained extensive logic for registering capabilities (tools, resources, prompts) both manually and through discovery. This led to a large and less modular `Builder` class. This change aims to improve the architecture by introducing dedicated loader classes for capability registration.

## How Has This Been Tested?
Existing tests for server capabilities and discovery should cover the functionality. No new tests were added as this is a refactoring. Let me know if I should add some as I'm not sure this is going to be accepted I didn't wanted to put unnecessary energy into unit testing. 

## Breaking Changes
No breaking changes are expected as this is an internal refactoring. The public API of the `Server/Builder` remains the same.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

## Additional context
This refactoring extracts the capability registration logic into `ArrayRegistryLoader` for manual registrations and `DiscoveryRegistryLoader` for discovery-based registrations. Both implement `RegistryLoaderInterface`. The `Server/Builder` now orchestrates these loaders, leading to a more extensible design.

PR #120, opened 2025-10-25
[Server] bugfix: Cache discovery is unavailable #113

okxaas:
## Problem Description

The cached discovery functionality was not working correctly because the `DiscoveryLoader` was not calling `$registry->setDiscoveryState($discoveryState)` when using cached results. This caused a critical issue where:

1. **Cache Hit Scenario**: When `CachedDiscoverer` returned cached `DiscoveryState` from the cache, the `DiscoveryLoader` would receive the cached state but never apply it to the registry via `setDiscoveryState()`.

2. **Registry State Mismatch**: The registry would remain empty or contain stale data, even though the discovery process had successfully retrieved cached results.

3. **Silent Failure**: This bug was particularly insidious because it would fail silently - the cache would appear to work (no errors thrown), but the discovered MCP elements (tools, resources, prompts, resource templates) would not be available in the registry.

## Root Cause Analysis

The issue occurred during the refactoring of the discovery process in PR #111. The original `Discoverer::discover()` method was responsible for both:
- Performing the discovery process
- Setting the discovery state on the registry via `$this->registry->setDiscoveryState($discoveryState)`

However, when the discovery logic was moved to `DiscoveryLoader`, the responsibility for setting the discovery state was not properly transferred. The `DiscoveryLoader` was calling the discoverer but not applying the returned `DiscoveryState` to the registry.

## Technical Details

**Before Fix:**
```php
// DiscoveryLoader.php - BROKEN
$cachedDiscoverer->discover($this->basePath, $this->scanDirs, $this->excludeDirs);
// Missing: $registry->setDiscoveryState($discoveryState);
```

**After Fix:**
```php
// DiscoveryLoader.php - FIXED
$discoveryState = $cachedDiscoverer->discover($this->basePath, $this->scanDirs, $this->excludeDirs);
$registry->setDiscoveryState($discoveryState);
```

## Impact

- **Performance**: Cached discovery was not providing the expected performance benefits
- **Functionality**: MCP servers using cached discovery would not expose their discovered capabilities
- **User Experience**: Tools, resources, and prompts would not be available to MCP clients

## Solution

This fix ensures that:
1. The `DiscoveryLoader` properly captures the `DiscoveryState` returned by the discoverer (whether cached or fresh)
2. The `DiscoveryState` is applied to the registry via `setDiscoveryState()`
3. Both cached and non-cached discovery paths work identically
4. The registry contains the correct discovered elements regardless of cache hit/miss

## Testing

The fix maintains backward compatibility and ensures that:
- Fresh discovery works as before
- Cached discovery now properly populates the registry
- All existing tests continue to pass
- The `stdio-cached-discovery` example now functions correctly

## Files Changed

- `src/Capability/Registry/Loader/DiscoveryLoader.php`: Added missing `setDiscoveryState()` call
- `src/Capability/Discovery/Discoverer.php`: Removed redundant `setDiscoveryState()` call (now handled by loader)

## Summary

This fix resolves the core issue where cached discovery results were not being applied to the registry, ensuring that MCP servers can properly utilize cached discovery for improved performance while maintaining full functionality.

---

**Type**: Bug Fix  
**Breaking Change**: No  
**Related Issue**: Fixes cached discovery functionality introduced in PR #111

### DIFF
--- a/src/Capability/Discovery/Discoverer.php
+++ b/src/Capability/Discovery/Discoverer.php
@@ -20,7 +20,6 @@ use Mcp\Capability\Completion\EnumCompletionProvider;
 use Mcp\Capability\Completion\ListCompletionProvider;
 use Mcp\Capability\Completion\ProviderInterface;
 use Mcp\Capability\Registry\PromptReference;
-use Mcp\Capability\Registry\ReferenceRegistryInterface;
 use Mcp\Capability\Registry\ResourceReference;
 use Mcp\Capability\Registry\ResourceTemplateReference;
 use Mcp\Capability\Registry\ToolReference;
@@ -48,7 +47,6 @@ use Symfony\Component\Finder\SplFileInfo;
 class Discoverer
 {
     public function __construct(
-        private readonly ReferenceRegistryInterface $registry,
         private readonly LoggerInterface $logger = new NullLogger(),
         private ?DocBlockParser $docBlockParser = null,
         private ?SchemaGenerator $schemaGenerator = null,
@@ -95,10 +93,7 @@ class Discoverer
                     'base_path' => $basePath,
                 ]);
 
-                $emptyState = new DiscoveryState();
-                $this->registry->setDiscoveryState($emptyState);
-
-                return $emptyState;
+                return new DiscoveryState();
             }
 
             $finder->files()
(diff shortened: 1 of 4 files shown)


i2mtams:
Due to the release of #111 , I updated the code related to this issue, the original PR #113

i2mtams:
Hey @chr-hertel  ,I have completed the code changes and submitted them to the specified branch

PR #122, opened 2025-10-27
Change visibility of StdioTransport methods to protected

luoyue712:
Hi, I need to extend `StdioTransport` because the framework I'm using has an event loop with stream read and write event listening capabilities, which can perfectly replace `while (!feof($this->input))`. However, some methods in `StdioTransport` are not `protected`, preventing me from implementing the required logic.

### DIFF
--- a/src/Server/Transport/StdioTransport.php
+++ b/src/Server/Transport/StdioTransport.php
@@ -75,7 +75,7 @@ class StdioTransport extends BaseTransport implements TransportInterface
         }
     }
 
-    private function processFiber(): void
+    protected function processFiber(): void
     {
         if (null === $this->sessionFiber) {
             return;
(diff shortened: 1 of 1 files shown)


chr-hertel:
How about decorating it or implementing a separate transport? which framework are you referring to?

luoyue712:
> How about decorating it or implementing a separate transport? which framework are you referring to?

The framework I mentioned is webman, which encapsulates different event loops into a unified interface, such as Swoole, Swow, and Fiber.

lvluoyue:
sorry,I found an issue in the above code. The initialization succeeds, but calling tools results in a timeout. Perhaps I should change it to use a similar handling approach as `StreamableHttpTransport`.

PR #124, opened 2025-10-27
[Server] Fix: standardize error handling across handlers with MCP specificatio…

CodeWithKyrian:
This PR standardizes error handling across all MCP handlers to comply with the MCP specification and provide better error visibility to clients while maintaining security.

## Motivation & Context
- MCP Specification Compliance: Tool execution errors were incorrectly returned as JSON-RPC errors instead of `CallToolResult` with `isError=true` as required by the [MCP specification](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling)
- Poor Error Visibility: Generic exception catching prevented actual error messages from reaching clients, making debugging difficult
- Unclear docs: Documentation didn't clearly specify which exceptions developers should throw in their handlers
- Inconsistent Patterns: Different handlers used different exception handling approaches

## What's Changed
- `CallToolHandler`: `ToolCallException` now returns Response with `CallToolResult(isError=true)` instead of JSON-RPC `Error`
- `ReadResourceHandler`: `ResourceReadException` now forwards actual exception messages to clients
- `GetPromptHandler`: `PromptGetException` now forwards actual exception messages to clients
- Exception Handling: All handlers now catch only specific domain exceptions, with generic exceptions returning generic error messages
- Documentation: Updated with explicit guidance on which exceptions to use for each handler type
- Updated tests for both domain-specific and generic exception scenarios
- **Exception-Driven Reference Provider**: `ReferenceProviderInterface` methods now throw `ToolNotFoundException`, `ResourceNotFoundException`, and `PromptNotFoundException` instead of returning null, with proper exception handling in all handlers
- **Simplified Exception Constructors**: Not found exceptions now accept simple string parameters (name/URI) instead of request objects, making them more reusable across the codebase

## Breaking Changes
Minor: `ToolCallException`, `ResourceReadException` and `PromptGetException` constructors now accept simple message strings instead of request objects. This aligns with documented usage patterns

### DIFF
--- a/src/Server/Handler/Request/CallToolHandler.php
+++ b/src/Server/Handler/Request/CallToolHandler.php
@@ -59,9 +59,6 @@ final class CallToolHandler implements RequestHandlerInterface
 
         try {
             $reference = $this->referenceProvider->getTool($toolName);
-            if (null === $reference) {
-                throw new ToolNotFoundException($request);
-            }
 
             $arguments['_session'] = $session;
 
--- a/src/Server/Handler/Request/GetPromptHandler.php
+++ b/src/Server/Handler/Request/GetPromptHandler.php
@@ -55,9 +55,6 @@ final class GetPromptHandler implements RequestHandlerInterface
 
         try {
             $reference = $this->referenceProvider->getPrompt($promptName);
-            if (null === $reference) {
-                throw new PromptNotFoundException($request);
-            }
 
             $arguments['_session'] = $session;
 
--- a/src/Capability/Registry.php
+++ b/src/Capability/Registry.php
@@ -23,6 +23,9 @@ use Mcp\Event\ResourceListChangedEvent;
 use Mcp\Event\ResourceTemplateListChangedEvent;
 use Mcp\Event\ToolListChangedEvent;
 use Mcp\Exception\InvalidCursorException;
+use Mcp\Exception\PromptNotFoundException;
+use Mcp\Exception\ResourceNotFoundException;
+use Mcp\Exception\ToolNotFoundException;
 use Mcp\Schema\Page;
 use Mcp\Schema\Prompt;
 use Mcp\Schema\Resource;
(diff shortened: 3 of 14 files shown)


src/Server/Handler/Request/GetPromptHandler.php
```diff
@@ -67,18 +66,18 @@ public function handle(Request $request, SessionInterface $session): Response|Er
...
-        } catch (PromptGetException|ExceptionInterface $e) {
-            $this->logger->error('Error while handling prompt', ['prompt_name' => $promptName]);
-
-            return Error::forInternalError('Error while handling prompt: '.$e->getMessage(), $request->getId());
+            return Error::forResourceNotFound($e->getMessage(), $request->getId());
         } catch (\Throwable $e) {
-            $this->logger->error('Error while handling prompt', ['prompt_name' => $promptName]);
+            $this->logger->error(\sprintf('Unexpected error while handling prompt "%s": "%s".', $promptName, $e->getMessage()));
 
-            return Error::forInternalError('Error while handling prompt: '.$e->getMessage(), $request->getId());
```
chr-hertel, line 81: yup, good decision - that was def too leaky :+1:

docs/mcp-elements.md
```diff
@@ -154,31 +154,37 @@ public function getMultipleContent(): array
 
 #### Error Handling
 
-Tools can throw exceptions which are automatically converted to proper JSON-RPC error responses:
+**Tools should ONLY throw `ToolCallException` for any error that occurs during tool execution.**
```
chr-hertel: that "should ONLY" confused me a bit - especially since we usually have that MoSCoW stuff

For example?
```suggestion
**Tools SHOULD throw `ToolCallException` for errors that occur during tool execution to be communicated to the client.**
```

docs/mcp-elements.md
```diff
@@ -154,31 +154,37 @@ public function getMultipleContent(): array
...
+        throw new ToolCallException("File not found: {$filename}");
     }
-    
+
     return file_get_contents($filename);
 }
 ```
 
-The SDK will convert these exceptions into appropriate JSON-RPC error responses that MCP clients can understand.
+**Critical Rule:** For tool handlers, **ALWAYS** throw `ToolCallException` for any error condition - validation errors, file not found, processing errors, etc. Do not use generic PHP exceptions.
```
chr-hertel: well, they can still throw kinda everything and the SDK will convert it to something JSON-RPC compliant, right?

devs should just use those explicit exceptions if they want to communicate something to clients

docs/mcp-elements.md
CodeWithKyrian: Ah. Fair point. The suggestion makes sense 👍🏼

docs/mcp-elements.md
CodeWithKyrian: Yeah.

CodeWithKyrian: How about now? Looks good?

src/Server/Handler/Request/CallToolHandler.php
```diff
@@ -77,17 +77,19 @@ public function handle(Request $request, SessionInterface $session): Response|Er
...
             ]);
 
-            return Error::forInternalError('Error while executing tool', $request->getId());
+            $errorContent = [new TextContent($e->getMessage())];
+
+            return new Response($request->getId(), CallToolResult::error($errorContent));
+        } catch (ToolNotFoundException $e) {
+            $this->logger->error('Tool not found', ['name' => $toolName]);
+
+            return new Error($request->getId(), Error::METHOD_NOT_FOUND, $e->getMessage());
```
chr-hertel, line 89: we could just go with an early exit in line 63 and make it simpler

chr-hertel: same for the other ones - or we move the not found exceptions into the providers instead of nullable return type.

CodeWithKyrian: > or we move the not found exceptions into the providers instead of nullable return type.

This makes more sense than the early exit, since it allows us still enforce the return types for Request handlers (either a Response or an Error and nothing else). Will update the reference provider interface and implementation. Thanks!

docs/mcp-elements.md
chr-hertel: yes - this is really good to understand now :+1:

PR #128, opened 2025-10-30
[Examples][Server] transport agnostic examples

chr-hertel:
@CodeWithKyrian is this what we want?

i think in combination with the docs on how to instantiate the two transports, that's fine and makes it easier for us

i left the tests like they were for now - easier also to proof that this works as before :+1:


CodeWithKyrian:
This is exactly what we want! Thanks @chr-hertel 

We may have to cut down on the examples now that its unified..maybe later

PR #129, opened 2025-10-31
Add _meta Field

kaipiyann:
Add _meta field to schema

## Motivation and Context
Field meta from the spec is missing 
https://modelcontextprotocol.io/specification/2025-06-18/basic#meta

## How Has This Been Tested?
Tested with MCP Inspector

## Breaking Changes
No

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/Capability/Attribute/McpPrompt.php
+++ b/src/Capability/Attribute/McpPrompt.php
@@ -21,12 +21,14 @@ namespace Mcp\Capability\Attribute;
 final class McpPrompt
 {
     /**
-     * @param ?string $name        overrides the prompt name (defaults to method name)
-     * @param ?string $description Optional description of the prompt. Defaults to method DocBlock summary.
+     * @param ?string               $name        overrides the prompt name (defaults to method name)
+     * @param ?string               $description Optional description of the prompt. Defaults to method DocBlock summary.
+     * @param ?array<string, mixed> $meta        Optional metadata
      */
     public function __construct(
         public ?string $name = null,
         public ?string $description = null,
+        public ?array $meta = null,
     ) {
     }
 }
--- a/src/Schema/Prompt.php
+++ b/src/Schema/Prompt.php
@@ -22,6 +22,7 @@ use Mcp\Exception\InvalidArgumentException;
  *     name: string,
  *     description?: string,
  *     arguments?: PromptArgumentData[],
+ *     _meta?: array<string, mixed>
  * }
  *
  * @author Kyrian Obikwelu <koshnawaza@gmail.com>
@@ -32,11 +33,13 @@ class Prompt implements \JsonSerializable
      * @param string                $name        the name of the prompt or prompt template
      * @param string|null           $description an optional description of what this prompt provides
      * @param PromptArgument[]|null $arguments   A list of arguments for templating. Null if not a template.
+     * @param ?array<string, mixed> $meta        Optional metadata
      */
     public function __construct(
         public readonly string $name,
         public readonly ?string $description = null,
         public readonly ?array $arguments = null,
+        public readonly ?array $meta = null,
     ) {
         if (null !== $this->arguments) {
             foreach ($this->arguments as $arg) {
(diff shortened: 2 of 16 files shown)


src/Capability/Attribute/McpPrompt.php
```diff
@@ -23,10 +23,12 @@ final class McpPrompt
     /**
      * @param ?string $name        overrides the prompt name (defaults to method name)
      * @param ?string $description Optional description of the prompt. Defaults to method DocBlock summary.
-     */
+     * @param ?array  $_meta       Optional metadata
+ */
     public function __construct(
         public ?string $name = null,
         public ?string $description = null,
+        public ?array $_meta = null
```
CodeWithKyrian: You could name the PHP variables `$meta` instead to stay consistent with how meta is handled in the `Request` class, and then when serializing to and from array, use the `_meta` key where neccesary

src/Schema/Prompt.php
```diff
@@ -63,7 +65,8 @@ public static function fromArray(array $data): self
         return new self(
             name: $data['name'],
             description: $data['description'] ?? null,
-            arguments: $arguments
+            arguments: $arguments,
+           _meta: isset($data['_meta']) ? (int) $data['_meta'] : null
```
CodeWithKyrian: Looks like the `(int)` cast might be a typo here.

Also, we should validate that `$data['_meta']` is an array when it’s not null, similar to how we already handle name and arguments. If it isn’t, then we throw a clear exception to prevent silent type issues later.

src/Schema/ResourceTemplate.php
```diff
@@ -80,7 +82,8 @@ public static function fromArray(array $data): self
             name: $data['name'],
             description: $data['description'] ?? null,
             mimeType: $data['mimeType'] ?? null,
-            annotations: isset($data['annotations']) ? Annotations::fromArray($data['annotations']) : null
+            annotations: isset($data['annotations']) ? Annotations::fromArray($data['annotations']) : null,
+            _meta: isset($data['_meta']) ? (int) $data['_meta'] : null
```
CodeWithKyrian: Same int cast issue here

CodeWithKyrian (changes requested):

src/Capability/Attribute/McpPrompt.php
kaipiyann: Yes I could have but : 
- it seems to me it was inconsistent before cause ListRootsResult and InitializeResult are using _meta 
- I thought it was easier to be consistent with the naming of the shema
- And now there is a lot more _meta than meta ;-)
I hope you don't mind

src/Schema/Prompt.php
kaipiyann: Ooops yes my fault, corrected and validation added.

src/Schema/ResourceTemplate.php
kaipiyann: Corrected also

kaipiyann:
And this time I did run all the tests also so it should pass the check.

src/Capability/Attribute/McpPrompt.php
chr-hertel: I agree with @CodeWithKyrian here, so let's fix that with #134

kaipiyann: @chr-hertel That's why I didn't want to start a PR... But I did it and hopefully you will be able to merge.

chr-hertel:
I struggle a bit with adding `meta` also to the attributes - i consider meta rather decoupled from the core schema elements. What was your reasoning here?

kaipiyann:
@chr-hertel I just need it. I tried to make MCP widget for Openai so I need to be able to define _meta for tool and ressource. 
Doc : https://developers.openai.com/apps-sdk/build/mcp-server

chr-hertel (approved):
Works for me - thanks :+1:

chr-hertel:
@kaipiyann please squash your changes to a single signed commit

kaipiyann:
@chr-hertel : Signed and squashed !

PR #132, opened 2025-11-04
[Server] ability to append loaders to server builder

soyuka:
To be able to use registry loaders we need this patch. 

See also https://github.com/modelcontextprotocol/php-sdk/pull/111


chr-hertel (approved):
Thanks @soyuka!

PR #133, opened 2025-11-04
[Server] Update session save to return success flag

soyuka:
Suggestion to change the return type of `SessionInterface::save` so that we can know whether it did save or not.

Let me know your thoughts.


chr-hertel:
Thank you @soyuka!

PR #135, opened 2025-11-07
[Server] Fix blob resource referencing itself while serializing instead of parent

### LINKED ISSUE #130: Strange Code
Working on the SDK I read  this method in BlobResourceContents.php and wonder if there is not a problem here :

```
    public function jsonSerialize(): array
    {
        return [
            'blob' => $this->blob,
            ...$this->jsonSerialize(),
        ];
    }
```
https://github.com/modelcontextprotocol/php-sdk/blob/757b959bfd2a0842c6af49e5d546e45426aea686/src/Schema/Content/BlobResourceContents.php#L78C1-L84C6

In TextResourceContents.php the method call parent :

```
    public function jsonSerialize(): array
    {
        return [
            'blob' => $this->blob,
            ...parent::jsonSerialize(),
        ];
    }
```

chr-hertel:
Fix #130

### DIFF
--- a/src/Schema/Content/BlobResourceContents.php
+++ b/src/Schema/Content/BlobResourceContents.php
@@ -79,7 +79,7 @@ class BlobResourceContents extends ResourceContents
     {
         return [
             'blob' => $this->blob,
-            ...$this->jsonSerialize(),
+            ...parent::jsonSerialize(),
         ];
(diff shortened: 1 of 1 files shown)


chr-hertel:
Is also part of #129

PR #140, opened 2025-11-07
[Server] Add tool name validation for SEP-986

### LINKED ISSUE #137: [Server][SEP-986] Specify Format for Tool Names
Issue to address and track [SEP-986](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/986):

> The Model Context Protocol (MCP) currently lacks a standardized format for tool names, resulting in inconsistencies and confusion for both implementers and users. This SEP proposes a clear, flexible standard for tool names: tool names should be 1–64 characters, case-sensitive, and may include alphanumeric characters, underscores (_), dashes (-), dots (.), and forward slashes (/). This aims to maximize compatibility, clarity, and interoperability across MCP implementations while accommodating a wide range of naming conventions.

Reference Implementation: https://github.com/modelcontextprotocol/csharp-sdk/pull/945

chr-hertel:
Close #137

### DIFF
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -15,6 +15,10 @@ parameters:
     ignoreErrors:
         -
             message: "#^Method .*::test.*\\(\\) has no return type specified\\.$#"
+        -
+            identifier: missingType.iterableValue
+            path: tests/
+
         # These errors should actually be fixed, but are ignored for now
         -
             identifier: missingType.iterableValue
--- a/src/Capability/Registry.php
+++ b/src/Capability/Registry.php
@@ -18,6 +18,7 @@ use Mcp\Capability\Registry\ReferenceRegistryInterface;
 use Mcp\Capability\Registry\ResourceReference;
 use Mcp\Capability\Registry\ResourceTemplateReference;
 use Mcp\Capability\Registry\ToolReference;
+use Mcp\Capability\Tool\NameValidator;
 use Mcp\Event\PromptListChangedEvent;
 use Mcp\Event\ResourceListChangedEvent;
 use Mcp\Event\ResourceTemplateListChangedEvent;
@@ -71,6 +72,7 @@ final class Registry implements ReferenceProviderInterface, ReferenceRegistryInt
     public function __construct(
         private readonly ?EventDispatcherInterface $eventDispatcher = null,
         private readonly LoggerInterface $logger = new NullLogger(),
+        private readonly NameValidator $nameValidator = new NameValidator(),
     ) {
     }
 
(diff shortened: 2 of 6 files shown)


src/Capability/Registry.php
```diff
@@ -100,12 +102,18 @@ public function registerTool(Tool $tool, callable|array|string $handler, bool $i
...
+                \sprintf('Ignoring discovered tool "%s" as it conflicts with a manually registered one.', $toolName),
             );
 
             return;
         }
 
+        if (!$this->nameValidator->isValid($toolName)) {
+            $this->logger->warning(
+                \sprintf('Tool name "%s" is invalid. Tool names should only contain letters from a-z, numbers, dots, hyphens, underscores, and forward slashes.', $toolName),
+            );
```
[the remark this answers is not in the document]
chr-hertel: yup, that's fine - we only raise a warning, but it keeps working - that's my understanding of the SEP

phpstan.dist.neon
```diff
@@ -15,6 +15,10 @@ parameters:
     ignoreErrors:
         -
             message: "#^Method .*::test.*\\(\\) has no return type specified\\.$#"
+        -
+            identifier: missingType.iterableValue
+            path: tests/
+
```
chr-hertel, line 21: this is addressing array shapes for DataProvider methods in tests - i don't think there is any value in defining them since it is only PHPUnit consuming those arrays not user land code.

src/Capability/Tool/NameValidator.php
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * A collaboration between Symfony and the PHP Foundation.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Mcp\Capability\Tool;
+
+final class NameValidator
```
CodeWithKyrian, line 14: I’m not sure we need to extract name validation into a separate class here. Tool names must follow a hard rule according the spec and not a user-configurable validation. So having dedicated `NameValidator` feels a bit unnecessary, especially since we don’t expect custom implementations of that validation.

Also, if we introduce a standalone validator for tool names, then for consistency we’d have to do the same for resource name validation (`RFC3986`), and resource template URI validation (`RFC6570`). But in those cases, the validation currently lives in the DTO constructors themselves, and that has worked well so far.

Since `Tool`, `Resource`, and `ResourceTemplate` are all DTOs, it feels more natural for `Tool` to validate its own name in the constructor, just like the others do. That way each object ensures its own invariants and we avoid adding more moving parts unless there’s a real need for pluggability (is there?)

Wdyt?

chr-hertel: yea, basically it was easiest since i wanted to do it test driven with those specs. and generally i'm totally in favor of small dedicated classes. but i get the OOP thinking of having it in the constructor and i thought about that, but for now it's not a hard requirement (if i got the SEP item correctly) but should rather only emit a warning.

we could make it a hard requirement tho and go one step further and just throw an exception in controller.

CodeWithKyrian (approved):

PR #141, opened 2025-11-07
[Schema][Server] Add support for icons and website url

### LINKED ISSUE #136: [Server][SEP-973] Expose additional metadata for Implementations, Resources, Tools and Prompts
Issue to address and track [SEP-973](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/973):

> This SEP proposes adding two optional fields—icons and websiteUrl. The icons and websiteUrl would be added to the Implementation schema so that clients can visually identify third-party implementations and link directly to their documentation. The icons parameter will also be added to the Tool, Resource and Prompt schemas. While this can be used by both servers and clients for all implementations, we expect it to be used initially for server-provided implementations.

Spec Change: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/955/files
Reference Impl: https://github.com/modelcontextprotocol/typescript-sdk/pull/904/files

chr-hertel:
Close #136

### DIFF
--- a/examples/schema-showcase/server.php
+++ b/examples/schema-showcase/server.php
@@ -13,11 +13,18 @@
 require_once dirname(__DIR__).'/bootstrap.php';
 chdir(__DIR__);
 
+use Mcp\Schema\Icon;
 use Mcp\Server;
 use Mcp\Server\Session\FileSessionStore;
 
 $server = Server::builder()
-    ->setServerInfo('Schema Showcase', '1.0.0')
+    ->setServerInfo(
+        'Schema Showcase',
+        '1.0.0',
+        'A showcase server demonstrating MCP schema capabilities.',
+        [new Icon('https://www.php.net/images/logos/php-logo-white.svg', 'image/svg+xml', ['any'])],
+        'https://github.com/modelcontextprotocol/php-sdk',
+    )
     ->setContainer(container())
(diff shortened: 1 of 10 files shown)


luoyue712:
I discovered that the lack of Attribute support will prevent obtaining the tools list.
```
Typed property Mcp\\Schema\\Tool::$icons must not be accessed before initialization
```
I have an idea: we can extract the icon into an annotation (meta should also be able to use this method), not sure if it's feasible.

chr-hertel:
1. you are right, the attributes should support this as well
2. how did you run into the error? cannot reproduce with an example :thinking:

luoyue712:
> 1. you are right, the attributes should support this as well
> 2. how did you run into the error? cannot reproduce with an example 🤔

I used Discovery Cache, and after clearing the cache, it worked normally. It seems that the cache was not refreshed, which caused the issue.

luoyue712:
> 1. you are right, the attributes should support this as well
> 2. how did you run into the error? cannot reproduce with an example 🤔

I used Discovery Cache, and after clearing the cache, it worked normally. It seems that the cache was not refreshed, which caused the issue.

chr-hertel:
That cache hiccup makes sense & and #142 should address your finding

PR #142, opened 2025-11-09
[Server] Add support for icons on explicit builder registration and attributes

chr-hertel:
Follow #141 

needs some testing still => decided on reusing existing examples of bringing in more


CodeWithKyrian:
Great stuff, @chr-hertel.
Any chance you could include the docs update in this PR too?

PR #146, opened 2025-11-11
[Server] Add support on Builder to inject Registry instance

chr-hertel:
Not exactly ideal and doesn't solve the trouble i have with that service slicing at this point, but would enable the concerns of #74, doesn't it? @camilleislasse

cc @soyuka

### DIFF
--- a/src/Capability/Registry.php
+++ b/src/Capability/Registry.php
@@ -13,8 +13,6 @@ namespace Mcp\Capability;
 
 use Mcp\Capability\Discovery\DiscoveryState;
 use Mcp\Capability\Registry\PromptReference;
-use Mcp\Capability\Registry\ReferenceProviderInterface;
-use Mcp\Capability\Registry\ReferenceRegistryInterface;
 use Mcp\Capability\Registry\ResourceReference;
 use Mcp\Capability\Registry\ResourceTemplateReference;
 use Mcp\Capability\Registry\ToolReference;
@@ -44,7 +42,7 @@ use Psr\Log\NullLogger;
  * @author Kyrian Obikwelu <koshnawaza@gmail.com>
  * @author Pavel Buchnev   <butschster@gmail.com>
  */
-final class Registry implements ReferenceProviderInterface, ReferenceRegistryInterface
+final class Registry implements RegistryInterface
 {
     /**
      * @var array<string, ToolReference>
@@ -91,6 +89,11 @@ final class Registry implements ReferenceProviderInterface, ReferenceRegistryInt
             logging: false,
             completions: true,
         );
+    }
+
+    public function setServerCapabilities(ServerCapabilities $serverCapabilities): void
+    {
+        $this->serverCapabilities = $serverCapabilities;
     }
 
     public function registerTool(Tool $tool, callable|array|string $handler, bool $isManual = false): void
@@ -455,10 +458,5 @@ final class Registry implements ReferenceProviderInterface, ReferenceRegistryInt
         }
 
         return array_values(\array_slice($items, $offset, $limit));
-    }
-
-    public function setServerCapabilities(ServerCapabilities $serverCapabilities): void
-    {
-        $this->serverCapabilities = $serverCapabilities;
     }
 }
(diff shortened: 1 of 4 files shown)


camilleislasse:
Hi @chr-hertel

I've been working on the Symfony profiler integration (the use case from #74), and I found a solution using the loader pattern that @soyuka suggested in PR #75.

My approach:

I created a ProfilingLoader that implements LoaderInterface to capture the Registry reference:

```php
  final class ProfilingLoader implements LoaderInterface
  {
      private ?ReferenceRegistryInterface $registry = null;

      public function load(ReferenceRegistryInterface $registry): void
      {
          // Capture the registry reference for profiling
          $this->registry = $registry;
      }

      public function getRegistry(): ?ReferenceRegistryInterface
      {
          return $this->registry;
      }
  }
```

Then register it in the Symfony container and add it to the ServerBuilder:

```php
  ->set('ai.mcp.profiling_loader', ProfilingLoader::class)

  ->set('mcp.server.builder', ServerBuilder::class)
      ->factory([Server::class, 'builder'])
      // ... 
      ->call('addLoaders', [service('ai.mcp.profiling_loader')])
```

The Symfony DataCollector then reads from this ProfilingLoader in lateCollect():

```php
  public function lateCollect(): void
  {
      $registry = $this->profilingLoader->getRegistry();

      if (null === $registry) {
          // Handle case where Registry wasn't built yet
          return;
      }

      // Collect tools, prompts, resources, templates
      foreach ($registry->getTools()->references as $tool) {
          // ...
      }
  }
```

This approach solves my use case. If you think your PR #146 is "not exactly ideal" and this loader approach (following the pattern from #111) suits you, maybe this PR becomes less necessary? I'll let you and @soyuka decide what's best 🙂

chr-hertel:
yea, i can see that working, you really dug for a solution & found one👍 

i think this PR would enable us to manage the `Registry` service via dependency injection container and have a traceable decorator for the profiler

camilleislasse:
> yea, i can see that working, you really dug for a solution & found one👍
> 
> i think this PR would enable us to manage the `Registry` service via dependency injection container and have a traceable decorator for the profiler

Would you recommend I wait for this PR to be merged before refactoring profiler to use `setRegistry()` instead?

src/Capability/Registry.php
```diff
@@ -93,6 +91,11 @@ public function getCapabilities(): ServerCapabilities
         );
     }
 
+    public function setServerCapabilities(ServerCapabilities $serverCapabilities): void
+    {
+        $this->serverCapabilities = $serverCapabilities;
+    }
+
```
soyuka, line 98: It'd be even cleaner to have the ServerCapabilities directly in dependency injection. 

```php
return $this->serverCapabilities ?? new ServerCapabilities(
            tools: [] !== $this->tools,
            toolsListChanged: $this->eventDispatcher instanceof EventDispatcherInterface,
            resources: [] !== $this->resources || [] !== $this->resourceTemplates,
            resourcesSubscribe: false,
            resourcesListChanged: $this->eventDispatcher instanceof EventDispatcherInterface,
            prompts: [] !== $this->prompts,
            logging: false,
            completions: true,
        );
```

This would allow more flexibility in my opinion. Loaders could then prepare tools, resources etc. without having to register them at runtime. In the end, they could more easily be cache-able. I'd still keep register methods for runtime changes.

src/Capability/RegistryInterface.php
```diff
@@ -0,0 +1,23 @@
...
+
+use Mcp\Capability\Registry\ReferenceProviderInterface;
+use Mcp\Capability\Registry\ReferenceRegistryInterface;
+use Mcp\Schema\ServerCapabilities;
+
+interface RegistryInterface extends ReferenceRegistryInterface, ReferenceProviderInterface
+{
+    public function getCapabilities(): ServerCapabilities;
+
+    public function setServerCapabilities(ServerCapabilities $serverCapabilities): void;
```
soyuka, line 22:
```suggestion

```

chr-hertel:
closed in favor of #146

PR #150, opened 2025-11-14
[Server] Rework registry to combine provider and registry interface, move capabilities out and enable Registry injection in Builder

### LINKED ISSUE #74: [Server] Expose Server Registry for debugging and profiling tools

chr-hertel:
Alternative to #146 

Outer interface does not change => no docs changes required

Fixes #74


src/Capability/RegistryInterface.php
```diff
@@ -0,0 +1,165 @@
...
+     * Gets a prompt reference by name.
+     *
+     * @throws PromptNotFoundException
+     */
+    public function getPrompt(string $name): PromptReference;
+
+    /**
+     * Checks if any elements (manual or discovered) are currently registered.
+     */
+    public function hasElements(): bool;
```
CodeWithKyrian: since we now have `hasTools`, `hasResources`, etc, I don't think this is necessary anymore

chr-hertel: removed :heavy_check_mark:

src/Server/Builder.php
```diff
@@ -455,16 +466,12 @@ public function build(): Server
...
-            $registry->setServerCapabilities($this->serverCapabilities);
-        }
-
         $sessionTtl = $this->sessionTtl ?? 3600;
         $sessionFactory = $this->sessionFactory ?? new SessionFactory();
         $sessionStore = $this->sessionStore ?? new InMemorySessionStore($sessionTtl);
         $messageFactory = MessageFactory::make();
 
-        $capabilities = $registry->getCapabilities();
+        $capabilities = $this->serverCapabilities ?? ServerCapabilitiesFactory::create($registry, $this->eventDispatcher instanceof EventDispatcherInterface);
```
CodeWithKyrian: Is there a strong reason this needs its own class (The `ServerCapabilitiesFactory` I mean?

Since we’re already assembling most services inline in the `build()` method (handlers, loaders, etc), part of me wonders if this could follow the same pattern instead of introducing one more abstraction.

Not saying it's wrong, just checking whether there's a clear benefit here, or if I'm just being allergic to new classes 😅

chr-hertel: tbh, i feel like we have quite some huge classes - `Builder` and `Registry` are good examples. and in general i'd like to decouple/encapsulate stuff a bit more.
felt like there was an opportunity to make the Builder and Registry smaller with moving it to a factory
- no getter/setter in registry -> less state
- less logic in Builder
- on top this could enable what @soyuka raised in #146 i guess

chr-hertel: move it into Builder completely

chr-hertel:
@soyuka & @camilleislasse we would merge that instead of #75 and #146 - does that work for the use-cases you have in mind?

src/Capability/Registry.php
```diff
@@ -279,6 +220,16 @@ public function getTools(?int $limit = null, ?string $cursor = null): Page
         return new Page($paginatedTools, $nextCursor);
     }
 
+    public function getTool(string $name): ToolReference
+    {
+        return $this->tools[$name] ?? throw new ToolNotFoundException($name);
+    }
```
soyuka, line 226: it'd be nice for reviewing to keep the previous order of methods :D.

src/Capability/Registry.php
```diff
@@ -301,26 +252,31 @@ public function getResources(?int $limit = null, ?string $cursor = null): Page
...
 
-        $paginatedPrompts = $this->paginateResults($prompts, $limit, $cursor);
+        $this->logger->debug('No resource matched URI.', ['uri' => $uri]);
 
-        $nextCursor = $this->calculateNextCursor(
-            \count($prompts),
-            $cursor,
-            $limit
-        );
+        throw new ResourceNotFoundException($uri);
```
soyuka, line 274: weird to log and to throw ?

src/Capability/Registry.php
```diff
@@ -301,26 +252,31 @@ public function getResources(?int $limit = null, ?string $cursor = null): Page
...
         }
 
-        if (null === $limit) {
-            return new Page($prompts, null);
+        if ($includeTemplates) {
+            foreach ($this->resourceTemplates as $template) {
+                if ($template->matches($uri)) {
+                    return $template;
+                }
+            }
```
soyuka, line 269:
```suggestion
        if (!$includeTemplates) {
            throw new ResourceNotFoundException($uri);
        }
        
        foreach ($this->resourceTemplates as $template) {
            if ($template->matches($uri)) {
                return $template;
            }
        }
```

src/Capability/Registry.php
chr-hertel: yea fair, i think we're currently a bit rough with PR scopes :grimacing:

PR #153, opened 2025-11-16
feat: Add output schema support to MCP tools

bigdevlarry:
The changes add comprehensive support for outputSchema in the Tools definition, enabling MCP servers to specify the expected structure and types of tool responses.

This includes:

Added outputSchema parameter to Builder::addTool() method
Updated the Tool class constructor and JSON serialization to support outputSchema
Added validation for outputSchema structure and type requirements
Enhanced existing tests to verify outputSchema functionality works correctly
Fixed related type annotation issues and code style problems

Screenshot 2025-12-06 at 23 13 03
Screenshot 2025-12-06 at 23 12 44

## Types of changes

- [ ] Documentation update

## Checklist

- [x ] I have added appropriate error handling

## Additional context

This will fix https://github.com/modelcontextprotocol/php-sdk/issues/68

### DIFF
--- a/src/Capability/Discovery/SchemaGenerator.php
+++ b/src/Capability/Discovery/SchemaGenerator.php
@@ -11,6 +11,7 @@
 
 namespace Mcp\Capability\Discovery;
 
+use Mcp\Capability\Attribute\McpTool;
 use Mcp\Capability\Attribute\Schema;
 use Mcp\Exception\InvalidArgumentException;
 use Mcp\Server\RequestContext;
@@ -79,6 +80,27 @@ class SchemaGenerator
         $parametersInfo = $this->parseParametersInfo($reflection);
 
         return $this->buildSchemaFromParameters($parametersInfo, $methodSchema);
+    }
+
+    /**
+     * Generates a JSON Schema object (as a PHP array) for a method's or function's return type.
+     *
+     * Only returns an outputSchema if explicitly provided in the McpTool attribute.
+     * Per MCP spec, outputSchema should only be present when explicitly provided.
+     *
+     * @return array<string, mixed>|null
+     */
+    public function generateOutputSchema(\ReflectionMethod|\ReflectionFunction $reflection): ?array
+    {
+        // Only return outputSchema if explicitly provided in McpTool attribute
+        $mcpToolAttrs = $reflection->getAttributes(McpTool::class, \ReflectionAttribute::IS_INSTANCEOF);
+        if ($mcpToolAttrs) {
+            $mcpToolInstance = $mcpToolAttrs[0]->newInstance();
+
+            return $mcpToolInstance->outputSchema;
+        }
+
+        return null;
     }
 
     /**
--- a/src/Capability/Registry/ToolReference.php
+++ b/src/Capability/Registry/ToolReference.php
@@ -111,4 +111,33 @@ class ToolReference extends ElementReference
 
         return [new TextContent($jsonResult)];
     }
+
+    /**
+     * Extracts structured content from a tool result using the output schema.
+     *
+     * @param mixed $toolExecutionResult the raw value returned by the tool's PHP method
+     *
+     * @return array<string, mixed>|null the structured content, or null if not extractable
+     *
+     * @throws \JsonException if JSON encoding fails for non-Content array/object results
+     */
+    public function extractStructuredContent(mixed $toolExecutionResult): ?array
+    {
+        if (\is_array($toolExecutionResult)) {
+            return $toolExecutionResult;
+        }
+
+        if (\is_object($toolExecutionResult) && !($toolExecutionResult instanceof Content)) {
+            $jsonResult = json_encode(
+                $toolExecutionResult,
+                \JSON_PRETTY_PRINT | \JSON_UNESCAPED_SLASHES | \JSON_UNESCAPED_UNICODE | \JSON_THROW_ON_ERROR | \JSON_INVALID_UTF8_SUBSTITUTE
+            );
+
+            return json_decode(
+                $jsonResult, true, 512, \JSON_THROW_ON_ERROR
+            );
+        }
+
+        return null;
+    }
 }
--- a/src/Schema/Result/CallToolResult.php
+++ b/src/Schema/Result/CallToolResult.php
@@ -83,6 +83,7 @@ class CallToolResult implements ResultInterface
      *     content: array<mixed>,
      *     isError?: bool,
      *     _meta?: array<string, mixed>,
+     *     structuredContent?: array<string, mixed>
      * } $data
      */
     public static function fromArray(array $data): self
--- a/src/Schema/Tool.php
+++ b/src/Schema/Tool.php
@@ -24,13 +24,21 @@ use Mcp\Exception\InvalidArgumentException;
  *     properties: array<string, mixed>,
  *     required: string[]|null
  * }
+ * @phpstan-type ToolOutputSchema array{
+ *     type: 'object',
+ *     properties?: array<string, mixed>,
+ *     required?: string[]|null,
+ *     additionalProperties?: bool|array<string, mixed>,
+ *     description?: string
+ * }
  * @phpstan-type ToolData array{
  *     name: string,
  *     inputSchema: ToolInputSchema,
  *     description?: string|null,
  *     annotations?: ToolAnnotationsData,
  *     icons?: IconData[],
- *     _meta?: array<string, mixed>
+ *     _meta?: array<string, mixed>,
+ *     outputSchema?: ToolOutputSchema
  * }
  *
  * @author Kyrian Obikwelu <koshnawaza@gmail.com>
@@ -38,14 +46,15 @@ use Mcp\Exception\InvalidArgumentException;
 class Tool implements \JsonSerializable
 {
     /**
-     * @param string                $name        the name of the tool
-     * @param ?string               $description A human-readable description of the tool.
-     *                                           This can be used by clients to improve the LLM's understanding of
-     *                                           available tools. It can be thought of like a "hint" to the model.
-     * @param ToolInputSchema       $inputSchema a JSON Schema object (as a PHP array) defining the expected 'arguments' for the tool
-     * @param ?ToolAnnotations      $annotations optional additional tool information
-     * @param ?Icon[]               $icons       optional icons representing the tool
-     * @param ?array<string, mixed> $meta        Optional metadata
+     * @param string                $name         the name of the tool
+     * @param ?string               $description  A human-readable description of the tool.
+     *                                            This can be used by clients to improve the LLM's understanding of
+     *                                            available tools. It can be thought of like a "hint" to the model.
+     * @param ToolInputSchema       $inputSchema  a JSON Schema object (as a PHP array) defining the expected 'arguments' for the tool
+     * @param ?ToolAnnotations      $annotations  optional additional tool information
+     * @param ?Icon[]               $icons        optional icons representing the tool
+     * @param ?array<string, mixed> $meta         Optional metadata
+     * @param ToolOutputSchema|null $outputSchema optional JSON Schema object (as a PHP array) defining the expected output structure
      */
     public function __construct(
         public readonly string $name,
@@ -54,6 +63,7 @@ class Tool implements \JsonSerializable
         public readonly ?ToolAnnotations $annotations,
         public readonly ?array $icons = null,
         public readonly ?array $meta = null,
+        public readonly ?array $outputSchema = null,
     ) {
         if (!isset($inputSchema['type']) || 'object' !== $inputSchema['type']) {
             throw new InvalidArgumentException('Tool inputSchema must be a JSON Schema of type "object".');
@@ -78,13 +88,23 @@ class Tool implements \JsonSerializable
             $data['inputSchema']['properties'] = new \stdClass();
         }
 
+        if (isset($data['outputSchema']) && \is_array($data['outputSchema'])) {
+            if (!isset($data['outputSchema']['type']) || 'object' !== $data['outputSchema']['type']) {
+                throw new InvalidArgumentException('Tool outputSchema must be of type "object".');
+            }
+            if (isset($data['outputSchema']['properties']) && \is_array($data['outputSchema']['properties']) && empty($data['outputSchema']['properties'])) {
+                $data['outputSchema']['properties'] = new \stdClass();
+            }
+        }
+
         return new self(
             $data['name'],
             $data['inputSchema'],
             isset($data['description']) && \is_string($data['description']) ? $data['description'] : null,
             isset($data['annotations']) && \is_array($data['annotations']) ? ToolAnnotations::fromArray($data['annotations']) : null,
             isset($data['icons']) && \is_array($data['icons']) ? array_map(Icon::fromArray(...), $data['icons']) : null,
-            isset($data['_meta']) && \is_array($data['_meta']) ? $data['_meta'] : null
+            isset($data['_meta']) && \is_array($data['_meta']) ? $data['_meta'] : null,
+            isset($data['outputSchema']) && \is_array($data['outputSchema']) ? $data['outputSchema'] : null,
         );
     }
 
@@ -95,7 +115,8 @@ class Tool implements \JsonSerializable
      *     description?: string,
      *     annotations?: ToolAnnotations,
      *     icons?: Icon[],
-     *     _meta?: array<string, mixed>
+     *     _meta?: array<string, mixed>,
+     *     outputSchema?: ToolOutputSchema
      * }
      */
     public function jsonSerialize(): array
@@ -115,6 +136,9 @@ class Tool implements \JsonSerializable
         }
         if (null !== $this->meta) {
             $data['_meta'] = $this->meta;
+        }
+        if (null !== $this->outputSchema) {
+            $data['outputSchema'] = $this->outputSchema;
         }
 
         return $data;
--- a/src/Server/Handler/Request/CallToolHandler.php
+++ b/src/Server/Handler/Request/CallToolHandler.php
@@ -65,13 +65,16 @@ final class CallToolHandler implements RequestHandlerInterface
 
             $result = $this->referenceHandler->handle($reference, $arguments);
 
+            $structuredContent = null;
             if (!$result instanceof CallToolResult) {
-                $result = new CallToolResult($reference->formatResult($result));
+                $structuredContent = $reference->extractStructuredContent($result);
+                $result = new CallToolResult($reference->formatResult($result), structuredContent: $structuredContent);
             }
 
             $this->logger->debug('Tool executed successfully', [
                 'name' => $toolName,
                 'result_type' => \gettype($result),
+                'structured_content' => $structuredContent,
             ]);
 
             return new Response($request->getId(), $result);
--- a/tests/Inspector/Http/snapshots/HttpSchemaShowcaseTest-tools_call-calculate_range.json
+++ b/tests/Inspector/Http/snapshots/HttpSchemaShowcaseTest-tools_call-calculate_range.json
@@ -5,5 +5,11 @@
             "text": "{\n    \"result\": 50,\n    \"operation\": \"10 multiply 5\",\n    \"precision\": 2,\n    \"within_bounds\": true\n}"
         }
     ],
+    "structuredContent": {
+        "result": 50,
+        "operation": "10 multiply 5",
+        "precision": 2,
+        "within_bounds": true
+    },
     "isError": false
 }
--- a/tests/Inspector/Http/snapshots/HttpSchemaShowcaseTest-tools_call-format_text.json
+++ b/tests/Inspector/Http/snapshots/HttpSchemaShowcaseTest-tools_call-format_text.json
@@ -5,5 +5,11 @@
             "text": "{\n    \"original\": \"Hello World Test\",\n    \"formatted\": \"HELLO WORLD TEST\",\n    \"length\": 16,\n    \"format_applied\": \"uppercase\"\n}"
         }
     ],
+    "structuredContent": {
+        "original": "Hello World Test",
+        "formatted": "HELLO WORLD TEST",
+        "length": 16,
+        "format_applied": "uppercase"
+    },
     "isError": false
 }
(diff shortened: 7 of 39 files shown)


bigdevlarry:
Hi @chr-hertel, & @CodeWithKyrian  the OutputSchema support PR is back up

bigdevlarry:
Hi @chr-hertel, @nyholm can I get some eyes on here for the output schema support. Ty

Nyholm:
Interesting. I do like the idea. I'll spend some more time to read the code.

tests/Inspector/Http/snapshots/HttpCombinedRegistrationTest-tools_call-discovered_status_check.json
```diff
@@ -5,5 +5,8 @@
       "text": "System status: OK (discovered)"
     }
   ],
-  "isError": false
+  "isError": false,
+  "structuredContent": {
+    "result": "System status: OK (discovered)"
+  }
```
chr-hertel: I wonder a bit about this change here - it's not adding value to me. spec is unclear about it, but i think we shouldn't just add it always.

chr-hertel (changes requested):
In general this looks pretty good already, just one thing i noticed with one test scenario. `structuredContent` should be optional in response payload

bigdevlarry:
@chr-hertel To make `structuredContent` optional in the response payload, I'll have to change the behavior so that outputSchema is only present when explicitly provided in the #[McpTool] attribute, not auto-generated from return types or DocBlocks. When an outputSchema is explicitly provided, `structuredContent` must be included in the response (as required by the MCP spec). When no outputSchema is provided, neither outputSchema nor structuredContent is included.

tests/Inspector/Http/snapshots/HttpCombinedRegistrationTest-tools_call-discovered_status_check.json
bigdevlarry: @chr-hertel this has now been actioned, so structrued content will only be added if outputschema is present and explicility specified

chr-hertel:
The case that i was pointing out was mostly just returning a string, but still had that `result => '...'` structure on top - i just can't find any reference for that in the spec or do i miss sth?

bigdevlarry:
> The case that i was pointing out was mostly just returning a string, but still had that `result => '...'` structure on top - i just can't find any reference for that in the spec or do i miss sth?

Good shout ! I was working with the wrong assumption that the result wrapper was in the spec. `structuredContent` now matches the outputSchema properties directly, no more top level result wrapping. The structure is as in the MCP spec example. I've updated the diagram on the PR to reflect the test as well

src/Capability/Registry/ToolReference.php
```diff
@@ -111,4 +111,48 @@ public function formatResult(mixed $toolExecutionResult): array
...
+        }
+
+        if (\is_array($toolExecutionResult)) {
+            if (array_is_list($toolExecutionResult) && isset($outputSchema['additionalProperties'])) {
+                // Wrap list in "object" schema for additionalProperties
+                return ['items' => $toolExecutionResult];
+            }
+
+            return $toolExecutionResult;
+        }
```
soyuka, line 128: shouldn't we leave that to implementations of the handler? I'm not sure we should alter the content at any point in the sdk.

src/Capability/Registry/ToolReference.php
```diff
@@ -111,4 +111,48 @@ public function formatResult(mixed $toolExecutionResult): array
...
+     * @throws \JsonException if JSON encoding fails for non-Content array/object results
+     */
+    private function normalizeValue(mixed $value): mixed
+    {
+        if (\is_object($value) && !($value instanceof Content)) {
+            return json_decode(json_encode($value, \JSON_THROW_ON_ERROR), true, 512, \JSON_THROW_ON_ERROR);
+        }
+
+        return $value;
+    }
```
soyuka: Not sure to understand this already happens in `formatResult` if our goal is to return a structured content maybe that we should add a condition above?

src/Schema/Result/CallToolResult.php
```diff
@@ -59,12 +59,13 @@ public function __construct(
...
-     * @param array<string, mixed>|null $meta    Optional metadata
+     * @param Content[]                 $content           The content of the tool result
+     * @param array<string, mixed>|null $meta              Optional metadata
+     * @param array<string, mixed>|null $structuredContent Optional structured content matching the tool's outputSchema
      */
-    public static function success(array $content, ?array $meta = null): self
+    public static function success(array $content, ?array $meta = null, ?array $structuredContent = null): self
     {
-        return new self($content, false, null, $meta);
+        return new self($content, false, $meta, $structuredContent);
```
soyuka: not sure to why this is needed content can be a structured content since https://github.com/modelcontextprotocol/php-sdk/pull/93 or am I missing something?

src/Schema/Tool.php
```diff
@@ -24,28 +24,37 @@
  *     properties: array<string, mixed>,
  *     required: string[]|null
  * }
+ * @phpstan-type ToolOutputSchema array{
+ *     type: 'object',
```
soyuka, line 28:
```suggestion
 *     type: 'object',
```
Can't the schema be a list? `object` type is not specified in the spec for the output.

src/Schema/Tool.php
```diff
@@ -78,13 +88,23 @@ public static function fromArray(array $data): self
...
         }
 
+        if (isset($data['outputSchema']) && \is_array($data['outputSchema'])) {
+            if (!isset($data['outputSchema']['type']) || 'object' !== $data['outputSchema']['type']) {
+                throw new InvalidArgumentException('Tool outputSchema must be of type "object".');
+            }
+            if (isset($data['outputSchema']['properties']) && \is_array($data['outputSchema']['properties']) && empty($data['outputSchema']['properties'])) {
+                $data['outputSchema']['properties'] = new \stdClass();
+            }
+        }
```
soyuka, line 98: I advise against validating this here we should leave that to the user discretion, it'll make the sdk more portable and subject to specification changes.

src/Server/Handler/Request/CallToolHandler.php
```diff
@@ -62,15 +62,22 @@ public function handle(Request $request, SessionInterface $session): Response|Er
...
-            if (!$result instanceof CallToolResult) {
-                $result = new CallToolResult($reference->formatResult($result));
+            $structuredContent = null;
+            if (null !== $reference->tool->outputSchema && !$rawResult instanceof CallToolResult) {
+                $structuredContent = $reference->extractStructuredContent($rawResult);
+            }
+
+            $result = $rawResult;
+            if (!$rawResult instanceof CallToolResult) {
+                $result = new CallToolResult($reference->formatResult($rawResult), structuredContent: $structuredContent);
```
soyuka: I'd try not to change this part of the code, I understand we want to comply with the spec that says:

```
Servers MUST provide structured results that conform to this schema.
```

If possible I think it should be handled in the `formatResult`.

My personal opinion would be to throw if the handler doesn't return a structured content.

tests/Inspector/Http/snapshots/HttpSchemaShowcaseTest-tools_call-calculate_range.json
```diff
@@ -1,9 +1,9 @@
...
-        }
-    ],
-    "isError": false
+  "content": [
+    {
+      "type": "text",
+      "text": "{\n    \"result\": 50,\n    \"operation\": \"10 multiply 5\",\n    \"precision\": 2,\n    \"within_bounds\": true\n}"
+    }
+  ],
+  "isError": false
```
soyuka: this should probably not change

tests/Inspector/Http/snapshots/HttpSchemaShowcaseTest-tools_call-format_text.json
```diff
@@ -1,9 +1,9 @@
...
-    ],
-    "isError": false
+  "content": [
+    {
+      "type": "text",
+      "text": "{\n    \"original\": \"Hello World Test\",\n    \"formatted\": \"HELLO WORLD TEST\",\n    \"length\": 16,\n    \"format_applied\": \"uppercase\"\n}"
+    }
+  ],
+  "isError": false
 }
```
soyuka, line 15: this should probably not change

tests/Inspector/Http/snapshots/HttpSchemaShowcaseTest-tools_call-generate_config.json
```diff
@@ -1,9 +1,9 @@
...
-        }
-    ],
-    "isError": false
+  "content": [
+    {
+      "type": "text",
+      "text": "{\n    \"success\": true,\n    \"config\": {\n        \"app\": {\n            \"name\": \"TestApp\",\n            \"env\": \"development\",\n            \"debug\": true,\n            \"url\": \"https://example.com\",\n            \"port\": 8080\n        },\n        \"generated_at\": \"2025-01-01T00:00:00+00:00\",\n        \"version\": \"1.0.0\",\n        \"features\": {\n            \"logging\": true,\n            \"caching\": false,\n            \"analytics\": false,\n            \"rate_limiting\": false\n        }\n    },\n    \"validation\": {\n        \"app_name_valid\": true,\n        \"url_valid\": true,\n        \"port_in_range\": true\n    }\n}"
+    }
+  ],
+  "isError": false
```
soyuka: this should probably not change (same for the other snapshots not sure why they change)

tests/Unit/Capability/Discovery/SchemaGeneratorFixture.php
```diff
@@ -337,6 +338,12 @@ public function mixedTypes(
     /**
      * Complex nested Schema constraints.
      */
+    #[McpTool(
+        outputSchema: [
+            'type' => 'object',
+            'additionalProperties' => true,
+        ]
+    )]
```
soyuka, line 346: on a general matter I think (if possible) that we should avoid to change existing fixtures but instead add new ones, as for that we need to change existing tests.

soyuka:
Nice addition we definitely miss this! Thanks!

tests/Inspector/Http/snapshots/HttpSchemaShowcaseTest-tools_call-calculate_range.json
bigdevlarry: Just indentation, no code change

tests/Inspector/Http/snapshots/HttpSchemaShowcaseTest-tools_call-format_text.json
bigdevlarry: Just indentation, no code change

tests/Inspector/Http/snapshots/HttpSchemaShowcaseTest-tools_call-generate_config.json
bigdevlarry: Just indentation, no code change

src/Server/Handler/Request/CallToolHandler.php
bigdevlarry: Yes, we're trying to comply with what the spec says here. I also think it reads better having it here otherwise, @chr-hertel feels otherwise

tests/Inspector/Http/snapshots/HttpSchemaShowcaseTest-tools_call-calculate_range.json
soyuka: I know but why as we didn't change anything to the json serialization this is weird to me and pollutes the diff

src/Schema/Tool.php
bigdevlarry: I was following the validation pattern we've for the inputSchema setup above. Also, if we don't add validation for outputSchema, how can we have a structured output format? Users might use a structure we don't support, don't you think?

src/Schema/Result/CallToolResult.php
bigdevlarry: Missed that PR. We can use the existing structure. Thanks for flagging

chr-hertel:
@bigdevlarry we have a misunderstanding here: my point was that output schema - in my understanding - is only needed for more complex structures - not for simple string responses for example.

compared with the typescript-sdk:

**simple tool response**
image

**structured tool response**
image

---

my take-away here is: there's no need for the `structuredContent` keyword in the response, if we only return a string - but feel free to point out what i miss

bigdevlarry:
@chr-hertel At the moment, if `structuredContent` returns a primitive (string, int, bool, etc.), extractStructuredContent() method in the `src/Capability/Registry/ToolReference.php` returns null, and `structuredContent` is not included in the response. So image 2, looks valid where you've get_weather because the outputSchema returns an object/array. If this is still not the intended behaviour, maybe you can point out what to add/remove in the extractStructuredContent method 
Screenshot 2025-12-08 at 21 31 59

src/Schema/Tool.php
soyuka: Yeah I'm actually not a huge fan of the above `inputSchema` implementation either :). 

I think it's the responsability of another abstraction to do this, and its also part of a bigger discussion around json schema and validation (I suggested already that these should not be coded into the php-sdk as they're quite complicated subjects).

phpstan-baseline.neon
```diff
@@ -5,3 +5,9 @@ parameters:
 			identifier: return.type
 			count: 1
 			path: src/Schema/Result/ReadResourceResult.php
+
+		-
+			message: '#^Method Mcp\\Tests\\Unit\\Capability\\Discovery\\DocBlockTestFixture\:\:methodWithMultipleTags\(\) has RuntimeException in PHPDoc @throws tag but it''s not thrown\.$#'
+			identifier: throws.unusedType
+			count: 1
+			path: tests/Unit/Capability/Discovery/DocBlockTestFixture.php
```
chr-hertel: this needs to be removed - see pipeline failure

src/Schema/Tool.php
chr-hertel: that's the only information i could find right now, states "object" but i see your point, yup

https://github.com/modelcontextprotocol/modelcontextprotocol/blob/668a94395b9cbceb78e347f37f345fb6347a2685/schema/2025-11-25/schema.ts#L1108-L1111

tests/Inspector/Http/snapshots/HttpSchemaShowcaseTest-tools_call-calculate_range.json
chr-hertel: would agree, let's try to keep the diff minimal in general

chr-hertel:
@bigdevlarry  @soyuka what do you think about this case? (open question - not having a solid opinion yet)
image

On the one hand side I like the idea of having simple return values/structures and the SDK taking care of wrapping it in spec-specific structures, but with that example of mine there is somehow a disconnect between complexity of definitiv the output schema (aware of the entire schema) and the return value (only a subset but straightforward)

i feel like it's the purpose of the SDK to take away and ease as much as possible - with the opportunity to opt into being super explicit in the handler

bigdevlarry:
@chr-hertel Thanks for raising the list concern, it made me revisit the extraction logic carefully. I agree that we need to handle lists correctly, but after reviewing, I think the main issue comes from the `array_is_list` + `additionalProperties` wrapping. That logic is interpreting schema semantics inside `extractStructuredContent()`, which can produce outputs the schema never declared and introduces surprising behavior.

To address this:
- We should remove the `array_is_list + additionalProperties` logic entirely.
`extractStructuredContent()` should simply normalize objects → arrays and pass arrays through.

Action Point 
I’ll follow up with a code refinement that:
- Simplifies extractStructuredContent() to just normalize objects → arrays without wrapping or guessing.
- Adds concrete examples showing list outputs are handled correctly and invalid outputs trigger validation errors.

bigdevlarry:
Hiya @chr-hertel, All feedback resolved now.

src/Capability/Registry/ToolReference.php
```diff
@@ -111,4 +111,36 @@ public function formatResult(mixed $toolExecutionResult): array
...
+     * @return array<string, mixed>|null the structured content, or null if not extractable
+     *
+     * @throws \JsonException if JSON encoding fails for non-Content array/object results
+     */
+    public function extractStructuredContent(mixed $toolExecutionResult): ?array
+    {
+        $outputSchema = $this->tool->outputSchema;
+        if (null === $outputSchema) {
+            return null;
+        }
```
Nyholm: We can/should return StructuredContent even if output schema is missing

I suggest remove this.

src/Capability/Registry/ToolReference.php
```diff
@@ -111,4 +111,36 @@ public function formatResult(mixed $toolExecutionResult): array
...
+
+        if (\is_object($toolExecutionResult) && !($toolExecutionResult instanceof Content)) {
+            return json_decode(
+                json_encode($toolExecutionResult, \JSON_THROW_ON_ERROR),
+                true,
+                512,
+                \JSON_THROW_ON_ERROR
+            );
+        }
```
Nyholm, line 139: Is this a way to covert an object to an array?

Please add some more flags to `json_encode()`. See `ToolReference::formatResult()`.

src/Server/Handler/Request/CallToolHandler.php
```diff
@@ -61,17 +61,23 @@ public function handle(Request $request, SessionInterface $session): Response|Er
...
```
Nyholm: Let's write this simpler: 


```php
$result = $this->referenceHandler->handle($reference, $arguments);

$structuredContent = null;
if (!$result instanceof CallToolResult) {
    $structuredContent = $reference->extractStructuredContent($result);
    $result = new CallToolResult($reference->formatResult($result), structuredContent: $structuredContent);
}
```

src/Capability/Registry/ToolReference.php
bigdevlarry: Yes, it's used to convert an object to array

soyuka:
ArrayLoader also calls the SchemaGenerator could you check if it needs some changes?

bigdevlarry:
@soyuka Thanks, it definitely needs. We should also define the output schema as it is below..
                
``` 
 $tool = new Tool(
                    name: $name,
                    inputSchema: $inputSchema,
                    description: $description,
                    annotations: $data['annotations'] ?? null,
                    icons: $data['icons'] ?? null,
                    meta: $data['meta'] ?? null,
                    outputSchema: $data['outputSchema'], // only used when explicitly defined
                );
```

bigdevlarry:
All good now ..

Cruiser13:
This looks great, thanks a bunch!

chr-hertel (dismissed):
Let's get this merged - thanks for your effort @bigdevlarry! :+1:

chr-hertel (approved):
Ready for merge from my POV after for 0.3.0

src/Capability/Discovery/SchemaGenerator.php
```diff
@@ -81,6 +82,28 @@ public function generate(\ReflectionMethod|\ReflectionFunction $reflection): arr
...
+        $mcpToolAttrs = $reflection->getAttributes(McpTool::class, \ReflectionAttribute::IS_INSTANCEOF);
+        if (!empty($mcpToolAttrs)) {
+            $mcpToolInstance = $mcpToolAttrs[0]->newInstance();
+            if (null !== $mcpToolInstance->outputSchema) {
+                return $mcpToolInstance->outputSchema;
+            }
+        }
+
+        return null;
+    }
```
soyuka, line 104:
```suggestion
    $mcpToolAttrs = $reflection->getAttributes(McpTool::class, \ReflectionAttribute::IS_INSTANCEOF);

    if ($mcpToolAttrs) {
        $mcpToolInstance = $mcpToolAttrs[0]->newInstance();
        return $mcpToolInstance->outputSchema;
    }
    
    return null;
```

Improves lisibility, the second check wasn't useful as the default value is null.

src/Schema/Tool.php
soyuka, line 98: suggestion:

```php
   /**
     * @param ToolData $data
     */
    public static function fromArray(array $data): self
    {
        if (empty($data['name']) || !\is_string($data['name'])) {
            throw new InvalidArgumentException('Invalid or missing "name" in Tool data.');
        }

        if (!isset($data['inputSchema']) || !\is_array($data['inputSchema'])) {
            throw new InvalidArgumentException('Invalid or missing "inputSchema" in Tool data.');
        }

        if (!isset($data['inputSchema']['type']) || 'object' !== $data['inputSchema']['type']) {
            throw new InvalidArgumentException('Tool inputSchema must be of type "object".');
        }

        $inputSchema = self::normalizeSchemaProperties($data['inputSchema']);
        $outputSchema = null;

        if (isset($data['outputSchema']) && \is_array($data['outputSchema'])) {
            if (!isset($data['outputSchema']['type']) || 'object' !== $data['outputSchema']['type']) {
                throw new InvalidArgumentException('Tool outputSchema must be of type "object".');
            }

            $outputSchema = self::normalizeSchemaProperties($data['outputSchema']);
        }

        return new self(
            $data['name'],
            $inputSchema,
            isset($data['description']) && \is_string($data['description']) ? $data['description'] : null,
            isset($data['annotations']) && \is_array($data['annotations']) ? ToolAnnotations::fromArray($data['annotations']) : null,
            isset($data['icons']) && \is_array($data['icons']) ? array_map(Icon::fromArray(...), $data['icons']) : null,
            isset($data['_meta']) && \is_array($data['_meta']) ? $data['_meta'] : null,
            $outputSchema
        );
    }

    /**
     * Normalize schema properties: convert empty properties array to stdClass
     *
     * @param array<string, mixed> $schema
     *
     * @return array{properties: object}&array<string, mixed> $schema
     */
    private static function normalizeSchemaProperties(array $schema): array
    {
        if (isset($schema['properties']) && \is_array($schema['properties']) && empty($schema['properties'])) {
            $schema['properties'] = new \stdClass();
        }

        return $schema;
    }
```

soyuka (changes requested):

src/Schema/Tool.php
bigdevlarry: @soyuka If we change the $data['inputSchema'] -> $inputSchema

 Parameter #2 $inputSchema of class Mcp\Schema\Tool constructor expects array{type: 'object', properties: array<string, mixed>, required: array<string>|null}, array{prop  
         erties: object} given.                                                                                                                                                    
         🪪  argument.type                                                                                                                                                         
         💡  Array does not have offset 'type'.
         
         
 Similar to the outputSchema as well .. The error is visible when you run 
` vendor/bin/phpstan analyse --memory-limit=528M`

soyuka: We could modify the array by reference 

    private static function normalizeSchemaProperties(array &$schema): void
   
Not adding any variables?

chr-hertel: @soyuka fine by you as a follow up?

bigdevlarry: @chr-hertel this can go, myself or Soyuka will make this as follow up

chr-hertel (approved):
Thanks again @bigdevlarry!

PR #158, opened 2025-11-21
Fix incorrect disconnect status code

luoyue712:
HTTP status code should return 200
image

### DIFF
--- a/src/Server/Transport/StreamableHttpTransport.php
+++ b/src/Server/Transport/StreamableHttpTransport.php
@@ -115,7 +115,7 @@ class StreamableHttpTransport extends BaseTransport implements TransportInterfac
 
         $this->handleSessionEnd($this->sessionId);
 
-        return $this->withCorsHeaders($this->responseFactory->createResponse(204));
+        return $this->withCorsHeaders($this->responseFactory->createResponse(200));
     }
 
     protected function createJsonResponse(): ResponseInterface


chr-hertel:
@CodeWithKyrian do you recall why it's 204 - can't find sth in the specs directly?

CodeWithKyrian:
Yeah, there’s nothing in the MCP spec that requires a specific code here. The only other 2xx mentioned is for notifications (202). We went with 204 because it’s the standard HTTP response for a successful DELETE with no body. I haven’t seen any MCP clients reject it, so it seemed like the most logical choice at the time.

luoyue712:
> Yeah, there’s nothing in the MCP spec that requires a specific code here. The only other 2xx mentioned is for notifications (202). We went with 204 because it’s the standard HTTP response for a successful DELETE with no body. I haven’t seen any MCP clients reject it, so it seemed like the most logical choice at the time.

I tested with MCP Inspector v0.17.2 and it cannot shut down properly. The error is shown in the previously mentioned image. I compared it with Typescript SDK, and they indeed return 200.

Nyholm:
Give me a few minutes to verify

Nyholm:
Im reading the specification [here](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#sending-messages-to-the-server). I fail to reproduce the errors your screenshot. 

(I think you are correct, but I fail to verify) Can you please share how you start the inspector and how to reproduce?

luoyue712:
> Im reading the specification [here](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#sending-messages-to-the-server). I fail to reproduce the errors your screenshot.
> 
> (I think you are correct, but I fail to verify) Can you please share how you start the inspector and how to reproduce?

Click the Disconnect button.
image

chr-hertel:
Could reproduce locally with inspector:
image

and also verify that it works with this branch - thanks @lvluoyue for the patch!

Nyholm:
Sorry, my bad. You need to use "proxy", and not "direct" setting. 

I see both 200 and 202 works and the specification does not cover this. Im happy with this change.

Issue #159, opened 2025-11-21
[Server] Expose request-level metadata (e.g., securitySchema) to tool handlers


AgentSlim:
**Is your feature request related to a problem? Please describe.**
Tool handlers currently receive only the parameters defined for the tool call.
Metadata included in the request envelope (for example a securitySchema) is not available inside the handler.
This makes it impossible to perform logic that depends on contextual information from the request itself.

**Describe the solution you’d like**
A way for tool handlers to access the metadata from the incoming request envelope.
This could be done by injecting either the full request object or just the meta section into the handler method.
Having this information available would allow tools to perform authorization checks, tenant selection, and similar context-dependent actions.

**Describe alternatives you’ve considered**
The only workaround at the moment is extracting envelope metadata before the request reaches the tool dispatching layer and storing it somewhere manually.
Once inside the handler, that information can no longer be accessed in a clean or reliable way.

**Additional context**
A common use case is reading a securitySchema sent by the client.
This information is part of the envelope, not part of the tool parameters, and is therefore currently inaccessible inside tool logic.

**What tool handlers currently look like**
```php
use MCP\Server\Attributes\McpTool;

final class ExampleTools
{
    #[McpTool(name: 'example_action')]
    public function exampleAction(string $input): array
    {
        // Only defined parameters are available.
        // Envelope metadata (e.g., securitySchema) cannot be accessed here.

        return ['result' => 'ok'];
    }
}
```

**Proposed Option A — Inject full request object**
```php
use MCP\Server\Attributes\McpTool;
use MCP\Types\CallToolRequest;

final class ExampleTools
{
    #[McpTool(name: 'example_action')]
    public function exampleAction(
        string $input,
        CallToolRequest $request,
    ): array {
        $meta = $request->meta ?? null;
        $schema = $meta['securitySchema'] ?? null;

        return [
            'result' => 'ok',
            'securitySchema' => $schema,
        ];
    }
}
```

**Proposed Option B — Inject only the meta section**
```php
use MCP\Server\Attributes\McpTool;

final class ExampleTools
{
    #[McpTool(name: 'example_action')]
    public function exampleAction(
        string $input,
        array $meta = [],
    ): array {
        $schema = $meta['securitySchema'] ?? null;

        return [
            'result' => 'ok',
            'securitySchema' => $schema,
        ];
    }
}
```

chr-hertel:
Yup, sounds valid to me, thankts @AgentSlim for raising this!

[...]

PR #161, opened 2025-11-21
Add request-level Metadata injection for tool handlers

AgentSlim:
This Pull Request implements the feature discussed in [Issue #159](https://github.com/modelcontextprotocol/php-sdk/issues/159):
Expose request-level metadata to tool handlers.
•	Introduced a new Mcp\Schema\Metadata value object to represent request-scoped metadata from the MCP message envelope.
•	Updated the handler invocation logic so that parameters type-hinted as Metadata automatically receive the _meta portion of the incoming request.
•	Enhanced CallToolHandler to forward the envelope metadata into the parameter resolver.
•	Added comprehensive unit tests covering:
•	metadata injection
•	fallback behavior when no metadata is provided
•	error handling during parameter resolution
•	backwards-compatibility with existing tool handlers
•	Updated documentation with usage examples showing how tool authors can access metadata such as securitySchema.

## Motivation and Context
This change addresses [Issue #159](https://github.com/modelcontextprotocol/php-sdk/issues/159), which highlights that tool handlers currently cannot access metadata from the request envelope.
Many real-world use cases depend on this metadata — for example:
•	authorization logic based on a securitySchema
•	tenant or environment selection
•	contextual behavior determined by the client request

Previously, developers had to rely on indirect workarounds (middleware, global state, decorators), because metadata was lost before reaching the handler.
This feature introduces a clean, explicit, opt-in approach based on type-hinting, keeping handler signatures transparent and predictable.

## How Has This Been Tested?
•	Added new unit tests verifying metadata extraction and injection.
•	Validated that existing handlers without a Metadata parameter continue working unchanged.
•	Tested optional and missing metadata cases to ensure robust behavior.
•	Verified handler execution flow with a real tool using a securitySchema passed via _meta.

## Breaking Changes
This change is not breaking.
Existing handlers and configurations continue to work as before.
Metadata is injected only when a parameter explicitly type-hints Mcp\Schema\Metadata.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context
This approach is fully backward-compatible and future-proof.
The type-hint-based injection model can be extended later to support additional request-level context types (e.g., full request object, execution context, or transport metadata).

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,63 @@ $server = Server::builder()
     ->build();
 ```
 
+### Request Metadata Injection
+
+You can access request-scoped metadata inside tool handlers by type-hinting the `Mcp\Schema\Metadata` value object. The SDK will inject it automatically when present on the request.
+
+How it works:
+- Clients may send arbitrary metadata in `params._meta` of the JSON-RPC request.
+- The server forwards this metadata for tool calls and the `ReferenceHandler` injects it into parameters typed as `Metadata` or `?Metadata`.
+- If your handler declares a non-nullable `Metadata` parameter but the request contains no `params._meta`, the SDK returns an internal error. Use `?Metadata` if it is optional.
+
+Example handler usage:
+
+```php
+<?php
+
+use Mcp\Capability\Attribute\McpTool;
+use Mcp\Schema\Metadata;
+
+final class ExampleTools
+{
+    #[McpTool(name: 'example_action')]
+    public function exampleAction(string $input, Metadata $meta): array
+    {
+        $schema = $meta->get('securitySchema');
+
+        return [
+            'result' => 'ok',
+            'securitySchema' => $schema,
+        ];
+    }
+
+    #[McpTool(name: 'example_optional')]
+    public function exampleOptional(string $input, ?Metadata $meta = null): string
+    {
+        return $meta?->get('traceId') ?? 'no-meta';
+    }
+}
+```
+
+Calling a tool with metadata (JSON-RPC example):
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": "1",
+  "method": "tools/call",
+  "params": {
+    "name": "example_action",
+    "arguments": { "input": "hello" },
+    "_meta": { "securitySchema": "secure-123", "traceId": "abc-xyz" }
+  }
+}
+```
+
+Notes:
+- For non-nullable `Metadata` parameters, the client must provide `params._meta`; otherwise an internal error is returned.
+- For nullable `?Metadata` parameters, `null` will be injected when `params._meta` is absent.
+
 ## Documentation
 
 **Core Concepts:**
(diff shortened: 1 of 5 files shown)


lvluoyue (changes requested):
https://github.com/modelcontextprotocol/php-sdk/pull/161/files#diff-c16cd16cec76740a8e2c4b7efecba1491529cbfd55fb9b6348c7d1b24d01c903L43-L47
Here also needs the following code:
```php

                // Inject request metadata if requested
                if (Metadata::class === $typeName) {
                    if (isset($arguments['_meta']) && \is_array($arguments['_meta'])) {
                        $finalArgs[$paramPosition] = new Metadata($arguments['_meta']);
                    } elseif ($parameter->allowsNull()) {
                        $finalArgs[$paramPosition] = null;
                    } else {
                        $reflectionName = $reflection instanceof \ReflectionMethod
                            ? $reflection->class.'::'.$reflection->name
                            : 'Closure';
                        throw RegistryException::internalError("Missing required request metadata for parameter `{$paramName}` in {$reflectionName}. Provide `_meta` in request params or make the parameter nullable.");
                    }
                    continue;
                }
```

Nxooah:
Hi, very well done. I need exactly the same thing. Hopefully it will be merged soon.

Nyholm:
Thank you. I appreciate the idea.

Correct me if Im wrong, but this is not specified in the mcp protocol, is it?

chr-hertel:
The spec only states this about metadata:
https://modelcontextprotocol.io/specification/2025-11-25/basic/index#meta

but it's not helpful if you can't access it when used - this is what this PR is about

CodeWithKyrian:
Exactly!

On a related note, maybe this is a good time to revisit the Context object idea. Beyond just `Meta` and `ClientGateway`, there are probably going to be more things handlers need access to over time, and having a single injectable `Context` that exposes all request-scoped services feels like the cleanest long-term approach.

Once auth lands, this becomes even more relevant, since handlers will likely need access to user/session state without turning parameters into part of the Tool schema.

CC @chr-hertel @Nyholm @soyuka

soyuka:
In my opinion we should provide the tooling so that users can access part of the request they need, though indeed we shouldn't multiply context value objects that can be injected as its an open door to too many changes (though this would be probably detected through argument resolving). 

:+1: @CodeWithKyrian a context would be nice, I'm usually preferring array structures then objects when it comes to exposing such context but in our case it may be cleaner with an object. We must think about changes and BC-layers though as the protocol being quite immature it'll probably change a lot, so does this library. I suggest that we should make clear to users that every bits of this context can probably be undefined and that the API should be prepared for that. To me it'd make sense to expose the full `Request` and leave users pick what they need (with a higher Meta abstraction getter if we feel it improves the DX).

Nyholm:
I did implement a single injectable `RequestContext` in #176. I think that ticks all the boxes and maybe even replace this PR.

chr-hertel:
We merged #176, assuming that your concerns got addressed as well - feel free to rebase and reopen in case I missed something here.

PR #166, opened 2025-11-22
Add CakePHP integration plugin to README

dereuromark:
Add https://github.com/josbeir/cakephp-synapse to the "PHP Libraries Using the MCP SDK"
cc @josbeir


Nyholm:
Great. Thank you. 

I am usually very hesitant of adding links to "just created" libraries. However, this library is also "just created". 

I suggest (hopefully other maintainers agree) that we allow this and then at some point in the future (6 months from now, or when we tag 1.0) we review all our "community links" links and remove the ones that does not meet our (not yet created) criteria.

PR #171, opened 2025-11-30
Fix code style

Nyholm:
Im not sure why, but the main branch is red. We need this CS fix


chr-hertel (approved):

chr-hertel:
Thanks, good catch! :+1:

Nyholm:
Thank you for merging

PR #172, opened 2025-11-30
[Server] Extend changelog for covering 0.1

chr-hertel:
Basically kicking of the version structure - we need to collect for 0.2 in case of breaking changes or new features.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
 # Changelog
 
 All notable changes to `mcp/sdk` will be documented in this file.
+
+0.1
+---
+
+* First tagged release of package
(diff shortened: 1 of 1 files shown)


CodeWithKyrian:
Won't it be nice to have the release notes of 0.1 in that CHANGELOG for a start? And maybe even a GitHub worlflow to automatically update it with release notes so that part is automated

chr-hertel:
that would nice, yup - nothing i can act on now tho, since i want to take a nap :laughing:

chr-hertel:
would you want to see this tackled instead of this PR or a follow up?

CodeWithKyrian:
A follow up

PR #104, opened 2025-10-10
[Server] Implemented MCP logging specification with auto-injection

Adebayo120:
This PR implements the Model Context Protocol (MCP) logging specification, providing centralized logging capabilities with auto-injection support for capability handlers. This enables developers to debug across multiple MCP servers from a single client interface, addressing the challenge of distributed logging in MCP ecosystems.

## Motivation and Context

**Fixes #70**

This implementation addresses the missing logging functionality as outlined in the [MCP logging specification](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging). 

**Problem solved:**
- Developers working with 10-15 MCP servers struggle with debugging across distributed logs
- No centralized logging mechanism for MCP servers
- Missing client-controlled log level filtering
- No easy way to access logging within tools/resources/prompts

**Key benefits:**
- **Centralized logging**: All server logs flow to the client for unified debugging
- **Client-controlled filtering**: Clients can set log levels to reduce noise
- **Auto-injection**: Zero-configuration logger access in capability handlers
- **Fallback support**: Compatible with existing external PSR-3 loggers

## How Has This Been Tested?

**Comprehensive test coverage:**
- ✅ **360 unit tests** with **1530 assertions** - all passing
- ✅ **0 PHPStan errors** - full static analysis compliance
- ✅ **0 PHP CS Fixer issues** - code style compliance

**Test scenarios:**
- Auto-injection of `McpLogger` into tools, resources, and prompts
- Client log level filtering (debug → emergency levels)
- Fallback to external PSR-3 loggers when MCP transport unavailable
- Notification serialization and transport integration
- Error handling and graceful degradation
- Builder configuration and server lifecycle integration

**Working examples:**
- `examples/stdio-logging-showcase/` - Complete demonstration of auto-injection
- Updated calculator example with comprehensive logging

## Breaking Changes

**None.** This implementation maintains full backward compatibility:
- Existing servers continue to work without modification
- Logging is opt-in via `->enableMcpLogging()` 
- No changes to existing method signatures
- Auto-injection only occurs when `McpLogger` or `LoggerInterface` parameters are declared

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
[...]

PR #176, opened 2025-12-01
[Server] Add RequestContext and logging

Nyholm:
This pull request continues the work in https://github.com/modelcontextprotocol/php-sdk/pull/104 and solves the problem in #161. Ping @Adebayo120 and @AgentSlim

We introduce a `RequestContext`

## Motivation and Context
We want to have request specific services that only live in the context of a `RequestHandlerInterface`. Ie services that has access to the `Session`. These cannot live inside the PSR-11 Container. 

PR #104 added a `ClientLogger` which would live next to our existing `ClientGateway`. To avoid adding more and more features in `ReferenceHandler`, I created a `RequestContext` which gives you access to session and request data (like metadata). The `RequestContext` will also build you a `ClientGateway`.

## How Has This Been Tested?
Tested with unit test and Inspector

## Breaking Changes
Yes. I have just deprecated `ClientAwareInterface`. I would like to remove it. Should I do it in this PR?

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

### DIFF
--- a/docs/mcp-elements.md
+++ b/docs/mcp-elements.md
@@ -516,7 +516,7 @@ The SDK automatically injects a `RequestContext` instance into handlers. This ca
 
 ```php
 use Mcp\Capability\Logger\ClientLogger;
-use Psr\Log\LoggerInterface;
+use Mcp\Server\RequestContext;
 
 #[McpTool]
 public function processData(string $input, RequestContext $context): array {


src/Server/Handler/Request/CallToolHandler.php
```diff
@@ -61,6 +61,7 @@ public function handle(Request $request, SessionInterface $session): Response|Er
             $reference = $this->registry->getTool($toolName);
 
             $arguments['_session'] = $session;
+            $arguments['_request'] = $request;
```
e0ipso, line 64: I see this exacerbates the need for #177 because my current workaround (for the Drupal MCP Server) involves a custom tool handler which will be outdated once this lands.

Nyholm: Hm.. not sure. I read #180, it seams like this PR will make things way easier for you to add DynamicTools

chr-hertel:
Yea, funny to see this - mostly in favor of @CodeWithKyrian opinion/style before - so I'm fine with going ahead.

The issues from design point of view that I have with this are:
a) a `RequestContext` object couples stuff even service instances together and injects them even if we don't need them - i don't explicitly require what I need but a larger container around it
b) I was in favor of having that marker interface instead of service injection via method signature because i feel like services should go into a construct or setter method - but that rather an OOP opinion.

and from the discussion that we had in Amsterdam i recall that you consider the capabilities more like controllers and therefore treating their method signature similar to what we do with controller signatures in Symfony would be consistent.

our compromise before was to just enable all three way:
* constructor based injection of services in case users prefer that and have their on container
* setter based optional dependencies with a marker interface
* method injection via argument type

they don't conflict with each other - and i don't see a huge overhead in having them

edit, just saw in #161 that @soyuka is also more in favor of having that `RequestContext`, right? so i'm even more in the minority here :D so let's go ahead

Nyholm:
Thank you for the review. 

> A) a RequestContext object couples stuff even service instances together and injects them even if we don't need them - i don't explicitly require what I need but a larger container around it

Not sure what you mean, it only (argument) injects it when it is required. 

> b) I was in favor of having that marker interface instead of service injection via method signature because i feel like services should go into a construct or setter method - but that rather an OOP opinion.
>  ...
>  our compromise before was to just enable all three way...

I think the marker interface is incorrect. Please correct me if I am wrong, but because we work with Fibers, then we need to manage state more carefully. 

Consider two tools like this: 

```php
class MyCoolTools implements ClientAwareInterface 
{
  // ...
  public function toolA() { /**/ }

  public function toolB() { /**/ }
}
```

When `toolA()` is called, we inject `ClientGateway` to `MyCoolTools` object. If we use the `ClientGateway`, we will suspend the Fiber, and keep reading from the input

If `toolB()` is call after `toolA()`, but before we suspend the fiber, we will inject a **NEW** (and different) `ClientGateway` object to the same `MyCoolTools` object. Whenever the first Fiber is resumed, it has a different `ClientGateway`. Ie, we have leaked memory/objects between Fibers/treads. 

This whole problem is avoided if we remove `ClientAwareInterface` and pass this request context as an argument instead. (Sure, the user can still mess up and store state incorrectly. But we dont encourage them anymore). 

This is why we cannot use setter injection :(

chr-hertel:
Aaah, didn't catch that you create new instances of the `ClientGateway` - and with a reason - got it :+1: 

The coupling comment was more about the fact that I would require `RequestContext` if I need the session - and I would also require `RequestContext` if i need the gateway. but that's a minor minor thing.

src/Server/ClientAwareInterface.php
```diff
@@ -11,6 +11,9 @@
 
 namespace Mcp\Server;
 
+/**
+ * @deprecated This is deprecated since 0.2.0 and will be removed in 0.3.0. Use RequestContext with argument injection instead.
+ */
 interface ClientAwareInterface
```
chr-hertel: yes, let's just remove it please :)

Nyholm:
Thank you for the review. The PR is updated

src/Server/RequestContext.php
```diff
@@ -0,0 +1,64 @@
...
+    public function getSession(): SessionInterface
+    {
+        return $this->session;
+    }
+
+    public function getClientGateway(): ClientGateway
+    {
+        if (null == $this->clientGateway) {
+            $this->clientGateway = new ClientGateway($this->session);
+        }
```
soyuka, line 51: could we have these in the constructor directly? It'll help with dependency injection in the mcp bundle

src/Server/RequestContext.php
```diff
@@ -0,0 +1,64 @@
...
+
+        return $this->clientGateway;
+    }
+
+    public function getClientLogger(): ClientLogger
+    {
+        if (null === $this->clientLogger) {
+            $this->clientLogger = new ClientLogger($this->getClientGateway(), $this->session);
```
soyuka, line 59: same for this any particular reason you don't put this in the constructor?

src/Server/RequestContext.php
Nyholm: No, we cannot :(

See https://github.com/modelcontextprotocol/php-sdk/pull/176#issuecomment-3620440018

src/Server/RequestContext.php
Nyholm: Yes, because that make the class stateful and that will cause issues because we are running different tools in different Fibers. 

https://github.com/modelcontextprotocol/php-sdk/pull/176#issuecomment-3620440018

src/Server/RequestContext.php
soyuka: Oh I see actually this is more a `runtime` context variable then a singleton.

soyuka (approved):

chr-hertel:
Thank you @Nyholm - merging this now - unblocking items of @e0ipso in #180

CodeWithKyrian:
Yeah @chr-hertel. This is exactly what I had in mind. Thanks @Nyholm for the fantastic work!

PR #180, opened 2025-12-05
[Server] feat(capability): add dynamic provider interfaces

### LINKED ISSUE #177: [Server] Dynamic provider interfaces to support runtime capability registration
**Is your feature request related to a problem? Please describe.**

The current MCP SDK architecture requires tools, prompts, and resources to be registered at build time with handlers that can be invoked via reflection-based parameter mapping. This poses significant challenges for framework integrations where:

1. **Capability definitions come from runtime sources** (configuration entities, database records, external services) rather than compile-time discovery
2. **Parameters are passed as associative arrays** from the MCP protocol, not as positional method arguments suitable for reflection
3. **Per-capability authorization requirements** cannot be evaluated until the capability is invoked with a specific context

Integrators are forced to implement workarounds:
- Custom request handlers that intercept calls and bypass the SDK's parameter matching
- Dummy/no-op handlers registered purely for metadata discovery during capability listing
- Custom loaders that construct schema objects manually to avoid reflection-based argument discovery

These workarounds are fragile, increase maintenance burden, and work against the SDK's intended extension points.

**Describe the solution you'd like**

Introduce three new interfaces that enable runtime capability registration and execution:

- **DynamicToolProviderInterface**: Provides tools at runtime via `getTools()`, checks support via `supportsTool()`, and executes via `executeTool()`
- **DynamicPromptProviderInterface**: Provides prompts at runtime via `getPrompts()`, checks support via `supportsPrompt()`, and retrieves via `getPrompt()`
- **DynamicResourceProviderInterface**: Provides resources at runtime via `getResources()`, checks support via `supportsResource()`, and reads via `readResource()`
- **DynamicResourceTemplateProviderInterface**: Provides resources at runtime via `getResourceTemplates()`, checks support via `supportsResourceRemplate()`, and reads via `readResourceTemplate()`

The Registry and request handlers should be extended to:
1. Accept registration of dynamic providers
2. Query dynamic providers during capability listing operations
3. Delegate execution to dynamic providers when they claim support for a capability

**Architectural Opportunities**

This pattern opens several valuable architectural doors:

1. **Clean Separation of Concerns**: Metadata discovery (`getTools()`), routing (`supportsTool()`), and execution (`executeTool()`) become explicit interface methods rather than implicit conventions

2. **Decorator-Based Cross-Cutting Concerns**: Authorization, logging, rate limiting, and caching can be implemented as provider decorators without modifying core execution logic

3. **Framework-Agnostic Integration Points**: Any framework (Drupal, Symfony, Laravel) can implement these interfaces to bridge their native plugin/service systems with MCP

4. **Lazy Loading**: Capabilities don't need to be fully instantiated at registration time; providers can defer expensive operations until actual invocation

5. **Multi-Tenancy Support**: Providers can filter capabilities based on session context, enabling per-tenant or per-user capability sets

6. **Hybrid Registration**: Static and dynamic providers can coexist, allowing simple tools to use traditional registration while complex integrations use dynamic providers

**Describe alternatives you've considered**

1. **Custom Request Handlers**: Intercept requests at the handler level and bypass SDK routing entirely. This works but duplicates error handling, logging, and result formatting logic.

2. **Build-Time Code Generation**: Generate static registrations from runtime sources during a compilation phase. This adds complexity and feels very brittle.

**Additional context**

The dynamic provider pattern aligns with how modern frameworks handle extensibility:
- Drupal uses plugin managers with runtime discovery
- Symfony uses ...
- Laravel uses ...

Providing first-class SDK support for this pattern may reduce friction for framework integrations and establish a clear architectural contract for capability providers.

e0ipso:
Closes #177

Introduce four new interfaces for runtime-defined MCP elements:
- `DynamicToolProviderInterface` for dynamic tool registration and execution
- `DynamicPromptProviderInterface` for dynamic prompt handling
- `DynamicResourceProviderInterface` for dynamic resource management
- `DynamicResourceTemplateProviderInterface` for URI template-based resources

These interfaces enable CMS plugins, database-driven tools, and proxy servers to expose MCP elements without reflection-based parameter mapping.

## Motivation and Context

Framework integrations (Drupal, Symfony, Laravel) need to register capabilities from runtime sources (config entities, database records, external services). The current reflection-based approach requires handlers with positional method arguments, forcing integrators into fragile workarounds.

Dynamic providers solve this by separating concerns:
- `getTools()` / `getPrompts()` / etc. for capability listing
- `supportsTool()` / `supportsPrompt()` / etc. for routing
- `executeTool()` / `getPrompt()` / etc. for execution

## How Has This Been Tested?

- Unit tests for all four provider interfaces
- Unit tests for result formatters
- Unit tests for completion handler with resource templates
- All existing tests pass

## Breaking Changes

~None. This is additive functionality. Existing registration patterns continue to work unchanged.~

RegistryInterface is modified, this is a BC break.

## Types of changes

## Checklist

## Additional context

New builder methods:
- `addDynamicToolProvider()`
- `addDynamicPromptProvider()`
- `addDynamicResourceProvider()`
- `addDynamicResourceTemplateProvider()`


src/Capability/Formatter/PromptResultFormatter.php
e0ipso, line 1: _Note for the reviewer:_ This code was moved from the `PromptReference`. This is now shared between `PromptReference` and `DynamicPromptReference`.

src/Capability/Formatter/ResourceResultFormatter.php
e0ipso, line 1: _Note for the reviewer:_ This code was moved from the `ResourceReference`. This is now shared between `ResourceReference` and `DynamicResourceReference`.

src/Capability/Formatter/ToolResultFormatter.php
e0ipso, line 1: _Note for the reviewer:_ This code was moved from the `ToolReference`. This is now shared between `ToolReference` and `DynamicToolReference`.

src/Capability/Provider/DynamicPromptProviderInterface.php
e0ipso, line 1: _Note for the reviewer:_ I don't foresee more classes implementing this interface other than `DynamicPromptProvider`. What is your preference in these cases, keep the interface or drop it?

src/Capability/Provider/DynamicResourceProviderInterface.php
e0ipso, line 1: _Note for the reviewer:_ I don't foresee more classes implementing this interface other than `DynamicResourceProvider`. What is your preference in these cases, keep the interface or drop it?

src/Capability/Provider/DynamicResourceTemplateProviderInterface.php
e0ipso, line 1: _Note for the reviewer:_ I don't foresee more classes implementing this interface other than `DynamicResourceTemplateProvider`. What is your preference in these cases, keep the interface or drop it?

src/Capability/Provider/DynamicToolProviderInterface.php
e0ipso, line 1: _Note for the reviewer:_ I don't foresee more classes implementing this interface other than `DynamicToolProvider`. What is your preference in these cases, keep the interface or drop it?

src/Capability/Registry/Loader/ArrayLoader.php
```diff
@@ -35,22 +34,20 @@
 
 /**
  * @author Antoine Bluchet <soyuka@gmail.com>
- *
- * @phpstan-import-type Handler from ElementReference
```
e0ipso, line 39: _Note for the reviewer:_ This was an unrelated fix to a PHPStan issue introduced in a direct push to `main`. Sorry for the noise. This was blocking the PR.

src/Capability/Registry/ArgumentPreparationInterface.php
e0ipso, line 1: _Note for the reviewer:_ I introduced this interface for better type hinting, but it can be skipped if you feel it is unnecessary.

src/Capability/Registry/DynamicArgumentPreparationTrait.php
e0ipso, line 1: _Note for the reviewer:_ All dynamic capabilities prepare arguments the same way. The interface + trait combo is my solution to code composition for DRY-ness. We could go with a base class strategy, but I [favored composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance).

src/Capability/Registry/DynamicPromptReference.php
e0ipso, line 1: _Note for the reviewer:_ This is the dynamic counterpart to the `PromptReference` class.

src/Capability/Registry/DynamicResourceReference.php
e0ipso, line 1: _Note for the reviewer:_ This is the dynamic counterpart to the `ResourceReference` class.

src/Capability/Registry/DynamicPromptReference.php
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    /**
+     * @param array<string, mixed> $arguments
+     */
+    public function __invoke(array $arguments): mixed
+    {
+        return $this->provider->getPrompt($this->promptName, $arguments);
+    }
```
e0ipso, line 53: _Note for the reviewer:_ I chose the `__invoke` pattern over a closure so I could add the `ClientAwareInterface` to gain access to client communications.

src/Capability/Registry/DynamicPromptReference.php
```diff
@@ -0,0 +1,62 @@
...
+    ) {
+        parent::__construct($this, false);
+    }
+
+    public function setClient(ClientGateway $clientGateway): void
+    {
+        if ($this->provider instanceof ClientAwareInterface) {
+            $this->provider->setClient($clientGateway);
+        }
+    }
```
e0ipso, line 45: _Note for the reviewer:_ There is some repetition with this in other places. Should we introduce a trait for this?

src/Capability/Registry/DynamicResourceTemplateReference.php
e0ipso, line 1: _Note for the reviewer:_ This is the dynamic counterpart to the `ResourceTemplateReference` class.

src/Capability/Registry/DynamicToolReference.php
e0ipso, line 1: _Note for the reviewer:_ This is the dynamic counterpart to the `ToolReference` class.

src/Capability/Registry/ElementReference.php
```diff
@@ -12,18 +12,39 @@
 namespace Mcp\Capability\Registry;
 
 /**
- * @phpstan-type Handler \Closure|array{0: object|string, 1: string}|string
```
e0ipso, line 15: _Note for the reviewer:_ This is still that unrelated PHPStan issue. It will show again later but I won't call it out again.

src/Capability/Registry/ElementReference.php
```diff
@@ -12,18 +12,39 @@
...
+class ElementReference implements ArgumentPreparationInterface
 {
     /**
-     * @param Handler $handler
+     * @param object|array{0: class-string|object, 1: string}|string $handler The handler can be a Closure, array method reference,
+     *                                                                        string function/class name, or a callable object (implementing __invoke)
      */
     public function __construct(
-        public readonly \Closure|array|string $handler,
+        public readonly object|array|string $handler,
```
e0ipso, line 26: _Note for the reviewer:_ `Closure` `is_a` object, so no BC-break (not that it's a problem at this point...).

I had to do this because now we have the dynamic handlers, that are objects with `__invoke`.

src/Capability/Registry/ElementReference.php
```diff
@@ -12,18 +12,39 @@
...
+     *
+     * Subclasses can override via traits or direct implementation to provide
+     * custom argument preparation (e.g., reflection-based mapping).
+     *
+     * @param array<string, mixed> $arguments       Raw arguments from MCP request
+     * @param callable             $resolvedHandler The resolved handler callable (unused in passthrough)
+     *
+     * @return array<int, mixed> The arguments as a list, ready to be spread
+     */
+    public function prepareArguments(
```
e0ipso, line 42: _Note for the reviewer:_ This is method is actually always overwritten via `DynamicArgumentPreparationTrait` or `ReflectionArgumentPreparationTrait`. However, it seemed to be a sensible default. Open to changes or to declare it `abstract`.

src/Capability/Registry/PromptReference.php
e0ipso, line 1: _Note for the reviewer:_ This was massively simplified only because we moved some code out of the class to a shared utility.

src/Capability/Registry/ReferenceHandler.php
```diff
@@ -29,270 +29,87 @@ public function __construct(
...
         }
 
-        if (\is_callable($reference->handler)) {
-            $reflection = $this->getReflectionForCallable($reference->handler, $session);
-            $arguments = $this->prepareArguments($reflection, $arguments);
-
-            return \call_user_func($reference->handler, ...$arguments);
+        // Set client on handler instances that implement ClientAwareInterface
+        if ($instance instanceof ClientAwareInterface) {
+            $instance->setClient($clientGateway);
```
e0ipso, line 52: _Note for the reviewer:_ This part is a bit murky, but solving for it more uniformly here will punt the problem to other places. I think this is the lesser evil, but I am happy to take ideas. The problem is that we don't really have an `$instance` for the dynamic capabilities (since they are discovered during runtime and are not added to the container).

src/Capability/Registry/ReferenceHandler.php
e0ipso, line 1: _Note for the reviewer:_ Alright, we made it to one of the meaty parts...

There are 2 main things going on in this change:

1. Argument preparation has been pushed to the references themselves, and then abstracted into two traits: `DynamicArgumentPreparationTrait` and `ReflectionArgumentPreparationTrait`. I _think_ this promotes better encapsulation & delagation.
2. `resolveHandler` was introduced for readability. Reflection capabilities and dynamic capabilities have inherently different ways to provide the handler to _actually_ do stuff. This helper method also serves as ergonomics to deal with that.

src/Capability/Registry/ReflectionArgumentPreparationTrait.php
e0ipso, line 1: _Note for the reviewer:_ This code used to be in the `ReferenceHandler`, I just moved it.

src/Capability/Registry/ReflectionArgumentPreparationTrait.php
```diff
@@ -0,0 +1,208 @@
...
+
+namespace Mcp\Capability\Registry;
+
+use Mcp\Exception\InvalidArgumentException;
+use Mcp\Exception\RegistryException;
+
+/**
+ * Uses reflection to match named parameters and perform type casting.
+ *
+ * @author Mateu Aguilo Bosch <mateu@mateuaguilo.com>
```
e0ipso: Ooops. I am not the author of this, I'll get this fixed.

src/Capability/Registry/ResourceReference.php
e0ipso, line 1: _Note for the reviewer:_ The simplification here comes from having extracted the formatting to a utility class.

src/Capability/Registry/ResourceTemplateReference.php
e0ipso, line 1: _Note for the reviewer:_ I moved the regular expression handling to a dedicated class following the [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle). However, I would prefer to avoid that class in favor of [`rize/uri-template`](https://github.com/rize/UriTemplate) for URI template extraction.

However, most of the simplification seen here is because we extracted the formatter code.

src/Capability/Registry/ToolReference.php
e0ipso, line 1: _Note for the reviewer:_ The simplification here comes from having extracted the formatting to a utility class.

src/Capability/Registry/UriTemplateMatcher.php
e0ipso, line 1: _Note for the reviewer:_ This was introduced to contain the URI template matching that was previously hosted in the `ResourceTemplateReference`. As stated earlier, I would prefer to avoid this class in favor of `rize/uri-template` for URI template extraction.

src/Capability/Registry.php
e0ipso, line 1: _Note for the reviewer:_ this is the other meaty place...

src/Capability/Registry.php
```diff
@@ -41,6 +51,58 @@
...
+     *     staticRegistry: string,
+     *     dynamicProviders: string,
+     *     itemsGetter: string,
+     *     keyProperty: string,
+     *     supportMethod: string,
+     *     event: class-string,
+     *     label: string,
+     * }>
+     */
+    private const CAPABILITIES = [
```
e0ipso, line 67: _Note for the reviewer:_ Initially I had similar methods implementing the same for the different capabilities. I ended up on a single solution with this config constant. Not attached to this if you think this actually harms readability.

src/Server/Handler/Request/CallToolHandler.php
```diff
@@ -71,6 +74,7 @@ public function handle(Request $request, SessionInterface $session): Response|Er
             $this->logger->debug('Tool executed successfully', [
                 'name' => $toolName,
                 'result_type' => \gettype($result),
+                'dynamic' => $reference instanceof DynamicToolReference,
```
e0ipso, line 77: _Note for the reviewer:_ This might be a bit too much implementation-detail oriented, but I feel it's fine.

src/Server/Handler/Request/CompletionCompleteHandler.php
e0ipso, line 1: _Note for the reviewer:_ This is to allow dynamic prompts and resource templates to present their completion providers.

src/Server/Handler/Request/ReadResourceHandler.php
```diff
@@ -89,4 +101,18 @@ public function handle(Request $request, SessionInterface $session): Response|Er
...
         }
     }
+
+    /**
+     * Gets the MIME type from a resource reference.
+     */
+    private function getMimeType(
+        ResourceReference|ResourceTemplateReference|DynamicResourceReference|DynamicResourceTemplateReference $reference,
+    ): ?string {
+        return match (true) {
```
e0ipso, line 111: _Note for the reviewer:_ I saw this pattern first in this repo. I like it!

src/Server/Builder.php
e0ipso, line 1: _Note for the reviewer:_ Changes here are just to account for the new dynamic providers so frameworks like Drupal can register them.

src/Capability/Formatter/PromptResultFormatter.php
chr-hertel: yea, i like that - the code itself is untouched?

src/Capability/Registry/PromptReference.php
```diff
@@ -55,228 +47,6 @@ public function __construct(
...
-                $resource['uri'],
-                $resource['mimeType'] ?? 'application/octet-stream',
-                $resource['blob']
-            );
-        } else {
-            throw new RuntimeException("Invalid resource{$indexStr}: Must contain 'text' or 'blob'.");
-        }
-
-        return new EmbeddedResource($resourceObj);
+        return (new PromptResultFormatter())->format($promptGenerationResult);
```
Nyholm, line 50: I like that this was moved to a shared class, but I dont like that we create new objects here. 

But let's leave it for now

src/Capability/Formatter/PromptResultFormatter.php
Nyholm: Yes, it is

src/Capability/Provider/DynamicPromptProviderInterface.php
Nyholm: I dont see any class but tests implementing this interface.. I assume this is a comment on an old version of the PR because `DynamicPromptProvider` does not exist. I assume this is implemented by SDK consumers. 

> What is your preference in these cases, keep the interface or drop it?

In general for open source: Keep the interface

src/Capability/Registry.php
```diff
@@ -338,6 +531,156 @@ public function getPrompt(string $name): PromptReference
...
+    }
+
+    private function findDynamicToolProvider(string $toolName): ?DynamicToolProviderInterface
+    {
+        $matching = array_filter(
+            $this->dynamicToolProviders,
+            fn (DynamicToolProviderInterface $p) => $p->supportsTool($toolName),
+        );
+
+        return array_shift($matching);
```
Nyholm, line 578: Why only the first one? Why not return `iterable<DynamicToolProviderInterface>`?

src/Capability/Registry/Loader/ArrayLoader.php
Nyholm: Can you make a PR will "just the noise"? 

I think I see a bunch of unrelated comments and doc blocks that changes. (I might be wrong) If so, please make them in a separate PR.

src/Capability/Registry/ArgumentPreparationInterface.php
Nyholm: Yes, please remove as it is never used (consumed)

src/Capability/Registry/ReferenceHandler.php
Nyholm: Yeah. I dont like this =)

I tried to solve that problem with #176. 

Using a setter like this does not work because we use Fiber and we should not keep objects in memory on the class like this.

src/Capability/Registry/ReferenceHandler.php
```diff
@@ -29,270 +29,87 @@ public function __construct(
...
+        // Resolve the handler to a callable and optionally an instance
+        [$callable, $instance] = $this->resolveHandler($reference->handler);
 
-                return \call_user_func($reference->handler, ...$arguments);
-            }
+        // Set client on reference if it implements ClientAwareInterface
+        // (e.g., dynamic references forward this to their providers)
+        if ($reference instanceof ClientAwareInterface) {
+            $reference->setClient($clientGateway);
         }
```
Nyholm, line 48: I am not sure why this is needed, can you give an example?

src/Server/Handler/Request/CallToolHandler.php
```diff
@@ -58,7 +59,9 @@ public function handle(Request $request, SessionInterface $session): Response|Er
         $this->logger->debug('Executing tool', ['name' => $toolName, 'arguments' => $arguments]);
 
         try {
-            $reference = $this->registry->getTool($toolName);
+            // Try dynamic tool first, fall back to static
+            $reference = $this->registry->getDynamicTool($toolName)
+                ?? $this->registry->getTool($toolName);
```
Nyholm, line 64: This means we only get a Tool if there are no dynamic Tools. Wouldn't that be cause for a lot of confusion?

src/Capability/Provider/DynamicToolProviderInterface.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+namespace Mcp\Capability\Provider;
+
+use Mcp\Schema\Tool;
+
+/**
+ * Provider for runtime tool discovery and execution.
+ *
+ * Implement ClientAwareInterface to access ClientGateway.
```
Nyholm, line 19: It does not =)

src/Capability/Registry/DynamicToolReference.php
```diff
@@ -0,0 +1,58 @@
...
+        public readonly string $toolName,
+    ) {
+        parent::__construct($this, false);
+    }
+
+    public function setClient(ClientGateway $clientGateway): void
+    {
+        if ($this->provider instanceof ClientAwareInterface) {
+            $this->provider->setClient($clientGateway);
+        }
```
Nyholm, line 40: Why does a DynamicToolProvider need a `ClientGateway`? Can you give an example? 

It seams like we calling the `DynamicToolProvider` -> Get tool -> Then give ClientInterface to the `DynamicToolProvider`-> then execute the tool. 

This feels wrong. Let's avoid keeping state and move references like this. 

(Shamelessly self-promoting #176)

Nyholm (changes requested):
I agree with you, we really want to have a feature like this. Thank you for submitting this PR

There are a few things I would like to do to make this more easy to review and "mergeable". 

- Move all the unrelated CS and comment fixes to a separate PR. 
- Move refactoring of src/Capability/Formatter/*ResultFormatter.php to a separate PR
- Move any other other kind of refactor to separate PRs too

One other (maybe small thing) about naming... Did you consider "Runtime" instead of "Dynamic"? Any tool class can be "dynamic" if it accepts parameters. But you always describe getting tools in runtime (ie from database). 

I dont like our current (what we do in main) way to handle client context. (Hence I created #176.) I dont think it is a good idea we should keep building on top of our current solution. We should merge #176 and then build form there. Correct me if Im wrong, but doesnt that make things easier? We would avoid passing the `ClientGateway` around.

chr-hertel:
There's a lot of great stuff here - can basically second most of what Tobias wrote - one thing that comes to my mind tho: next to this new layer we already have interface/methods with similar signature and purpose, e.g. `Mcp\Capability\Registry::getPrompts` vs. `Mcp\Capability\Provider\DynamicPromptProviderInterface::getPrompts`

in the current architecture they need to collaborate, but i just wonder if we'd in the end could just decouple the `Registry` more into verticals.

maybe having a `PromptProvider` or `PromptRepository` that has a state for prompts that can be populated as part of the initial discovery or dynamically (i favor runtime as wording more as well).

not sure if possible, but that would slim down the architecture and request handlers, that read on the Registry right now, only need on kind of capability anyways and could use an explicit `PromptProvider` dependency.

CodeWithKyrian:
+1 on splitting the Formatter/*ResultFormatter.php refactor into its own PR. The abstraction itself is solid and exactly the kind of improvement we need!

e0ipso:
Thanks for the feedback folks! It feels very good to have so many perspectives brought on this.

I will try to consolidate your words with my understanding and preferences to formulate a plan. At a glance I see:

- **Action 1**: split the PHPStan fixes to a separate PR. Likely to be merged before this one. _Status: unblocked_
- **Action 2**: split the formatter refactor to a separate PR. Likely to be merged before this one. _Status: unblocked_
- **Feedback consensus 1**: use _runtime_ wording instead of dynamic. _Status: decided_
- **Feedback suggestion 1**: split `Registry` into per-capability registries. Perhaps a shared interface + a base class + overridden classes where necessary is the pattern here? _Status: seeking maintainer consensus, please help._
- **Feedback suggestion 2**: incorporate the `RequestContext` work into this PR. _Status: blocked_

The last item seems to be a blocker, since there is still ongoing discussion for #176.

My overall sensation: I think we all want this feature. However, the current design is a bit coupled to the reflection based definition. This affects how capabilities are discovered and executed. Introducing runtime capabilities will make us choose between :one: bolt the new paradigm on the current code base with almost-duplicated methods with almost-exact signatures, or :two: refactor the current architecture to abstract the operations so they can be fulfilled by reflection-based or dynamic capabilities. Despite this sounding fancy, it is nothing more than a strategy pattern. Option :one: will cause a smaller footprint in terms of PR, option :two: will result in a more maintainable code base. Option :three: is to go with option :one: first, with the intention to do a subsequent PR to achieve :two:. Option :four: is to do the refactor first (:two:) and then introduce the runtime capabilities (:one:).

My preferences would be options :two: and :three: because we can create the abstractions with the usages of them in mind.

Thoughts?

chr-hertel:
Thanks @e0ipso for that summary - very helpful! :+1: 

> split Registry into per-capability registries. Perhaps a shared interface + a base class + overridden classes where necessary is the pattern here? Status: seeking maintainer consensus, please help.

I'd be in favor of that, yes.

My initial thought was more to go with option 4️⃣, but I see the point that it might lead to another round of changes when introducing the runtime level - otherwise I'd be more in favor of 2️⃣ - i'm currently not that afraid of larger refactorings in on go since we have quite high-level tests to ensure functionality

chr-hertel:
Hey @e0ipso, want to continue here? would need a rebase and some adoptions, i guess

e0ipso:
@chr-hertel sorry for the long delay. I am trying to negotiate sponsorships to implement this feature. I know this is taking longer than anyone expected. I hope to have something more definitive soon. I appreciate the patience you all have shown so far, and I don't take it for granted.

soyuka:
I'm unsure about the need for this, we're doing something like this on API Platform and I do also dynamic providers on another project but all I need is to hook custom handlers.

chr-hertel:
This one is obsolete with #294 and we could close, right?

e0ipso:
Correct. Let's close this.

PR #183, opened 2025-12-08
[Server] align addLoaders with addRequestHandlers

soyuka:
Use `iterable` as we did in `addRequestHandlers` and others.

This helps with framework integration and unifies code source.


soyuka:
Done @chr-hertel pending I'll change the mcp-bundle once this is merged (as we can use tagged_iterator instead of the compiler pass)

PR #190, opened 2025-12-19
Fix: use RequestContext in client-communication server example

CodeWithKyrian:
This PR updates the client-communication example to use `RequestContext` instead of the deprecated `ClientGateway` type hint.

## Motivation and Context

PR #176 introduced `RequestContext` as the sole injectable context object for tool handlers, replacing direct injection of `ClientGateway` and `Session`. While `ClientAwareService` was updated, the inline callback tool registered in `server.php` was missed.


Nyholm:
Thank you. I must have missed that.

Issue #185, opened 2025-12-12
[Client] Implement PHP MCP Client Component in php-sdk


robertoperuzzo:
## Is your feature request related to a problem? Please describe

The current PHP SDK focuses exclusively on server-side MCP implementation. However, the Model Context Protocol (MCP) specification defines both server and client sides, and a complete SDK should support both directions of communication. Without a client component, PHP applications cannot programmatically connect to and consume services from MCP servers, limiting the ecosystem's potential for integration and automation.

Currently, PHP developers looking to build MCP clients must either use standalone libraries (like [php-mcp/client](https://github.com/php-mcp/client) or [swisnl/mcp-client](https://github.com/swisnl/mcp-client)) or implement their own solutions, fragmenting the ecosystem. This issue aims to bring a unified, officially supported client implementation into the `modelcontextprotocol/php-sdk` repository as a sub-component, following the same development practices and standards as the existing server component.

## Describe the solution you'd like

Implement a `src/Client` sub-component within the php-sdk repository that provides a comprehensive, well-designed PHP client for interacting with MCP servers. The implementation should:

1. **Follow the existing SDK architecture and patterns**
   - Align with the server component's code structure and conventions
   - Adopt Symfony's coding standards and backward compatibility promise
   - Use attribute-based configuration where appropriate
   - Integrate with the existing transport layer concepts

2. **Provide both synchronous and asynchronous APIs**
   - Simple, blocking methods for straightforward use cases (`$client->listTools()`, `$client->callTool()`)
   - Promise-based async methods for advanced scenarios (`$client->listToolsAsync()`)
   - Leverage existing async infrastructure (ReactPHP or consider php-runtime/runtime for transport agnosticism)

3. **Support multiple transport mechanisms**
   - STDIO transport (for spawning and communicating with server processes)
   - HTTP/SSE transport (for communicating with remote MCP servers)
   - Extensible architecture for custom transports

4. **Handle the complete client lifecycle**
   - Connection initialization and MCP handshake
   - Capability negotiation
   - Session management
   - Graceful disconnection and cleanup

5. **Implement core MCP client operations**
   - List and call tools (`tools/list`, `tools/call`)
   - List and read resources (`resources/list`, `resources/read`, `resources/subscribe`)
   - List and get prompts (`prompts/list`, `prompts/get`)
   - Handle server notifications (e.g., `resources/updated`, `tools/list_changed`)
   - Support logging levels and progress notifications

6. **Provide robust error handling**
[...]

Issue #15, opened 2025-08-29
[Client] (Re-)Implement Client component


chr-hertel:
The MCP specification defines a client-side, and also the [php-mcp had a component](https://github.com/php-mcp/client) tackling that.

The goal of this issues is to bring in a first state of a Client as `Mcp\Client`.

---

Please announce yourself before working on this - so we can avoid duplicate effort.

stof:
should this be part of the same package or should we have separate packages for the server side and client side SDKs ? Most projects will probably need only one side of the protocol.

chr-hertel:
totally get the point - was @CodeWithKyrian and thinking as well, but i would also look at what other SDKs do here - with PHP we'd need some splitting tho for multiple packages per repo

CodeWithKyrian:
Yeah, most of the other sdks in the organization combine both in same package so it makes sense to join the trend in this case. 


It just means we have to make sure both the client and server are on the same spec version at all times

jormeijer:
From my perspective, separation makes a lot of sense. In most cases you only need one of the two, and keeping them independent avoids unnecessary dependencies. A shared layer for common definitions (DTOs, contracts) could still add consistency without tying server and client together too tightly.

I’ve already have experience with building a PHP MCP client, so I’d be happy to share insights from that journey if it’s helpful. Not to push a specific implementation, but to give some inspiration when you start designing the client side of the SDK. [swisnl/mcp-client](https://github.com/swisnl/mcp-client)

If there’s interest, I’d love to contribute or collaborate on shaping the client feature.

chr-hertel:
Hi @jormeijer, it's not that we don't see the value - but also the harm would only persist with having unnecessary or conflicting dependencies by client vs server - if it's only a bunch of PHP files. Anyhow, the motivation is to keep it in one repo, and for now without split unless we solve a real issue.

Coming to your second point, yes, contributions on the client would be great - with @CodeWithKyrian's solution in `php-mcp/client` we also have a starting point, but someone needs to take over that action item :)

butschster:
I’d actually argue in favor of **splitting client and server into two separate repositories/packages**.
The reasoning is mostly around **release policy and developer clarity**:

* **Independent release cycles**
  Very often, client and server evolve at different speeds. Having them in separate packages makes it crystal clear to users what exactly changed — if a new release is published for the client, it doesn’t imply anything changed for the server, and vice versa. This lowers confusion and gives consumers more confidence in upgrading.

[...]

PR #192, opened 2025-12-21
[Client] Feat: Implement MCP client component

CodeWithKyrian:
This PR introduces an **MCP client component** for the SDK, enabling PHP applications to connect to MCP servers (via STDIO subprocess or HTTP) and interact with server-exposed tools, resources, and prompts.

> [!IMPORTANT]
> This PR is currently in **draft mode** to gather early feedback on the API design and architecture before finalizing. The core client functionality is working and I'm opening this early to ensure alignment with the project's direction before solidifying specific implementation decisions.

## Motivation & Context

This addresses several community requests for client-side functionality (#185 and #15). The MCP ecosystem requires both servers (exposing tools, resources, prompts) and clients (consuming them). While the SDK has robust server support, this PR adds the complementary client-side implementation.

This will fix https://github.com/modelcontextprotocol/php-sdk/issues/185 and #15 

### Core Client Components

- **`Client`** — High-level API for connecting to servers, initializing sessions, calling tools, listing resources/prompts, and handling real-time notifications
- **`Builder`** — Fluent builder pattern for constructing `Client` instances with timeouts, capabilities, handlers, and logger
- **`Protocol`** — Central message dispatcher handling JSON-RPC request/response routing, server notifications, and server-initiated requests (sampling)
- **`Configuration`** — Value object holding client settings (timeouts, capabilities, retries)

### Transports

- **`StdioClientTransport`** — Spawns a subprocess and communicates via stdin/stdout (ideal for local MCP servers)
- **`HttpClientTransport`** — Connects to HTTP-based MCP servers with SSE streaming support. 
- **`ClientTransportInterface`** — Contract for custom transport implementations

### Handler Registration Approach

Unlike the server, which auto-registers internal handlers for all standard notifications/requests (allowing user overrides with precedence), the client takes a more explicit opt-in approach:

- **Auto-registered**: Only `ProgressNotificationHandler` is internal — it stores progress data so the transport's tick loop can pull and execute user callbacks
- **Opt-in**: `LoggingNotificationHandler` and `SamplingRequestHandler` must be explicitly registered by the user

This is intentional as logging and sampling are capabilities the client not only chooses to support, but has to provide a mechanism on how to handle them, so handlers are registered when needed. To reduce boilerplate, I created **convenience handlers** that accept typed callbacks:

```php
// Instead of implementing NotificationHandlerInterface manually:
->addNotificationHandler(new LoggingNotificationHandler(function (LoggingMessageNotification $n) {
    echo "[{$n->level->value}] {$n->data}\n";
}))

->addRequestHandler(new SamplingRequestHandler(function (CreateSamplingMessageRequest $req): CreateSamplingMessageResult {
    return $this->callMyLLM($req); // User implements their LLM integration
}))
```

Regardless, users can still register their own custom handlers as they choose just like the server component

### Session Management

- **`ClientSession`** — In-memory session tracking pending requests, responses, and progress data
- **`ClientSessionInterface`** — Contract for session storage

### Exceptions

- **`TimeoutException`** — Thrown when requests exceed configured timeout
- **`RequestException`** — Thrown when server returns an error response
- **`ConnectionException`** — Thrown when transport connection fails
- **`SamplingException`** - Can be thrown by user callbacks in `SamplingRequestHandler` to return custom error messages to the server. This exception's message is surfaced in the JSON-RPC error response, while other unexpected exceptions are caught and logged internally with a generic error message to avoid leaking implementation details.

### Bug Fixes

- Fixed `CreateSamplingMessageRequest::fromParams()` to properly hydrate raw JSON arrays into `SamplingMessage` objects

### Examples Reorganization

- Moved server examples to `examples/server/`
- Added client examples in `examples/client/`:
  - `stdio_client_communication.php` — Demo with logging, progress, and sampling
  - `http_client_communication.php` — Same demo over HTTP with SSE streaming
  - `stdio_discovery_calculator.php` — Tool discovery example
  - `http_discovery_calculator.php` — Same over HTTP

**Why client examples are separated by transport**: Unlike server examples (which auto-detect transport based on runtime context), client examples are inherently transport-specific — STDIO requires specifying the command and arguments to spawn, while HTTP requires an endpoint URL. Keeping them separate makes the examples cleaner and easier to follow (for now at least)

## Request for Comments

I'd appreciate feedback on a few design decisions:

### 1. Transport Naming Convention

Currently:
- **Server**: `StdioTransport`, `StreamableHttpTransport` (namespace `Mcp\Server\Transport`)
- **Client**: `StdioClientTransport`, `HttpClientTransport` (namespace `Mcp\Client\Transport`)

**Options:**
1. **Current approach**: Class names include context (`StdioClientTransport` and `StdioServerTransport` vs `StdioTransport`) — explicit but verbose
2. **Namespace-only**: Both use the same class names (`StdioTransport`), differentiated only by namespace — cleaner but requires careful imports

What naming convention would be preferred?

### 2. Handler Interface Sharing
I originally intended to have shared handler interfaces at `Mcp\Handler `(see `src/Handler/RequestHandlerInterface.php` and `src/Handler/NotificationHandlerInterface.php`) that both client and server would use. The client uses them, but server handlers need a `SessionInterface`parameter that the shared interface doesn't include.

What's the best approach here?

- **Keep separate interfaces** -  Server keeps `handle($request, $session)`, client uses interface with `handle($request)` but moved to `src/Client/Handler`. Two different `RequestHandlerInterface` definitions.
- **Unified interface with context object** - Create either `HandlerContext` object that wraps optional parameters (session, etc.), or a simple `array $context`. Both sides use handle($request, $context) where context contents differ between client/server (I'm leaning towards this, but with an array context, though it'll be a breaking change)
- **Other suggestions?** - Open to alternative patterns for sharing handler contracts while accommodating different parameter needs.

### 3. STDIO Implementation: `proc_open` vs `symfony/process`

`StdioClientTransport` uses native `proc_open()` with pipes for subprocess management. I was tempted to pull in `symfony/process` as a more mature solution, but since the server's `StdioTransport` also uses native functions, I kept it consistent.

Should we consider adopting `symfony/process` for more robust process handling esp. for weird cases (Windows and co.), or stick with native functions to minimize dependencies?

### 4. ClientSession Role

`ClientSession` currently acts as in-memory storage for:
- Pending request tracking (request ID → timestamp/timeout)
- Response buffering before fiber resumption
- Progress data consumption

Unlike server sessions, it's **not persisted**.  It exists only for the lifetime of a connection, which works well. Should this remain a simple in-memory store, or is there a use case for persistent client sessions?

Looking forward to feedback on the overall direction.

### DIFF
--- a/src/Client/Handler/Request/SamplingRequestHandler.php
+++ b/src/Client/Handler/Request/SamplingRequestHandler.php
@@ -32,16 +32,10 @@ use Psr\Log\NullLogger;
  */
 class SamplingRequestHandler implements RequestHandlerInterface
 {
-    private readonly LoggerInterface $logger;
-
-    /**
-     * @param callable(CreateSamplingMessageRequest): CreateSamplingMessageResult $callback
-     */
     public function __construct(
-        private readonly mixed $callback,
-        ?LoggerInterface $logger = null,
+        private readonly SamplingCallbackInterface $callback,
+        private readonly LoggerInterface $logger = new NullLogger(),
     ) {
-        $this->logger = $logger ?? new NullLogger();
     }
 
     public function supports(Request $request): bool
@@ -57,7 +51,7 @@ class SamplingRequestHandler implements RequestHandlerInterface
         \assert($request instanceof CreateSamplingMessageRequest);
 
         try {
-            $result = ($this->callback)($request);
+            $result = $this->callback->__invoke($request);
 
             return new Response($request->getId(), $result);
         } catch (SamplingException $e) {
--- a/examples/client/http_client_communication.php
+++ b/examples/client/http_client_communication.php
@@ -35,6 +35,7 @@ require_once __DIR__.'/../../vendor/autoload.php';
 
 use Mcp\Client;
 use Mcp\Client\Handler\Notification\LoggingNotificationHandler;
+use Mcp\Client\Handler\Request\SamplingCallbackInterface;
 use Mcp\Client\Handler\Request\SamplingRequestHandler;
 use Mcp\Client\Transport\HttpTransport;
 use Mcp\Schema\ClientCapabilities;
@@ -50,18 +51,21 @@ $loggingNotificationHandler = new LoggingNotificationHandler(static function (Lo
     echo "[LOG {$n->level->value}] {$n->data}\n";
 });
 
-$samplingRequestHandler = new SamplingRequestHandler(static function (CreateSamplingMessageRequest $request): CreateSamplingMessageResult {
-    echo "[SAMPLING] Server requested LLM sampling (max {$request->maxTokens} tokens)\n";
+$samplingRequestHandler = new SamplingRequestHandler(new class implements SamplingCallbackInterface {
+    public function __invoke(CreateSamplingMessageRequest $request): CreateSamplingMessageResult
+    {
+        echo "[SAMPLING] Server requested LLM sampling (max {$request->maxTokens} tokens)\n";
 
-    $mockResponse = 'Based on the incident analysis, I recommend: 1) Activate the on-call team, '.
-        '2) Isolate affected systems, 3) Begin root cause analysis, 4) Prepare stakeholder communication.';
+        $mockResponse = 'Based on the incident analysis, I recommend: 1) Activate the on-call team, '.
+            '2) Isolate affected systems, 3) Begin root cause analysis, 4) Prepare stakeholder communication.';
 
-    return new CreateSamplingMessageResult(
-        role: Role::Assistant,
-        content: new TextContent($mockResponse),
-        model: 'mock-gpt-4',
-        stopReason: 'end_turn',
-    );
+        return new CreateSamplingMessageResult(
+            role: Role::Assistant,
+            content: new TextContent($mockResponse),
+            model: 'mock-gpt-4',
+            stopReason: 'end_turn',
+        );
+    }
 });
 
 $client = Client::builder()
--- a/examples/client/stdio_client_communication.php
+++ b/examples/client/stdio_client_communication.php
@@ -26,6 +26,7 @@ require_once __DIR__.'/../../vendor/autoload.php';
 
 use Mcp\Client;
 use Mcp\Client\Handler\Notification\LoggingNotificationHandler;
+use Mcp\Client\Handler\Request\SamplingCallbackInterface;
 use Mcp\Client\Handler\Request\SamplingRequestHandler;
 use Mcp\Client\Transport\StdioTransport;
 use Mcp\Schema\ClientCapabilities;
@@ -39,18 +40,21 @@ $loggingNotificationHandler = new LoggingNotificationHandler(static function (Lo
     echo "[LOG {$n->level->value}] {$n->data}\n";
 });
 
-$samplingRequestHandler = new SamplingRequestHandler(static function (CreateSamplingMessageRequest $request): CreateSamplingMessageResult {
-    echo "[SAMPLING] Server requested LLM sampling (max {$request->maxTokens} tokens)\n";
+$samplingRequestHandler = new SamplingRequestHandler(new class implements SamplingCallbackInterface {
+    public function __invoke(CreateSamplingMessageRequest $request): CreateSamplingMessageResult
+    {
+        echo "[SAMPLING] Server requested LLM sampling (max {$request->maxTokens} tokens)\n";
 
-    $mockResponse = 'Based on the incident analysis, I recommend: 1) Activate the on-call team, '.
-        '2) Isolate affected systems, 3) Begin root cause analysis, 4) Prepare stakeholder communication.';
+        $mockResponse = 'Based on the incident analysis, I recommend: 1) Activate the on-call team, '.
+            '2) Isolate affected systems, 3) Begin root cause analysis, 4) Prepare stakeholder communication.';
 
-    return new CreateSamplingMessageResult(
-        role: Role::Assistant,
-        content: new TextContent($mockResponse),
-        model: 'mock-gpt-4',
-        stopReason: 'end_turn',
-    );
+        return new CreateSamplingMessageResult(
+            role: Role::Assistant,
+            content: new TextContent($mockResponse),
+            model: 'mock-gpt-4',
+            stopReason: 'end_turn',
+        );
+    }
 });
 
 $client = Client::builder()
--- a/src/Client/Handler/Request/SamplingCallbackInterface.php
+++ b/src/Client/Handler/Request/SamplingCallbackInterface.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of the official PHP MCP SDK.
+ *
+ * A collaboration between Symfony and the PHP Foundation.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Mcp\Client\Handler\Request;
+
+use Mcp\Schema\Request\CreateSamplingMessageRequest;
+use Mcp\Schema\Result\CreateSamplingMessageResult;
+
+/**
+ * Contract for callbacks used by SamplingRequestHandler.
+ *
+ * Implementations perform the actual LLM sampling when requested by the server.
+ */
+interface SamplingCallbackInterface
+{
+    public function __invoke(CreateSamplingMessageRequest $request): CreateSamplingMessageResult;
+}


composer.json
```diff
@@ -26,24 +26,26 @@
...
     "phpunit/phpunit": "^10.5",
     "psr/cache": "^3.0",
     "psr/simple-cache": "^3.0",
     "symfony/cache": "^6.4 || ^7.3 || ^8.0",
     "symfony/console": "^6.4 || ^7.3 || ^8.0",
-    "symfony/process": "^6.4 || ^7.3 || ^8.0",
-    "nyholm/psr7": "^1.8",
-    "nyholm/psr7-server": "^1.1",
-    "laminas/laminas-httphandlerrunner": "^2.12"
+    "symfony/http-client": "^7.4",
```
chr-hertel: would need a version range as well I'd say - at least to be consistent
```suggestion
    "symfony/http-client": "^6.4 || ^7.3 || ^8.0",
```

CodeWithKyrian: Oh true true!

chr-hertel:
i like this very much and can follow your reason about opting into features unlike the server side :+1:

---

1. What naming convention would be preferred?
=> I usually prefer to go short and take the namespace into account => `StdioTransport`

2. Handler Interface Sharing
  => First I thought, that'd be nice, but what is actually the benefit of sharing those interfaces? would we be able to reuse implementations on client- and server-side? if it's just for the sake of symmetric design, i would rather not do it
  => anyhow, you asked for other suggestions, here you go - just dropping them:
    * make session part of the request
    * inject session into handlers in a different way - don't use the method arguments
    * how does the client handle session? why is it injected there via constructor (which i def favor over method arguments)

3. I'd favor symfony/process tbh - less to care about and it doesn't drag in more dependencies

4. let's go incremental here - totally fine to start with in-memory first I'd say

examples/client/README.md
```diff
@@ -0,0 +1,27 @@
...
+## HTTP Client
+
+Connects to an MCP server over HTTP:
+
+```bash
+# First, start an HTTP server
+php -S localhost:8080 examples/http-discovery-userprofile/server.php
+
+# Then run the client
+php examples/client/http_discovery_userprofile.php
```
chr-hertel:
```suggestion
php examples/client/http_discovery_calculator.php
```

examples/client/README.md
```diff
@@ -0,0 +1,27 @@
...
+php examples/client/stdio_discovery_calculator.php
+```
+
+## HTTP Client
+
+Connects to an MCP server over HTTP:
+
+```bash
+# First, start an HTTP server
+php -S localhost:8080 examples/http-discovery-userprofile/server.php
```
chr-hertel: client example expects port 8000 and server example moved
```suggestion
php -S localhost:8000 examples/server/discovery-calculator/server.php
```

not sure which one you wanted:
```suggestion
php -S localhost:8000 examples/server/discovery-userprofile/server.php
```

chr-hertel:
can we please pull out the move of the current examples into the server subfolder to a separate PR?
would clean up the diff here a bit and make it easier.

CodeWithKyrian:
> can we please pull out the move of the current examples into the server subfolder to a separate PR?
> 
> would clean up the diff here a bit and make it easier.

Done, thanks to @Nyholm

CodeWithKyrian:
Thanks for the feedback @chr-hertel 

- Transport naming: Done

- Handler interface sharing: My thought was mostly symmetry since `Request`/`Notification` classes are shared Schema. But the handlers on each side handle different methods per spec, and the server genuinely needs session context. Given that, separate interfaces makes more sense so I've kept them distinct now

src/Client/Handler/Request/SamplingRequestHandler.php
```diff
@@ -0,0 +1,73 @@
...
+ */
+class SamplingRequestHandler implements RequestHandlerInterface
+{
+    private readonly LoggerInterface $logger;
+
+    /**
+     * @param callable(CreateSamplingMessageRequest): CreateSamplingMessageResult $callback
+     */
+    public function __construct(
+        private readonly mixed $callback,
```
chr-hertel: can we put an interface or at least a type on that `mixed $callback`?

i agree tho that it's a tricky thing since we can't really provide something an implementation out of the box, can we? but we do have assumptions here about how we going to invoke and what to return from that - it feels a bit brittle like that

src/Client/Handler/Notification/LoggingNotificationHandler.php
chr-hertel, line 1: i feel like this class really provides little

src/Client/Builder.php
```diff
@@ -0,0 +1,170 @@
...
+use Mcp\Schema\Implementation;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
+
+/**
+ * Fluent builder for creating Client instances.
+ *
+ * @author Kyrian Obikwelu <koshnawaza@gmail.com>
+ */
+class Builder
```
chr-hertel: is this open for extension or should be final?

src/Client/Transport/BaseTransport.php
```diff
@@ -0,0 +1,125 @@
...
+     * Perform initialization via the registered callback.
+     *
+     * @return mixed The result from the initialization callback
+     *
+     * @throws \RuntimeException If no initialize listener is registered
+     */
+    protected function handleInitialize(): mixed
+    {
+        if (!\is_callable($this->initializeCallback)) {
+            throw new \RuntimeException('No initialize listener registered');
```
chr-hertel: should be `Mcp\Exception\RuntimeException`

src/Client/Transport/TransportInterface.php
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Connect to the MCP server and perform initialization handshake.
+     *
+     * This method blocks until:
+     * - Initialization completes successfully
+     * - Connection fails (throws ConnectionException)
+     *
+     * @throws \Mcp\Exception\ConnectionException
+     */
+    public function connectAndInitialize(): void;
```
chr-hertel: maybe just `connect`?
```suggestion
    public function connect(): void;
```

chr-hertel (changes requested):
Another round of review - not quite sure yet how to tackle those client side handlers in the SDK - they are quite barebone right now, but don't have a good idea.

i don't think we need the same flexibility tho we have with tools/prompts/etc on server side - i think it's fair to make more assumptions and tell people more clear how to integrate.

src/Client/State/ClientState.php
```diff
@@ -0,0 +1,136 @@
...
+    {
+        unset($this->pendingRequests[$requestId]);
+    }
+
+    public function getPendingRequests(): array
+    {
+        return $this->pendingRequests;
+    }
+
+    public function storeResponse(int $requestId, array $responseData): void
```
xgin: According to https://www.jsonrpc.org/specification, `id` can be number, string or null. I've been testing your implementation with FastMCP and getting type error here in case the server could not receive a request properly: `{"jsonrpc":"2.0","id":"server-error","error":{ ... }}`

src/Schema/ClientCapabilities.php
```diff
@@ -34,8 +34,8 @@ public function __construct(
      * @param array{
```
xgin, line 35: https://modelcontextprotocol.io/specification/2025-06-18/schema#clientcapabilities must be object too. Please, add the typecast in `jsonSerialize` to properly handle empty array as an object in JSON.

src/Client/Transport/HttpTransport.php
```diff
@@ -0,0 +1,298 @@
...
+        }
+
+        if (!$this->activeStream->eof()) {
+            $chunk = $this->activeStream->read(4096);
+            if ('' !== $chunk) {
+                $this->sseBuffer .= $chunk;
+            }
+        }
+
+        while (false !== ($pos = strpos($this->sseBuffer, "\n\n"))) {
```
xgin, line 207: https://html.spec.whatwg.org/multipage/server-sent-events.html allows end-of-line to be one of ( cr lf / cr / lf ). In case of FastMCP running on linux, `\r\n\r\n` is received here. Please, handle this case too.

madskildegaard:
Hey, what's the status of this? Looking forward to using it!

src/Client/Handler/Request/SamplingRequestHandler.php
CodeWithKyrian: Agreed. But then I’m torn between just typing it as `Closure` (simpler) vs adding a `SamplingCallbackInterface` (clearer contract but more boilerplate for users). Which way would you prefer? Also does that interface name even make sense to you? 

Or should we just dump `SamplingRequestHandler` and `LoggingNotificationHandler` entirely and just document how the user can create their handlers to handle those requests?

src/Client/Builder.php
CodeWithKyrian: Final!

src/Client/Transport/TransportInterface.php
CodeWithKyrian: Sure

src/Client/State/ClientState.php
CodeWithKyrian: Thanks for pointing that out @xgin.  My bad. Its now int or string for requestId. Null is for notifications

src/Client/Handler/Request/SamplingRequestHandler.php
chr-hertel: i would go with the `SamplingCallbackInterface` - also with that naming

CodeWithKyrian: Fair enough. IN that case the `LoggingNotificationHandler` should be dropped and the user has to implement it themselves

PR #194, opened 2025-12-23
OAuth 2.0 authorization flow

sveneld:
## Motivation and Context
This pull request is a draft implementation of oAuth authorization. Its purpose is to outline the general approach, structure, and integration points for further discussion and iteration. At this stage, it is not intended to be a final solution, but rather a starting point for aligning on the overall direction.

## How Has This Been Tested?
I attempted to test the Microsoft oAuth authorization flow using `npx @modelcontextprotocol/inspector`.

However, due to a known authorization bug in the MCP Inspector, the authorization process cannot be completed successfully at the moment. This issue prevents full end-to-end testing via the Inspector (see https://github.com/modelcontextprotocol/inspector/issues/927).

## Breaking Changes
No breaking changes are introduced. Existing functionality remains unaffected, and no updates to user code or configuration are required.

## Types of changes
- [ ] Documentation update

## Checklist
- [] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [] I have added appropriate error handling

## Additional context
This pull request is intentionally a draft and is primarily meant to facilitate discussion around the oAuth authorization approach and architecture. The implementation may change significantly based on feedback before moving toward a final version.


chr-hertel:
Thanks so much @sveneld to working on this - I'll give it a proper read and test the next days!

Do you think we could also bring a test setup into the repo maybe on top of keycloak or fusionauth docker containers in combination with some fixtures? even if microsoft and okta/auth0 are basically the most important ones i like those local test setups for being a bit more flexible and independent or what do you think?

src/Server/Transport/OAuth2HttpTransport.php
```diff
@@ -0,0 +1,228 @@
...
+ * Implements:
+ * - Bearer token validation
+ * - Protected Resource Metadata endpoint (RFC 9728)
+ * - WWW-Authenticate challenges (RFC 6750)
+ *
+ * @extends StreamableHttpTransport
+ *
+ * @author Volodymyr Panivko <sveneld300@gmail.com>
+ */
+class OAuth2HttpTransport extends StreamableHttpTransport
```
Nyholm: This is tightly coupled to StreamableHttpTransport. What do you think about making it a part of StreamableHttpTransport instead?

chr-hertel:
Another thought that I just had since I was a bit surprised not to see any 3rd party lib - did you consider using a ready-made package or explicitly tried to avoid it?

src/Server/Transport/OAuth2HttpTransport.php
CodeWithKyrian: +1 on this too. Auth is a fundamental part of the StreamableHttpTransport

sveneld: Ok, I will add it to StreamableHttpTransport

sveneld:
> Do you think we could also bring a test setup into the repo maybe on top of keycloak or fusionauth docker containers in combination with some fixtures? even if microsoft and okta/auth0 are basically the most important ones i like those local test setups for being a bit more flexible and independent or what do you think?

I can make in examples implementation with docker compose for using keycloak 



> Another thought that I just had since I was a bit surprised not to see any 3rd party lib - did you consider using a ready-made package or explicitly tried to avoid it?

I try to avoid using 3rd party lib for less dependency. If you have any suggestion about 3rd party lib for OAuth implementation which can be used in lib write it and I will look it.

mahono:
> > Another thought that I just had since I was a bit surprised not to see any 3rd party lib - did you consider using a ready-made package or explicitly tried to avoid it?
> 
> I try to avoid using 3rd party lib for less dependency. If you have any suggestion about 3rd party lib for OAuth implementation which can be used in lib write it and I will look it.

How about integrating https://oauth2-client.thephpleague.com/ ?

There are tons of 3rd party integrations. But I don’t know if this matches the use case.

chr-hertel:
Yup, that's the one that would come to my mind as well.

@Nyholm @CodeWithKyrian @soyuka - any thoughts about adopting 3rd party here?

src/Server/Transport/OAuth2HttpTransport.php
sveneld: For implementation I offer to add middleware in StreamableHttpTransport. https://github.com/modelcontextprotocol/php-sdk/pull/218

Nyholm:
@sveneld did you mean to close this PR?

chr-hertel:
Closed in favor of #221

PR #195, opened 2025-12-24
Add a way to stop the stdioTransport in a graceful way

Nyholm:
Add a way to pragmatically stop the server

## Motivation and Context

When the server is running with stdioTransport, there is no way to stop it. With this PR, it allows Symfony to do a "reload" feature. See https://github.com/symfony/ai/pull/1280

I could have added this feature into the stdioTransport class directly, but that is less flexible. 

## How Has This Been Tested?
Local only

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to `mcp/sdk` will be documented in this file.
 
+0.2.1
+-----
+
+* Add `RunnerControl` for `StdioTransport` to allow break out from continuously listening for new input.
+
 0.2.0
 -----
 
@@ -12,7 +17,6 @@ All notable changes to `mcp/sdk` will be documented in this file.
 * Removed `ClientAwareTrait`
 * Removed `Protocol::getTransport()`
 * Added parameter for `TransportInterface` to `Protocol::processInput()`
-
 
 0.1.0
 -----


src/Server/Transport/StdioTransport.php
```diff
@@ -51,10 +55,17 @@ public function listen(): int
             $this->processInput();
             $this->processFiber();
             $this->flushOutgoingMessages();
+            if (RunnerState::RUNNING !== $this->runnerControl->getState()) {
```
chr-hertel: I wonder if this should even be part of the while expression

chr-hertel:
Is this only a STDIO concern? If so, maybe move it to `Mcp\Server\Transport\Stdio\` namespace?

src/Server/Transport/RunnerState.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * State for the transport.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+enum RunnerState
+{
+    case RUNNING;
+    case STOP_AND_DESTROY_SESSION;
+    case STOP;
```
chr-hertel: struggling a bit with the naming and if `RunnerState::STOP` really expresses the purpose - it's restarting/reloading the connection without killing/destroying the session, right?

Nyholm: It is breaking out of the while(true) loop without cleaning up sessions. 

It does not restart or reload. (That is handled by other processes)

Nyholm:
Thank you. I've updated the PR

chr-hertel (approved):
Thanks Tobias!

Do we need something in the changelog about it?

Nyholm:
Sure, that is a good idea

Nyholm:
Thank you for merging

PR #196, opened 2025-12-24
House keeping in cs fixer config and makefile

chr-hertel:
Just noticed this while working on sth else.


CodeWithKyrian:
Nice catch!

PR #197, opened 2025-12-26
Allow to install with sf 5.4

Nyholm:
Allow us to use the SDK in a SF 5.4 application

## Motivation and Context
We depend of Uid and Finder components. They are both very stable and has not changed over the years. This is minimal effort for us to support. I would be happy to increase minimum sf level if it proves a problem in the future. 

## How Has This Been Tested?
I added extra rules in the CI

## Breaking Changes
no

## Types of changes

- [ ] Documentation update

## Checklist


chr-hertel:
I'm struggling a bit with this, but for the components we're using right now it works out.

From my point of view it's alright as long as that wide range is not slowing us down or needs extra work, like Tobias in chat
> But the day we write "if sf5.4, do this..." then we should drop support

@CodeWithKyrian what do you say?

CodeWithKyrian:
I agree with you @chr-hertel....as long as it doesn't mean extra work to support, then why not 🤷. If it does at any point, then maybe we nuke it as suggested

Nyholm:
Thank you for merging

PR #198, opened 2025-12-27
Move examples to server directory

Nyholm:
This will prepare for the client implementation in https://github.com/modelcontextprotocol/php-sdk/pull/192

When reviewing #192, I thought I can make the diff smaller by moving the existing examples to `examples/server` directory (as suggested [here](https://github.com/modelcontextprotocol/php-sdk/pull/192#issuecomment-3690592318)). 

@CodeWithKyrian I hope you dont mind me helping out with this.


CodeWithKyrian:
No that's fine. Thanks for the assist ❤️

chr-hertel:
Test failure is real i'm afraid - namespaces in server.php files maybe?

Nyholm:
> Test failure is real i'm afraid - namespaces in server.php files maybe?

Yes, I move the bootstrap.php too. I see in #192, that the client is not using the bootstrap file.

CodeWithKyrian (approved):

Nyholm:
Wohoo. CI is green

PR #200, opened 2025-12-27
bump changelog version for v0.3

chr-hertel:
go with v0.3 instead


Nyholm:
I prefer to stay on 0.2.x

https://github.com/modelcontextprotocol/php-sdk/pull/201

PR #201, opened 2025-12-27
make sure StdioTransport does not break BC for 0.2.1

Nyholm:
This is a way that we can be backwards compatible with 0.2.0.


Nyholm:
Thank you for merging

PR #202, opened 2025-12-27
Be more generous with PSR versions

Nyholm:
The only difference is return types. Since we dont implement these interfaces, it does not matter for us. There is no need to restrict one version or another.


Nyholm:
Thank you for merging

PR #193, opened 2025-12-22
Remove SchemaValidator

Nyholm:
## Motivation and Context
This is more of an issue or question. 

I noticed the `SchemaValidator`. It is never used and not documented. Do we really need it in the sdk?

## Breaking Changes

Yes, removing classes is a breaking change

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #203, opened 2025-12-27
Use SchemaValidator to validate tool input

Nyholm:
Opposite of #193. Let's use the unused SchemaValidator

## Motivation and Context

Since we have an input schema, we can validate the input to match the schema. This makes it easier to write a tool. 

## How Has This Been Tested?

Unit tests and inspector

## Breaking Changes
Nope

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/Schema/JsonRpc/Error.php
+++ b/src/Schema/JsonRpc/Error.php
@@ -91,12 +91,9 @@ class Error implements MessageInterface
         return new self($id, self::METHOD_NOT_FOUND, $message);
     }
 
-    public static function forInvalidParams(string $message, string|int $id = '' /* mixed $data = null */): self
+    public static function forInvalidParams(string $message, string|int $id = '', mixed $data = null): self
     {
-        // $data parameter was added in 0.2.2 and will be default in 0.3.0
-        $data = \func_get_args()[2] ?? null;
-
-        return new self($id, self::INVALID_PARAMS, $message, data: $data);
+        return new self($id, self::INVALID_PARAMS, $message, $data);
     }
 
     public static function forInternalError(string $message, string|int $id = ''): self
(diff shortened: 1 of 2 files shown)


CodeWithKyrian:
#193 totally skipped my mind. This was exactly the purpose for the `SchemaValidator` actually. It was an artifact from my old package but I guess it wasn't integrated during tool call to validate the input against the generated JSON Schema. 


Haven't gone through the code yet, but seeing this reminded me of that...so you're definitely thinking in the right direction 👍🏼

src/Capability/Discovery/SchemaValidator.php
```diff
@@ -220,7 +220,7 @@ private function formatValidationError(ValidationError $error): string
                 break;
             case 'type':
                 $expected = implode('|', (array) ($args['expected'] ?? []));
-                $used = $args['used'] ?? 'unknown';
```
Nyholm, line 223: Looking at the `opis/json-schema` package, I cannot find that `$args['used']` is ever set.

src/Schema/JsonRpc/Error.php
```diff
@@ -91,9 +91,12 @@ public static function forMethodNotFound(string $message, string|int $id = ''):
         return new self($id, self::METHOD_NOT_FOUND, $message);
     }
 
-    public static function forInvalidParams(string $message, string|int $id = ''): self
+    public static function forInvalidParams(string $message, string|int $id = '' /* mixed $data = null */): self
```
chr-hertel: can we just do the BC break and move on to 0.3.0? PRs are piling up :)

chr-hertel:
Rebased after #153 and switched to v0.3.0

Issue #168, opened 2025-11-25
[Server] Schema not removed after injecting SessionInterface


lvluoyue:
**Describe the bug**
I noticed that `$_session` can be injected in tools, but it requires entering a session in the `inspector`. However, after entering it, it doesn't get overridden, but rather becomes `SessionInterface`.

**To Reproduce**
```php
<?php

namespace app\mcp;

use Mcp\Capability\Attribute\McpTool;
use Mcp\Server\Session\SessionInterface;

class example
{
    #[McpTool(name: 'example_tool')]
    public function exampleTool(SessionInterface $_session): array
    {
        return [
            'status' => 'ok',
            'session_id' => $_session->getId()->toRfc4122(),
        ];
    }
}

```

**Expected behavior**
Support `SessionInterface` injection and remove related `schema`

**Additional context**

<img width="1027" height="804" alt="Image" src="https://github.com/user-attachments/assets/a86d216d-3bc2-4c6c-a97a-f0cb655a5122" />

Nyholm:
This is partly solved in #176. But we should forbid $_session to be injected in favor of `RequestContext`

PR #206, opened 2025-12-27
Remove ability to use parameter name _session and _request

Nyholm:
These are reserved for internal use. This will fix https://github.com/modelcontextprotocol/php-sdk/issues/168

## Motivation and Context

The `_session` and `_request` are used internally, they are not meant to be injected via argument injection. 
An alternative to this PR would be to remove them in `ReferenceHandler::prepareArguments()`, but that would throw an exception when the tool is called. By doing it in the `SchemaGenerator` we get an exception when the server was started

## How Has This Been Tested?

I modified an example and tested. 

## Breaking Changes

No. I consider this a bug fix. See https://github.com/modelcontextprotocol/php-sdk/issues/168. 

The "feature" of naming a variable `$_session` is not documented nor covered by tests. 

## Types of changes

- [ ] Documentation update

## Checklist

- [ x New and existing tests pass locally

### DIFF
--- a/tests/Unit/Capability/Discovery/SchemaGeneratorFixture.php
+++ b/tests/Unit/Capability/Discovery/SchemaGeneratorFixture.php
@@ -412,4 +412,16 @@ class SchemaGeneratorFixture
         $inferredParam,
     ): void {
     }
+
+    public function withParameterNamedSession(string $_session): void
+    {
+    }
+
+    public function withParameterNamedSessionWithWeirdCase(string $_sesSion): void
+    {
+    }
+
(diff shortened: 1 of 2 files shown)


chr-hertel:
Nice hardening - could you please extend the `SchemaGeneratorTest` briefly to bake this in?

Nyholm:
Sure, will do

Nyholm:
I've updated the PR with a test

chr-hertel (approved):
Thanks @Nyholm!

Nyholm:
Thank you for merging

PR #207, opened 2025-12-27
Use the throwable object in $context['exception']

Nyholm:
## Motivation and Context

The current behavior IS compliant with PSR-3. But it is not what I would expect. PSR-3 reads: 

> If an Exception object is passed in the context data, it MUST be in the 'exception' key. Logging exceptions is a common pattern and this allows implementors to extract a stack trace from the exception when the log backend supports it. Implementors MUST still verify that the 'exception' key is actually an Exception before using it as such, as it MAY contain anything.

So when we pass an Exception as context, it must be under the `exception` key. We are using the `exception` key as a string. 

This PR make sure we pass the full Throwable object to the logger context. 

This PR also prints the exception message when running the examples. I found that easier to understand whats happening

## How Has This Been Tested?

Yes, with examples

## Breaking Changes

No. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/examples/server/env-variables/EnvToolHandler.php
+++ b/examples/server/env-variables/EnvToolHandler.php
@@ -24,7 +24,7 @@ final class EnvToolHandler
      * @return array<string, string|int> the result, varying by APP_MODE
      */
     #[McpTool(name: 'process_data_by_mode')]
-    public function processData(string $input, $_session): array
+    public function processData(string $input): array
     {
         $appMode = getenv('APP_MODE'); // Read from environment
 


examples/server/bootstrap.php
```diff
@@ -70,6 +70,10 @@ public function log($level, Stringable|string $message, array $context = []): vo
                 ([] === $context || !$debug) ? '' : json_encode($context),
```
chr-hertel, line 73: should we make sure, that the exception is not added in the json encoded string if there's additional handling later on? feels a bit duplicated than

Nyholm: Nice catch!

I've updated the PR

chr-hertel (approved):
Looks good, thank you!

Nyholm:
Thank you for merging

PR #216, opened 2025-12-29
Add support for binary resource in conformance server example

### LINKED ISSUE #213: [Server][Conformance] Support for resource with binary content missing
Not sure about this one, but to me it looks like the support for resources with binary content is missing, and therefore I could implement scenario `resources-read-binary`, see `examples/server/conformance/server.php`.

Reproduce:
```
make conformance-server
```
```
npx @modelcontextprotocol/conformance server --url http://localhost:8000 --scenario resources-read-binary
```

luoyue712:
close #213


chr-hertel:
Thanks @lvluoyue!

PR #218, opened 2025-12-31
Add Middleware handlers to StreamableHttpTransport

sveneld:
## Motivation and Context
This pull request gives a possibility to add different middlewares to StreamableHttpTransport for different approach.

## How Has This Been Tested?
Unit tests

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/Server/Transport/StreamableHttpTransport.php
+++ b/src/Server/Transport/StreamableHttpTransport.php
@@ -12,11 +12,14 @@
 namespace Mcp\Server\Transport;
 
 use Http\Discovery\Psr17FactoryDiscovery;
+use Mcp\Exception\InvalidArgumentException;
 use Mcp\Schema\JsonRpc\Error;
+use Mcp\Server\Transport\Http\MiddlewareRequestHandler;
 use Psr\Http\Message\ResponseFactoryInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\StreamFactoryInterface;
+use Psr\Http\Server\MiddlewareInterface;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\Uid\Uuid;
 
@@ -36,19 +39,22 @@ class StreamableHttpTransport extends BaseTransport
     /** @var array<string, string> */
     private array $corsHeaders;
 
+    /** @var list<MiddlewareInterface> */
+    private array $middleware = [];
+
     /**
-     * @param array<string, string> $corsHeaders
+     * @param array<string, string>         $corsHeaders
+     * @param iterable<MiddlewareInterface> $middleware
      */
     public function __construct(
-        private readonly ServerRequestInterface $request,
+        private ServerRequestInterface $request,
         ?ResponseFactoryInterface $responseFactory = null,
         ?StreamFactoryInterface $streamFactory = null,
         array $corsHeaders = [],
         ?LoggerInterface $logger = null,
+        iterable $middleware = [],
     ) {
         parent::__construct($logger);
-        $sessionIdString = $this->request->getHeaderLine('Mcp-Session-Id');
-        $this->sessionId = $sessionIdString ? Uuid::fromString($sessionIdString) : null;
 
         $this->responseFactory = $responseFactory ?? Psr17FactoryDiscovery::findResponseFactory();
         $this->streamFactory = $streamFactory ?? Psr17FactoryDiscovery::findStreamFactory();
@@ -59,6 +65,13 @@ class StreamableHttpTransport extends BaseTransport
             'Access-Control-Allow-Headers' => 'Content-Type, Mcp-Session-Id, Mcp-Protocol-Version, Last-Event-ID, Authorization, Accept',
             'Access-Control-Expose-Headers' => 'Mcp-Session-Id',
         ], $corsHeaders);
+
+        foreach ($middleware as $m) {
+            if (!$m instanceof MiddlewareInterface) {
+                throw new InvalidArgumentException('Streamable HTTP middleware must implement Psr\\Http\\Server\\MiddlewareInterface.');
+            }
+            $this->middleware[] = $m;
+        }
     }
 
     public function send(string $data, array $context): void
@@ -69,17 +82,17 @@ class StreamableHttpTransport extends BaseTransport
 
     public function listen(): ResponseInterface
     {
-        return match ($this->request->getMethod()) {
-            'OPTIONS' => $this->handleOptionsRequest(),
-            'POST' => $this->handlePostRequest(),
-            'DELETE' => $this->handleDeleteRequest(),
-            default => $this->createErrorResponse(Error::forInvalidRequest('Method Not Allowed'), 405),
-        };
+        $handler = new MiddlewareRequestHandler(
+            $this->middleware,
+            \Closure::fromCallable([$this, 'handleRequest']),
+        );
+
+        return $this->withCorsHeaders($handler->handle($this->request));
     }
 
     protected function handleOptionsRequest(): ResponseInterface
     {
-        return $this->withCorsHeaders($this->responseFactory->createResponse(204));
+        return $this->responseFactory->createResponse(204);
     }
 
     protected function handlePostRequest(): ResponseInterface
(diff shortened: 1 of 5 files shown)


sveneld:
@chr-hertel please take a look

chr-hertel:
On the first glance, i feel like it makes it more sophisticated, is it needed tho?
The idea would be that frameworks could register their own middleware and solve auth on their own? or would you rework #194 to use that middleware than? or both eventually i guess?

maybe i'm missing the point here ...

sveneld:
Yep, that’s the idea. #194 will be rewritten on top of this middleware approach, and it also opens the door for community/custom middlewares without patching core.

sveneld:
@chr-hertel https://github.com/modelcontextprotocol/php-sdk/pull/221 is a pull request with oAuth implementation based on middleware. I created a new pull request instead of update previous

sveneld:
@chr-hertel maybe we should go step by step and this pull request will be first part for oauth implementing?

chr-hertel:
In the beginning I was a bit conflicted about bringing in this extension point, bc it felt like we're just doing it for that one case, but it's fair to assume there might be more - and yes, this would be the first step, auth second.

WDYT @Nyholm @CodeWithKyrian @soyuka? (see #221 for more context)

src/Server/Transport/StreamableHttpTransport.php
```diff
@@ -59,6 +64,13 @@ public function __construct(
             'Access-Control-Allow-Headers' => 'Content-Type, Mcp-Session-Id, Mcp-Protocol-Version, Last-Event-ID, Authorization, Accept',
             'Access-Control-Expose-Headers' => 'Mcp-Session-Id',
         ], $corsHeaders);
+
+        foreach ($middlewares as $middleware) {
+            if (!$middleware instanceof MiddlewareInterface) {
+                throw new \InvalidArgumentException('Streamable HTTP middleware must implement Psr\\Http\\Server\\MiddlewareInterface.');
```
chr-hertel: should be `Mcp\Exception\InvalidArgumentException` instead of the native one

src/Server/Transport/StreamableHttpTransport.php
```diff
@@ -242,15 +252,65 @@ protected function createErrorResponse(Error $jsonRpcError, int $statusCode): Re
...
+
+                public function handle(ServerRequestInterface $request): ResponseInterface
+                {
+                    return $this->middleware->process($request, $this->handler);
+                }
+            };
+        }
+
+        return $handler;
+    }
```
chr-hertel: does it need to be an anonymous class here?

src/Server/Transport/StreamableHttpTransport.php
sveneld: fixed

src/Server/Transport/StreamableHttpTransport.php
sveneld: it is a compact way to create handlers and do not open private method handleRequest

src/Server/Transport/StreamableHttpTransport.php
```diff
@@ -36,19 +39,22 @@ class StreamableHttpTransport extends BaseTransport
     /** @var array<string, string> */
     private array $corsHeaders;
 
+    /** @var list<MiddlewareInterface> */
+    private array $middlewares = [];
+
     /**
      * @param array<string, string> $corsHeaders
+     * @param iterable<MiddlewareInterface> $middlewares
```
chr-hertel: that's the pipeline failure:

```suggestion
     * @param array<string, string>         $corsHeaders
     * @param iterable<MiddlewareInterface> $middlewares
```

src/Server/Transport/StreamableHttpTransport.php
chr-hertel: ah understood - so alternative to marking it `@internal` basically, get the intention then :+1:

src/Server/Transport/StreamableHttpTransport.php
```diff
@@ -36,19 +39,22 @@ class StreamableHttpTransport extends BaseTransport
...
      */
     public function __construct(
-        private readonly ServerRequestInterface $request,
+        private ServerRequestInterface $request,
         ?ResponseFactoryInterface $responseFactory = null,
         ?StreamFactoryInterface $streamFactory = null,
         array $corsHeaders = [],
         ?LoggerInterface $logger = null,
+        iterable $middlewares = [],
```
Nyholm: Grammar police here 👋🏽 

The plural of "middleware" is "middleware". 

=)

src/Server/Transport/StreamableHttpTransport.php
Nyholm: Let's make this a proper (internal) class

Nyholm (changes requested):
I like the idea. 

It will help making the SDK more extendable. I have a few small suggestions I want you to address. 

Good work!

src/Server/Transport/StreamableHttpTransport.php
sveneld: fixed

src/Server/Transport/StreamableHttpTransport.php
sveneld: take a look

src/Server/Transport/StreamableHttpTransport.php
```diff
@@ -242,15 +255,45 @@ protected function createErrorResponse(Error $jsonRpcError, int $statusCode): Re
...
+
+        return match ($request->getMethod()) {
+            'OPTIONS' => $this->handleOptionsRequest(),
+            'POST' => $this->handlePostRequest(),
+            'DELETE' => $this->handleDeleteRequest(),
+            default => $this->createErrorResponse(Error::forInvalidRequest('Method Not Allowed'), 405),
+        };
+    }
+
+    private function createRequestHandler(): RequestHandlerInterface
```
Nyholm: Can you move the 4 lines of logic here to the `listen()` method?

Nyholm (commented):
Thank you. I like this PR. 

I would like us to consider the stack trace. If I add 14 middleware, then I go in to `StreamableHttpTransport::handleRequest()`, would I like to see these 14 middleware + 14 `MiddlewareRequestHandler` in my stack trace? 

(Im thinking out loud) 

What are the pros and cons for creating a RequestHandler looking something like

```php
class MiddlewareRequestHandler implements RequestHandlerInterface
{
    /**
     * @param MiddlewareInterface[] $middleware
     */
    public function __construct(
        private array $middleware,
        private \Closure $application,
    ) {
    }

    public function handle(ServerRequestInterface $request): ResponseInterface
    {
        $middleware = array_shift($this->middleware);
        if (null === $middleware) {
            return ($this->application)($request);
        }

        return $middleware->process($request, $this);
    }
}
```

With this only "active" middleware is in the stack. And there is no extra `RequestHandler`.

src/Server/Transport/StreamableHttpTransport.php
sveneld: in this case i should make createRequestHandler protected. Is it ok?

sveneld:
> I would like us to consider the stack trace. If I add 14 middleware, then I go in to `StreamableHttpTransport::handleRequest()`, would I like to see these 14 middleware + 14 `MiddlewareRequestHandler` in my stack trace?

Yes, the stack trace will contain 14+ middleware, because they are called one by one.


> What are the pros and cons for creating a RequestHandler looking something like
Thank for idea.

I updated a pull request and made one commit.

src/Server/Transport/StreamableHttpTransport.php
Nyholm: Im not sure what you mean. 

I suggest to update the `listen` method like: 

```php
 public function listen(): ResponseInterface
    {
        $handler = new MiddlewareRequestHandler(
            $this->middleware,
            \Closure::fromCallable([$this, 'handleRequest']),
        );;

        return $this->withCorsHeaders($handler->handle($this->request));
    }
```

sveneld: updated

sveneld:
> Thank you.
> 
> Before we release next version, I think we should move CORS to a middleware (Not in this PR)

Ok. Than i will be wait for a merge of this pull request

Nyholm:
Thank you

PR #220, opened 2026-01-03
[Server] Add missing handler for resource subscribe/unsubscribe handlers

bigdevlarry:
Add missing handler for resource subscribe/unsubscribe handlers

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/Capability/Registry.php
+++ b/src/Capability/Registry.php
@@ -352,7 +352,7 @@ final class Registry implements RegistryInterface
     }
 
     /**
-     * Set discovery state, replacing all discovered elements.
+     * Set the discovery state, replacing all discovered elements.
      * Manual elements are preserved.
      */
     public function setDiscoveryState(DiscoveryState $state): void
--- a/src/Server/Resource/SubscriptionManagerInterface.php
+++ b/src/Server/Resource/SubscriptionManagerInterface.php
@@ -1,0 +1,53 @@
+<?php
+
+/*
+ * This file is part of the official PHP MCP SDK.
+ *
+ * A collaboration between Symfony and the PHP Foundation.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Mcp\Server\Resource;
+
+use Mcp\Server\Protocol;
+use Mcp\Server\Session\SessionInterface;
+use Psr\SimpleCache\InvalidArgumentException;
+
+/**
+ * Resource subscription interface.
+ *
+ * @author Larry Sule-balogun <suleabimbola@gmail.com>
+ */
+interface SubscriptionManagerInterface
+{
+    /**
+     * Subscribes a session to a specific resource URI.
+     *
+     * @throws InvalidArgumentException
+     */
+    public function subscribe(SessionInterface $session, string $uri): void;
+
+    /**
+     * Unsubscribes a session from a specific resource URI.
+     *
+     * @throws InvalidArgumentException
+     */
+    public function unsubscribe(SessionInterface $session, string $uri): void;
+
+    /**
+     * Check if a session is subscribed to a resource URI.
+     *
+     * @throws InvalidArgumentException
+     */
+    public function isSubscribed(SessionInterface $session, string $uri): bool;
+
+    /**
+     * Notifies all sessions subscribed to the given resource URI that the
+     * resource has changed. Sends a ResourceUpdatedNotification for each subscriber.
+     *
+     * @throws InvalidArgumentException
+     */
+    public function notifyResourceChanged(Protocol $protocol, SessionInterface $session, string $uri): void;
+}
--- a/tests/Conformance/server.php
+++ b/tests/Conformance/server.php
@@ -9,6 +9,8 @@
  * file that was distributed with this source code.
  */
 
+ini_set('display_errors', '0');
+
 require_once dirname(__DIR__, 2).'/vendor/autoload.php';
 
 use Http\Discovery\Psr17Factory;
@@ -51,7 +53,6 @@ $server = Server::builder()
     ->addResource(static fn () => 'This is the content of the static text resource.', 'test://static-text', 'static-text', 'A static text resource for testing')
     ->addResource(static fn () => fopen('data://image/png;base64,'.Elements::TEST_IMAGE_BASE64, 'r'), 'test://static-binary', 'static-binary', 'A static binary resource (image) for testing')
     ->addResourceTemplate([Elements::class, 'resourceTemplate'], 'test://template/{id}/data', 'template', 'A resource template with parameter substitution', 'application/json')
-    // TODO: Handler for resources/subscribe and resources/unsubscribe
     ->addResource(static fn () => 'Watched resource content', 'test://watched-resource', 'watched-resource', 'A resource that can be watched')
     // Prompts
     ->addPrompt(static fn () => [['role' => 'user', 'content' => 'This is a simple prompt for testing.']], 'test_simple_prompt', 'A simple prompt without arguments')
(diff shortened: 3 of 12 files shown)


tests/Conformance/server.php
```diff
@@ -9,10 +9,13 @@
  * file that was distributed with this source code.
  */
 
+ini_set('display_errors', '0');
```
bigdevlarry, line 12: This prevents PHP from printing errors to the response body fixes the JSON-RPC error of outputting html

src/Capability/Registry.php
```diff
@@ -61,6 +64,11 @@ final class Registry implements RegistryInterface
      */
     private array $resourceTemplates = [];
 
+    /**
+     * @var array<string, array<string, SessionInterface>>
+     */
+    private array $resourceSubscriptions = [];
```
bigdevlarry: How do we make the subscription persistent to make it work across multiple request?

chr-hertel: i'd say with the session

src/Capability/Registry.php
chr-hertel, line 1: Is there a specific reason to bring this into the registry?
from a cohesion point of view those methods there is no direct connection or i fail to see it.
i mean, i think this could be a standalone service instead of extending the `Registry`, which is large already

src/Capability/Registry.php
```diff
@@ -391,6 +399,64 @@ public function setDiscoveryState(DiscoveryState $state): void
...
+                    continue;
+                }
+
+                if (!isset($sessionArray['resource_subscriptions'][$uri])) {
+                    continue;
+                }
+
+                $session = $this->sessionFactory->createWithId($sessionId, $this->sessionStore);
+                $protocol->sendNotification(new ResourceUpdatedNotification($uri), $session);
+            } catch (\Throwable $e) {
```
chr-hertel: please catch an explicit exception instead of a global one - better to be specific and let global exception handling happen in a global point

src/Server/Session/FileSessionStore.php
```diff
@@ -150,6 +150,44 @@ public function gc(): array
...
+            }
+
+            $mtime = @filemtime($path) ?: 0;
+            if (($now - $mtime) > $this->ttl) {
+                continue;
+            }
+
+            try {
+                $sessionIds[] = Uuid::fromString($entry);
+            } catch (\Throwable) {
```
chr-hertel: `Uuid::fromString` only throws `Symfony\Component\Uid\Exception\InvalidArgumentException` - please be specific

src/Server/Session/FileSessionStore.php
```diff
@@ -150,6 +150,44 @@ public function gc(): array
...
     }
 
+    public function getAllSessionIds(): array
+    {
+        $sessionIds = [];
+        $now = $this->clock->now()->getTimestamp();
+
+        $dir = @opendir($this->directory);
+        if (false === $dir) {
+            return $sessionIds;
```
chr-hertel:
```suggestion
            return  [];
```

src/Server/Session/FileSessionStore.php
```diff
@@ -150,6 +150,44 @@ public function gc(): array
         return $deleted;
     }
 
+    public function getAllSessionIds(): array
+    {
+        $sessionIds = [];
+        $now = $this->clock->now()->getTimestamp();
```
chr-hertel: let's move this closer to where it's needed => before the while loop

src/Server/Session/FileSessionStore.php
```diff
@@ -150,6 +150,44 @@ public function gc(): array
...
+
+        $dir = @opendir($this->directory);
+        if (false === $dir) {
+            return $sessionIds;
+        }
+
+        while (($entry = readdir($dir)) !== false) {
+            // Skip dot entries
```
chr-hertel: don't think that comment is needed - code is quite clear here

src/Capability/Registry.php
bigdevlarry: Hi @chr-hertel this make sense, and i agree on making it a standalone. I’ll create a `src/Server/Resource` directory with a `ResourceSubscription` class and interface, then follow up on the remaining comments :)

src/Server/Session/FileSessionStore.php
bigdevlarry: Ah ! I've seen the usage this way hence my reasoning. I've given it a closer look now and I've seen the Uuid extends AbstractUid which throws the `InvalidArgumentException`. So I'll update this and other usage in the sdk, good shout

chr-hertel (changes requested):
Hey @bigdevlarry, I'm afraid we need to rollback a bit here - you really needed to go an extra mile to bring in the cross-session support, but i have some doubts about the `getAllSessionIds`. I think while integrating the SDK some will just replace the built-in session with something they have in their application infrastucture, and that might get tricky with that method - you already had to do some extra work here.

I'd like to ask you for this:
* let's rename the `Mcp\Server\Resource\ResourceSubscriptionInterface` to `Mcp\Server\Resource\SubscriptionManagerInterface`
* let's rename the `Mcp\Server\Resource\ResourceSubscription` to `Mcp\Server\Resource\SessionSubscriptionManager` as one concrete implementation we ship out of the box, which only supports notifications per session

we would add a disclaimer as code comment and in documentation, but also the openness to inject custom built - potentially cross client - `SubscriptionManagerInterface` implementations - you basically enable that already with the method on the `Builder` that you added -> rename to `Builder::setResourceSubscriptionManager`

On top, please rebase to get rid of pipeline failure, also see #230.

Thanks again for working on this! :)

bigdevlarry:
Hey @chr-hertel I've implemented the following feedback 

- Renamed `Mcp\Server\Resource\ResourceSubscriptionInterface` → `SubscriptionManagerInterface`

- Renamed `ResourceSubscription` → `SessionSubscriptionManager` as the default concrete implementation

- Updated the builder API to `Builder::setResourceSubscriptionManager`

- Added code comments to `SessionSubscriptionManager` stating that the shipped implementation is session-scoped only, while allowing consumers to inject their own (including cross-client / cross-session) managers

This aligns with your suggestion to keep session-specific mechanics out of the subscription contract while still supporting the built-in session-based notifier which is brilliant :)

src/Server/Session/SessionStoreInterface.php
```diff
@@ -61,4 +61,11 @@ public function destroy(Uuid $id): bool;
      * @return Uuid[]
      */
     public function gc(): array;
+
+    /**
+     * Get all active session IDs.
+     *
+     * @return Uuid[]
+     */
+    public function getAllSessionIds(): array;
```
chr-hertel: is this still needed when we reduce the scope to only notify within the same session? there is no need to iterate over all sessions anymore, right?
or at least that was my intention, described in https://github.com/modelcontextprotocol/php-sdk/pull/220#pullrequestreview-3735587872

but just realized, that my comment was indeed a bit confusing since i didn't mention removing that new method again

src/Server/Resource/SubscriptionManagerInterface.php
```diff
@@ -0,0 +1,46 @@
...
+     */
+    public function unsubscribe(SessionInterface $session, string $uri): void;
+
+    /**
+     * Notifies all sessions subscribed to the given resource URI that the
+     * resource has changed. Sends a ResourceUpdatedNotification for each subscriber.
+     *
+     * @throws InvalidArgumentException
+     */
+    public function notifyResourceChanged(Protocol $protocol, string $uri): void;
```
chr-hertel: regarding that notify method, i was curious about injecting the `Protocol` here and while tackling #234 I understood why you went with it - it's a bit hard to access that.

Happy to hear your thoughts about that other PR, we could adopt that event handling here as well.

src/Server/Session/SessionStoreInterface.php
bigdevlarry: Gotcha .. I wasn't sure if this was the preferred scope. I've added it now

src/Server/Resource/SubscriptionManagerInterface.php
bigdevlarry: I agree with the proposed approach; we could adopt an event listener approach here rather than the current approach, which requires injecting Session alongside the Protocol. I'll add my thoughts to #234

PR #233, opened 2026-02-01
[Server] Throw package specific exceptions in CallbackStream

chr-hertel:
minor, but still, we only want to throw package specific exceptions.


Nyholm (approved):
Good. 

Thank you

PR #234, opened 2026-02-01
[Server] Wire up notifications for changing prompt, resource & tools lists

chr-hertel:
This is a working draft for emitting change notifications on dynamically added prompts, resources and tools.

The main issue here is that we need that indirection between Registry, source of truth on changing lists, and Protocol, that is able so send notifications. That's why we have those events, but we never really adopted them.

I introduced a slim EventDispatcher/ListenerProvider setup, but it feels a bit flaky - see wire up in example. The service landscape within the SDK just grows in complexity.

**TBD**

1. Event Dispatcher
The issue I had was mostly about PSR-14 - we want to empower library users to inject their own event dispatcher, to easily register subscriber/listener/whatever - but with the current implementation we also want to add our own listeners for triggering the notifications (again, we need a decoupling between registry and protocol currently).
With PSR-14 the `EventDispatcherInterface` used in the `Registry` doesn't offer us to register our own listeners after a user injected it via `Builder::setEventDispatcher()`

2. Session ID State
I need to call the `Protocol::sendNotification(...)` from the listeners, but didn't have the session at hand - and there is basically no state in the Protocol - which is great, but we don't have a service concept to retrieve the session - it is only state in the transport or arguments looped through half the SDK.
I went for introducing the session as state in the `Protocol` - needs better null-checks tho, but wanted to check with @CodeWithKyrian first - the other option would be to keep the transport as state, since it also has the session already as state.

WDYT?

edit: this currently messes up the session handling within a fiber intermediate client request.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,7 @@
   "autoload-dev": {
     "psr-4": {
       "Mcp\\Example\\Server\\CachedDiscovery\\": "examples/server/cached-discovery/",
+      "Mcp\\Example\\Server\\ChangeEvents\\": "examples/server/change-events/",
       "Mcp\\Example\\Server\\ClientCommunication\\": "examples/server/client-communication/",
       "Mcp\\Example\\Server\\ClientLogging\\": "examples/server/client-logging/",
       "Mcp\\Example\\Server\\CombinedRegistration\\": "examples/server/combined-registration/",
--- a/examples/server/change-events/ListChangingHandlers.php
+++ b/examples/server/change-events/ListChangingHandlers.php
@@ -1,0 +1,66 @@
+<?php
+
+/*
+ * This file is part of the official PHP MCP SDK.
+ *
+ * A collaboration between Symfony and the PHP Foundation.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Mcp\Example\Server\ChangeEvents;
+
+use Mcp\Capability\RegistryInterface;
+use Mcp\Schema\Content\PromptMessage;
+use Mcp\Schema\Content\TextContent;
+use Mcp\Schema\Enum\Role;
+use Mcp\Schema\Prompt;
+use Mcp\Schema\Resource;
+use Mcp\Schema\Tool;
+
+final class ListChangingHandlers
+{
+    public function __construct(
+        private readonly RegistryInterface $registry,
+    ) {
+    }
+
+    public function addPrompt(string $name, string $content): string
+    {
+        $this->registry->registerPrompt(
+            new Prompt($name),
+            static fn () => [new PromptMessage(Role::User, new TextContent($content))],
+            isManual: true,
+        );
+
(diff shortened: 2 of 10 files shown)


soyuka:
I'm working on an MCP server that dynamically registers tools at runtime (tools are discovered progressively as the user navigates an API), so I'm hitting exactly the two problems you describe. Wanted to share what I ended up doing.

## Notification delivery via Fiber::suspend()

I don't use the event system — I use `Fiber::suspend()` directly from the tool handler after a batch of registrations:

```php
// After registering N tools in a loop:
if ($newToolsRegistered) {
    \Fiber::suspend([
        'type' => 'notification',
        'notification' => new ToolListChangedNotification(),
    ]);
}
```

This already works with the current SDK and has the advantage of **batching** — one notification after registering all discovered tools, rather than one event per `registerTool()` call.

Worth considering: the event-based approach and the Fiber::suspend approach need to coexist cleanly. If `registerTool()` synchronously dispatches an event that calls `Protocol::sendNotification()` while we're inside a Fiber handler, there could be reentrancy issues with the session state on Protocol.

## Session state (your TBD #2)

This was the hardest part for me too. I needed session access outside of the handler to persist tool state across HTTP requests.

What I ended up doing:

- **For write paths** (during tool handling): the session comes from the handler context. API Platform now propagates it via `$context['mcp_session']`: [api-platform/core@86b97d5](https://github.com/api-platform/core/commit/86b97d5eab188c5090d27a87cb310841a559b408)
- **For read paths** (e.g. restoring tools on a subsequent request): I resolve the session from `SessionStoreInterface` using the `Mcp-Session-Id` header, with an in-memory cache fallback for stdio.

Storing session on Protocol (your current approach) makes sense as the central place, but I'd flag the Fiber concern — if `$this->session` gets set to `null` after `save()` in `processInput()`, but a Fiber is still suspended and later resumed, the session reference may be gone.

## Thoughts on the PR

1. **`listChanged` capabilities always `true`** — I'd keep these conditional. A static server that never calls `registerTool(..., isManual: true)` shouldn't advertise `toolsListChanged`. Clients may behave differently (polling, etc.) based on this capability.

2. **Per-registration events vs. batch** — A single tool call in my server can discover 5-10 new tools at once. Getting one `ToolListChangedEvent` per `registerTool()` call would be noisy. Maybe a `Registry::flush()` / deferred dispatch pattern, or let consumers opt into batching?

3. **PSR-14 complexity** — I share your concern about the growing service landscape. The `ListenerProvider` + `Dispatcher` + `ChangeListener` + `Builder` wiring is a lot of ceremony for what is essentially "notify the client when the tool list changes." The Fiber::suspend path already handles the in-handler case with zero new abstractions.

Related: #251 — I also ran into `clear()` wiping dynamically registered tools and needed to track discovered state on references.

Note: LLM generated comment based on my work I hope you're fine with it I thought it'd be nice to share my solutions :)

chr-hertel:
@soyuka pretty much valid, yes - i guess a fiber based approach makes more sense here - feels more consistent as well - will close here for now - topic is up for grabs

PR #238, opened 2026-02-08
[Server] Shift away session decode from constructor

chr-hertel:
Goal was to shift that `json_decode` away from construct, but maybe a bit opinionated, and arguable when it comes to robustness due to the referenced array handling atm, but not a big fan of doing work like that in the constructor.

### DIFF
--- a/tests/Unit/Server/Session/SessionTest.php
+++ b/tests/Unit/Server/Session/SessionTest.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the official PHP MCP SDK.
+ *
+ * A collaboration between Symfony and the PHP Foundation.
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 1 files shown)


src/Server/Session/Session.php
```diff
@@ -22,23 +22,20 @@
...
      * - log_level: string|null
+     *
+     * @var array<string, mixed>
      */
+    private array $data;
+
     public function __construct(
-        protected SessionStoreInterface $store,
-        protected Uuid $id = new UuidV4(),
-        protected array $data = [],
```
Nyholm, line 37: This parameter was used only if we could not read from the store... 

Seams weird. It is a BC break to remove it (I've updated the label)

Nyholm:
Thank you

chr-hertel:
Thanks for taking care @Nyholm :pray:

Issue #239, opened 2026-02-10
remove final keyword on MCP attributes classes


fox-john:
# Feature Request: Remove `final` keyword from MCP Attributes classes

## **Is your feature request related to a problem? Please describe.**
Yes. Currently, the classes `McpPrompt`, `McpResource`, and `McpResourceTemplate` are defined with the `final` keyword. This strictly prevents inheritance.

In my current implementation, I am building a **CustomDiscoverer**. I need to extend these classes to add custom metadata and specialized logic required by my discovery process. The `final` restriction makes it impossible to leverage polymorphism, forcing me to use hacky workarounds or duplicate code.

## **Describe the solution you'd like**
I would like to remove the `final` keyword from the following classes:
* `McpPrompt.php`
* `McpResource.php`
* `McpResourceTemplate.php`

This change will allow these core entities to be extended by developers who need to implement custom behavior while staying compatible with the base types expected by the SDK.

## **Describe alternatives you've considered**
* **Duplicating Classes:** This is a maintenance nightmare and breaks interoperability with the rest of the library.

## **Additional context**
Removing `final` aligns with the "Open-Closed Principle" (open for extension, closed for modification), especially for DTO-like or Entity-like classes in a SDK that aims to be flexible. 

This is particularly critical for:
1. Customizing how resources are identified.
2. Injecting additional context into prompts during the discovery phase.
3. Supporting advanced use cases in a `CustomDiscoverer` implementation.

PR #240, opened 2026-02-10
chore: remove final keyword on attributes classes

fox-john:
This PR removes the `final` keyword from three core classes: `McpPrompt`, `McpResource`, and `McpResourceTemplate`. This change enables developers to extend these classes to implement custom logic, particularly within discovery processes.

## Motivation and Context
The current use of the `final` keyword prevents any inheritance, which is too restrictive for advanced SDK integrations. 

Specifically, when implementing a **CustomDiscoverer**, it is often necessary to extend these entities to:
- Inject additional metadata or custom attributes.
- Override or supplement default behavior without duplicating the entire class logic.
- Maintain type compatibility with the SDK's internal methods that expect these specific types.

By removing `final`, we follow the **Open-Closed Principle**, allowing for extension without modifying the core library code.

## How Has This Been Tested?
- **Inheritance Test:** Created child classes for `McpPrompt`, `McpResource`, and `McpResourceTemplate` in a local project to verify they can be instantiated and used.
- **SDK Compatibility:** Integrated these extended classes into a `CustomDiscoverer` implementation.
- **Unit Tests:** Ran existing test suites to ensure that removing the keyword does not introduce regressions in the base functionality.

## Breaking Changes
**None.** This is a non-breaking change. Existing code using these classes will continue to work exactly as before. It simply unlocks new possibilities for developers wishing to extend them.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
While `final` is often used to prevent fragile base class problems, these

Fix https://github.com/modelcontextprotocol/php-sdk/issues/239

### DIFF
--- a/src/Capability/Attribute/McpPrompt.php
+++ b/src/Capability/Attribute/McpPrompt.php
@@ -20,7 +20,7 @@ use Mcp\Schema\Icon;
  * @author Kyrian Obikwelu <koshnawaza@gmail.com>
  */
 #[\Attribute(\Attribute::TARGET_METHOD | \Attribute::TARGET_CLASS)]
-final class McpPrompt
+class McpPrompt
 {
     /**
      * @param ?string               $name        overrides the prompt name (defaults to method name)
--- a/src/Capability/Attribute/McpResource.php
+++ b/src/Capability/Attribute/McpResource.php
@@ -21,7 +21,7 @@ use Mcp\Schema\Icon;
  * @author Kyrian Obikwelu <koshnawaza@gmail.com>
  */
 #[\Attribute(\Attribute::TARGET_METHOD | \Attribute::TARGET_CLASS)]
-final class McpResource
+class McpResource
 {
     /**
      * @param string                $uri         The specific URI identifying this resource instance. Must be unique within the server.
--- a/src/Capability/Attribute/McpResourceTemplate.php
+++ b/src/Capability/Attribute/McpResourceTemplate.php
@@ -20,7 +20,7 @@ use Mcp\Schema\Annotations;
  * @author Kyrian Obikwelu <koshnawaza@gmail.com>
  */
 #[\Attribute(\Attribute::TARGET_METHOD | \Attribute::TARGET_CLASS)]
-final class McpResourceTemplate
+class McpResourceTemplate
 {
     /**
      * @param string                $uriTemplate the URI template string (RFC 6570)


Nyholm:
Hello. 

Thank you for this PR. I think those `final` was put there to specifically to avoid others to extend it. 

>  This change enables developers to extend these classes to implement custom logic, particularly within discovery processes.

Can you give a concrete example of what you are trying to achieve?

fox-john:
> Hello.
> 
> Thank you for this PR. I think those `final` was put there to specifically to avoid others to extend it.
> 
> > This change enables developers to extend these classes to implement custom logic, particularly within discovery processes.
> 
> Can you give a concrete example of what you are trying to achieve?

Hello @Nyholm,

Thank you for your feedback.

Here is a concrete example: I am currently developing an SDK implementation within a plugin ecosystem (PrestaShop). As part of this, I've built a `CustomDiscoverer` that automates the registration and synchronization of tools, prompts, and resources into a local database whenever a new module is detected.

The motivation behind removing the `final` keyword is twofold:

1. **Metadata Enrichment**: By extending these classes, I can inject ecosystem-specific metadata (such as internal module IDs or specific configuration flags) directly into the objects handled during the discovery process, while maintaining type compatibility with SDK methods that expect these specific types.
2. **Abstractions for Third-Party Developers**: I am providing an abstraction layer for third-party module developers. By extending your attribute classes, I can offer my own attributes for developers to use in their code. This allows them to make their modules compatible with my MCP server without forcing them to include the full SDK as a direct dependency in their own projects. This prevents potential dependency conflicts and keeps their modules lightweight, as the core application handles the heavy lifting of the SDK logic.

By removing `final`, you allow the SDK to serve as a flexible foundation for more complex and decoupled integration systems, following the **Open-Closed Principle**.

Thank you for considering this change!

chr-hertel:
Hi @fox-john, interesting thing - do you have a public PR or sth?
I wonder if your `CustomDiscoverer` still relies on some logic of the SDK - or what's the benefit of reusing the built-in attributes if you have your own discoverer in place.

fox-john:
Hi @chr-hertel,

Unfortunately, I don't have any public code to share at the moment. However, I can clarify how our implementation works.

The SDK is integrated into a core PrestaShop module, which allows third-party modules to register their own tools, prompts, and resources.

To avoid forcing every third-party module to include the SDK as a direct dependency, we provide wrapper classes they can use as PHP attributes. Our CustomDiscoverer is then responsible for detecting these specific attributes instead of the default ones. Since we want to maintain full compatibility with your SDK's structure, our classes simply extend your base classes (e.g., Tool, Prompt) while using custom names (e.g., PsMcpTool).

Here is an exemple of structure :

```php
// --- 1. In PsMcpServer module ---
namespace MyAwesomeNamespace;

use Mcp\Capability\Attribute\McpTool;

#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::TARGET_CLASS)]
class PsMcpTool extends McpTool
{
}


// --- 2. In a third-party module ---

namespace MyAwesomeAnotherNamespace;

use MyAwesomeNamespace\PsMcpTool;

class ExportService 
{
    #[PsMcpTool(name: "export_products", description: "Exports products to CSV")]
    public function exportProducts(int $categoryId): string 
    {
        return "Exporting category $categoryId...";
    }
}

chr-hertel:
Understood - thank you @fox-john!

Works for me, wdyt @Nyholm & @CodeWithKyrian?

chr-hertel (approved):

PR #246, opened 2026-02-16
chore(http): create constant for session id header

EdouardCourty:
## Motivation and Context

The `Mcp-Session-Id` value is re-written a few times in the codebase. It should be stored in a constant to avoid duplication.

## How Has This Been Tested?

Tests are passing.

## Breaking Changes

No behavioral or API change.

## Types of changes

- [ ] Documentation update
- [x] Refactor

## Checklist

### DIFF
--- a/src/Server/Transport/StreamableHttpTransport.php
+++ b/src/Server/Transport/StreamableHttpTransport.php
@@ -30,6 +30,17 @@ use Symfony\Component\Uid\Uuid;
  */
 class StreamableHttpTransport extends BaseTransport
 {
+    private const SESSION_HEADER = 'Mcp-Session-Id';
+
+    private const ALLOWED_HEADER = [
+        'Accept',
+        'Authorization',
+        'Content-Type',
+        'Last-Event-ID',
+        'Mcp-Protocol-Version',
+        self::SESSION_HEADER,
+    ];
+
     private ResponseFactoryInterface $responseFactory;
     private StreamFactoryInterface $streamFactory;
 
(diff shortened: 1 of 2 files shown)


src/Server/Transport/StreamableHttpTransport.php
```diff
@@ -30,6 +30,8 @@
  */
 class StreamableHttpTransport extends BaseTransport
 {
+    public const MCP_SESSION_ID_HEADER = 'Mcp-Session-Id';
```
chr-hertel: let's not reuse the const in tests, and only go with private scope - maybe shorter as well?
```suggestion
    private const SESSION_HEADER = 'Mcp-Session-Id';
```

src/Server/Transport/StreamableHttpTransport.php
```diff
@@ -62,8 +64,8 @@ public function __construct(
         $this->corsHeaders = array_merge([
             'Access-Control-Allow-Origin' => '*',
             'Access-Control-Allow-Methods' => 'GET, POST, DELETE, OPTIONS',
-            'Access-Control-Allow-Headers' => 'Content-Type, Mcp-Session-Id, Mcp-Protocol-Version, Last-Event-ID, Authorization, Accept',
-            'Access-Control-Expose-Headers' => 'Mcp-Session-Id',
+            'Access-Control-Allow-Headers' => 'Content-Type, '.self::MCP_SESSION_ID_HEADER.', Mcp-Protocol-Version, Last-Event-ID, Authorization, Accept',
```
chr-hertel: what about introducing a const here as well?
```php
    private const ALLOWED_HEADER = [
        'Accept',
        'Authorization',
        'Content-Type',
        'Last-Event-ID',
        'Mcp-Protocol-Version',
        self::SESSION_HEADER,
    ];
```
in combination with:

```suggestion
            'Access-Control-Allow-Headers' => implode(',', self::ALLOWED_HEADER),
```

EdouardCourty: Fair!

PR #247, opened 2026-02-17
[Server] Encapsulate session handling from Protocol to new SessionManager

chr-hertel:
Extracted from another working branch - goal is to encapsulate/isolate concerns a bit more.

BC Break
* for `Builder::setSession(...)` + dropping ttl => is implicit when injecting store or manager anyways
* that `SessionFactory` to `SessionManager` renaming incl store as dependency
* `SessionInterface` drops `getStore()` method


src/Server/Session/SessionManager.php
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+namespace Mcp\Server\Session;
+
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
+use Symfony\Component\Uid\Uuid;
+
+/**
+ * Default implementation of SessionFactoryInterface.
```
CodeWithKyrian: SessionManagerInterface

src/Server/Session/SessionManagerInterface.php
```diff
@@ -19,17 +19,29 @@
...
+
+    /**
+     * Checks if a session with the given UUID exists.
+     */
+    public function exists(Uuid $id): bool;
+
+    /**
+     * Destroys the session with the given UUID.
+     */
+    public function destroy(Uuid $sessionId): bool;
```
CodeWithKyrian: Slight nitpick, but shoudn't the argument name match the rest in the interface ie `$id` instead of `$sessionId`?

PR #250, opened 2026-02-18
Update SECURITY.md to use GitHub Security Advisories

localden:
Replaces Anthropic HackerOne VDP references with standard GitHub Security Advisory process.


chr-hertel (approved):
Thanks @localden!

PR #252, opened 2026-03-04
fix(session): initialize data lazily in save() to prevent uninitialized property access

soyuka:
I think the code is self-explanatory


soyuka:
If someone can re-run the pipeline it'd be lovely :)

chr-hertel (approved):
Good catch, thanks!

PR #265, opened 2026-03-11
[Server] add a reference handler setter to server builder

simonchrz:
Added a setReferenceHandler method to Mcp\Server\Builder.

## Motivation and Context
The MCP SDK's Builder hardcodes new ReferenceHandler($container) internally — there's no way to customize how tool/resource/prompt handlers are invoked. This prevents consumers from adding cross-cutting concerns like authorization checks before tool execution.

We want to decorate the ReferenceHandler with a SecurityReferenceHandler that reads Symfony's #[IsGranted] attribute from tool methods and checks permissions via the AuthorizationCheckerInterface before delegating to the actual handler. Without setReferenceHandler(), the only alternatives are fragile workarounds (event listener hacks, replacing internal request handlers via reflection, or duplicating authorization logic inside every tool method).

## How Has This Been Tested?
I've added Mcp\Tests\Unit\Server\BuilderTest and tested it manually agains librechat mcp client.

## Breaking Changes
no, it's optional

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 
 * Add built-in authentication middleware for HTTP transport using OAuth
 * Add client component for building MCP clients
+* Add `Builder::setReferenceHandler()` to allow custom `ReferenceHandlerInterface` implementations (e.g. authorization decorators)
 
 0.4.0
 -----


tests/Unit/Server/BuilderTest.php
```diff
@@ -0,0 +1,107 @@
...
+        $this->assertSame('intercepted', $result);
+    }
+
+    private function callTool(Server $server, string $toolName): mixed
+    {
+        $protocol = (new \ReflectionClass($server))->getProperty('protocol')->getValue($server);
+        $requestHandlers = (new \ReflectionClass($protocol))->getProperty('requestHandlers')->getValue($protocol);
+
+        foreach ($requestHandlers as $handler) {
+            if ($handler instanceof \Mcp\Server\Handler\Request\CallToolHandler) {
```
chr-hertel: please use imports for the FQCNs instead of fully inline

```suggestion
            if ($handler instanceof CallToolHandler) {
```

chr-hertel: and below

simonchrz: fixed

Issue #263, opened 2026-03-11
[Server] Make Symfony Finder component optional


chr-hertel:
Since the file based discovery is only an optional feature, we should make the dependency to `symfony/finder` optional as well, and notify the user with a hard exception when using the discovery without symfony/finder installed.

PR #267, opened 2026-03-15
[Server] Fix to make Symfony Finder component optional

bigdevlarry:
## Motivation and Context
The changes make the Symfony Finder component optional, as referenced in issue #263

## How Has This Been Tested?

Yes

## Breaking Changes

Not a breaking change

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 * Add client component for building MCP clients
 * Add `Builder::setReferenceHandler()` to allow custom `ReferenceHandlerInterface` implementations (e.g. authorization decorators)
 * Add elicitation enum schema types per SEP-1330: `TitledEnumSchemaDefinition`, `MultiSelectEnumSchemaDefinition`, `TitledMultiSelectEnumSchemaDefinition`
+* [BC break] Make Symfony Finder component optional. Users would need to install `symfony/finder` now themselves
 
 0.4.0
 -----
--- a/src/Capability/Discovery/Discoverer.php
+++ b/src/Capability/Discovery/Discoverer.php
@@ -24,6 +24,7 @@ use Mcp\Capability\Registry\ResourceReference;
 use Mcp\Capability\Registry\ResourceTemplateReference;
 use Mcp\Capability\Registry\ToolReference;
 use Mcp\Exception\ExceptionInterface;
+use Mcp\Exception\RuntimeException;
 use Mcp\Schema\Prompt;
 use Mcp\Schema\PromptArgument;
 use Mcp\Schema\Resource;
@@ -53,6 +54,10 @@ final class Discoverer implements DiscovererInterface
         private ?DocBlockParser $docBlockParser = null,
         private ?SchemaGeneratorInterface $schemaGenerator = null,
     ) {
+        if (!class_exists(Finder::class)) {
+            throw new RuntimeException('File-based discovery requires symfony/finder. Run: composer require symfony/finder');
+        }
+
         $this->docBlockParser = $docBlockParser ?? new DocBlockParser(logger: $this->logger);
         $this->schemaGenerator = $schemaGenerator ?? new SchemaGenerator($this->docBlockParser);
     }
--- a/composer.json
+++ b/composer.json
@@ -32,8 +32,10 @@
     "psr/http-server-handler": "^1.0",
     "psr/http-server-middleware": "^1.0",
     "psr/log": "^1.0 || ^2.0 || ^3.0",
-    "symfony/finder": "^5.4 || ^6.4 || ^7.3 || ^8.0",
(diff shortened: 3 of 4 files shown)


src/Capability/Discovery/Discoverer.php
```diff
@@ -66,6 +66,10 @@ public function __construct(
      */
     public function discover(string $basePath, array $directories, array $excludeDirs = []): DiscoveryState
     {
+        if (!class_exists(Finder::class)) {
+            throw new \RuntimeException('File-based discovery requires symfony/finder. Run: composer require symfony/finder');
+        }
+
```
soyuka: should the Discoverer be actually loaded in the registry if the finder is not available?

chr-hertel: good one, not sure. i'm a bit torn - it feels like we could directly check this in the `Mcp\Server\Builder` (in `setDiscovery(...)`, `createDiscoverer(...)` or `build()`, but on the other hand it makes sense to have it close to the place where the requirement is coming from originally, which is the `Discoverer` implementation - and I think I would avoid having it twice

maybe in constructor here? trace would be `Builder::build()` => `Builder::createDiscoverer()` => `Discoverer::__construct(...)`

src/Capability/Discovery/Discoverer.php
```diff
@@ -66,6 +66,10 @@ public function __construct(
      */
     public function discover(string $basePath, array $directories, array $excludeDirs = []): DiscoveryState
     {
+        if (!class_exists(Finder::class)) {
+            throw new \RuntimeException('File-based discovery requires symfony/finder. Run: composer require symfony/finder');
```
chr-hertel: no matter where we throw it, it should be `Mcp\Exception\RuntimeException` instead of global one

src/Capability/Discovery/Discoverer.php
soyuka: I mean we wouldn't instantiate the Discoverer if no Finder is available, gracefully logging that it's not present. Then if someone manually tries to construct one we may want to add this to the discoverer arguments so that it can't be?

soyuka:
Can you also add a change to https://github.com/modelcontextprotocol/php-sdk/blob/main/src/Server/Builder.php#L526 ? Avoid calling the `createDiscoverer` when Finder isn't accessible + log? This makes php-sdk not dependent from the Finder.

bigdevlarry:
@soyuka, this means the discoverer will never be called if there's no finder, and we won't be able to raise the exception to prompt the installation of the finder itself. By your suggestion, adding a null check silences the error in the logs, and the program continues as expected. Is this the ideal behaviour, or should we halt the program with the exception as we have it, so users go back to install? Let me know what you think as well @chr-hertel

soyuka:
Mhh from this PR I thought you wanted to move the `Finder` to a dev-requirement making this a soft dependency. From my point of view if that's the case the code should work after installing with composer and I'd have thought that it'd make sense to skip classes dependent from the Finder. 

If we don't want to move this requirement, then I'm not sure I understand how this code path is going to happen as its required in our dependencies.

chr-hertel:
> Mhh from this PR I thought you wanted to move the Finder to a dev-requirement making this a soft dependency.
yes

and yes, the code should work unless you want to use the built-in discoverer.

if a user calls `setDiscovery` we can expect the intend is clear - calling `setDiscovery`, not calling `setDiscoverer` with an custom implementation, but not installing the `symfony/finder` should raise a warning or exception from my point of view.

We should add this as BC break to changelog tho, since users would need to install `symfony/finder` now themselves.

chr-hertel:
@soyuka this is now what you've been thinking, right?

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to `mcp/sdk` will be documented in this file.
 
+0.6.0
+-----
+
+* [BC break] Make Symfony Finder component optional. Users would need to install `symfony/finder` now themselves
```
chr-hertel: needs to be 0.5.0 please

chr-hertel (changes requested):

soyuka (approved):

luoyue712:
Hello @bigdevlarry @chr-hertel  , I think we should remove DiscoveryLoader from the builder, since we can register DiscoveryLoader using addLoader.
 
Benefits:
 
1. This makes the code cleaner
2. No additional checks are needed to determine whether the Symfony Finder component is installed
3. Makes the discovery loader more flexible and allows the implementation class to be replaced freely

chr-hertel (approved):

chr-hertel:
Thank you @bigdevlarry!

Btw, @luoyue712 let's shift that discussion to a follow-up issue - but, tbh, i think for standalone usage the DiscoveryLoader is quite convenient to have around.

PR #260, opened 2026-03-10
[Server] Add DNS rebinding protection feature with middleware

chr-hertel:
Adding HTTP middleware for DNS middleware protection - following conformance requirements.

- [x] need to review myself first

PR #277, opened 2026-04-11
[Server] Extract CORS handling into CorsMiddleware

### LINKED ISSUE #280: Hardcoded Wildcard CORS (Access-Control-Allow-Origin: *)
**Describe the bug**
Hardcoded Wildcard CORS (Access-Control-Allow-Origin: * ) is the default CORS headers to every response via withCorsHeaders().

**Details**
Here is the code -
https://github.com/modelcontextprotocol/php-sdk/blob/main/src/Server/Transport/StreamableHttpTransport.php#L74

The CORS override can happen at https://github.com/modelcontextprotocol/php-sdk/blob/main/src/Server/Transport/StreamableHttpTransport.php#L64, but I believe the SDK should follow the secure by default principle and allow the deployers to explicitly set insecure values if they need.

**Recommendation**
In the SDK, the transport layer should not own CORS policy. Server implementors who need cross-origin access can add a CORS filter at the servlet filter or Spring Security layer.

chr-hertel:
## Summary

- Extracts inline CORS handling from `StreamableHttpTransport` into a dedicated `CorsMiddleware` (PSR-15)
- `CorsMiddleware` is automatically prepended to the middleware chain, handling OPTIONS preflight and applying CORS headers to all responses
- Changes default `Access-Control-Allow-Origin` from `*` (allow all) to not set (block cross-origin), with explicit opt-in via `allowedOrigins`
  - addressing a security design flaw reported by @srikanthramu
  - Fix #280 

## Breaking Changes

- The `$corsHeaders` constructor parameter on `StreamableHttpTransport` is replaced by `?CorsMiddleware $corsMiddleware`
- Default CORS behavior no longer sets `Access-Control-Allow-Origin: *` — use `new CorsMiddleware(allowedOrigins: ['*'])` to restore

### DIFF
--- a/src/Server/Transport/StreamableHttpTransport.php
+++ b/src/Server/Transport/StreamableHttpTransport.php
@@ -14,6 +14,7 @@ namespace Mcp\Server\Transport;
 use Http\Discovery\Psr17FactoryDiscovery;
 use Mcp\Exception\InvalidArgumentException;
 use Mcp\Schema\JsonRpc\Error;
+use Mcp\Server\Transport\Http\Middleware\CorsMiddleware;
 use Mcp\Server\Transport\Http\MiddlewareRequestHandler;
 use Psr\Http\Message\ResponseFactoryInterface;
 use Psr\Http\Message\ResponseInterface;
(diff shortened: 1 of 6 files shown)


CodeWithKyrian:
This is a solid refactor...makes a lot of sense.

One thing I was wondering though: instead of introducing a dedicated $corsMiddleware parameter on the transport, would it be feasible to just work with the existing $middleware array?

For example:
- accept CorsMiddleware as part of the middleware list
- detect if it’s already present
- ensure it’s in the correct position (e.g. prepend or reorder if needed)
- optionally inject a default if none is provided

That way everything stays within a single middleware pipeline instead of splitting configuration paths.

I get that this might complicate the setup a bit, so maybe that’s why you went this route...but I was just curious if you explored that approach at all?

chr-hertel:
@CodeWithKyrian Thanks, yeah, true - so similar to the approach in #260, right?

CodeWithKyrian:
Yes exactly!

src/Server/Transport/StreamableHttpTransport.php
```diff
@@ -70,19 +57,21 @@ public function __construct(
...
             }
+            if ($m instanceof CorsMiddleware) {
+                $hasCorsMiddleware = true;
+            }
             $this->middleware[] = $m;
         }
+
+        if (!$hasCorsMiddleware) {
+            array_unshift($this->middleware, new CorsMiddleware());
+        }
```
soyuka, line 74: let's say I use the mcp bundle with the famous https://github.com/nelmio/NelmioCorsBundle how am I going to avoid conflicts?

chr-hertel: Haven't tested, but we'd need a way to opt-out completely you mean?

soyuka: yes can we just remove these checks? we add it by default in our server in here but lets not add it if not found as it prevents other libraries to do their own cors handling.

chr-hertel:
Closed in favor of #307

PR #281, opened 2026-04-11
[Schema][Server] Add MCP Apps extension (io.modelcontextprotocol/ui) support

chr-hertel:
## Summary

- Add typed schema classes under `Mcp\Schema\Extension\Apps\` for the MCP Apps extension (`io.modelcontextprotocol/ui`, spec `2026-01-26`)
- Add `extensions` field to both `ServerCapabilities` and `ClientCapabilities` for extension negotiation
- Add `ToolVisibility` enum (`model`/`app`)
- Add weather dashboard example in `examples/server/mcp-apps/` demonstrating UI resources and tool-UI linkage
- Add 34 new tests covering all schema classes, serialization, and capabilities backward compatibility

## Details

The MCP Apps extension enables servers to expose interactive HTML UI applications as MCP resources. This PR adds server-side support:

- **Schema classes**: `McpApps` (constants/helpers), `UiResourceCsp`, `UiResourcePermissions`, `UiResourceContentMeta`, `UiToolMeta` — all with `fromArray()`/`jsonSerialize()` and `toMetaArray()` helpers for embedding in `_meta`
- **Capabilities**: `extensions` parameter on both capability classes (backward-compatible, defaults to `null`)
- **No new JSON-RPC methods** — the extension uses existing `resources/list`, `resources/read`, `tools/list`, `tools/call`
- **Out of scope**: View-Host protocol (`ui/initialize`, `ui/open-link`, etc.) which runs between the rendered HTML and the host application

## Test plan

- [x] `make cs` — coding standards pass
- [x] `make phpstan` — static analysis passes (0 errors)
- [x] `make unit-tests` — all 754 tests pass (34 new)
- [x] Manual: `npx @modelcontextprotocol/inspector php examples/server/mcp-apps/server.php`

## Running in Goose

[Screencast from 2026-05-11 23-45-55.webm](https://github.com/user-attachments/assets/c5d50d64-9e57-4181-800e-6af807bdbc62)


chr-hertel:
This could use some love as follow up when it comes to the Builder. maybe a `addApp()` method would be nice, but didn't want to bloat the PR even more.

src/Schema/Extension/Apps/UiResourceContentMeta.php
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public static function fromArray(array $data): self
+    {
+        return new self(
+            csp: isset($data['csp']) ? UiResourceCsp::fromArray($data['csp']) : null,
+            permissions: isset($data['permissions']) ? UiResourcePermissions::fromArray($data['permissions']) : null,
+            domain: $data['domain'] ?? null,
+            prefersBorder: $data['prefersBorder'] ?? null,
+        );
+    }
```
soyuka, line 52: 🟡 **risk** — `jsonSerialize()` return type annotation `UiResourceContentMetaData` claims `csp?: UiResourceCspData` (array shape) but the property stores a `UiResourceCsp` object. PHPStan-level lie; serialization still works via recursive `JsonSerializable`.

**Fix:** annotate as `array{csp?: UiResourceCsp, permissions?: UiResourcePermissions, domain?: string, prefersBorder?: bool}`, or call `->jsonSerialize()` explicitly on each.

src/Schema/Extension/Apps/UiResourcePermissions.php
```diff
@@ -0,0 +1,74 @@
...
+ * @phpstan-type UiResourcePermissionsData array{
+ *     camera?: \stdClass|array<string, mixed>,
+ *     microphone?: \stdClass|array<string, mixed>,
+ *     geolocation?: \stdClass|array<string, mixed>,
+ *     clipboardWrite?: \stdClass|array<string, mixed>
+ * }
+ *
+ * @author Christopher Hertel <mail@christopher-hertel.de>
+ */
+final class UiResourcePermissions implements \JsonSerializable
```
soyuka, line 29: 🟡 **risk** — Permission set hardcoded to 4 entries (`camera`/`microphone`/`geolocation`/`clipboardWrite`). If the spec adds new permissions, `fromArray()` silently drops them and `jsonSerialize()` cannot emit them.

**Fix:** keep unknown keys as passthrough (`array<string, \stdClass>`) or document the closed set explicitly.

src/Schema/Extension/Apps/UiResourcePermissions.php
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public static function fromArray(array $data): self
+    {
+        return new self(
+            camera: \array_key_exists('camera', $data),
+            microphone: \array_key_exists('microphone', $data),
+            geolocation: \array_key_exists('geolocation', $data),
+            clipboardWrite: \array_key_exists('clipboardWrite', $data),
+        );
+    }
```
soyuka, line 52: 🔵 **nit** — `\array_key_exists('camera', $data)` treats `'camera' => null` as requested. Spec payload sends `'camera' => {}`.

**Fix:** tighten to `isset()`, or accept either form and document the chosen contract.

src/Schema/Extension/Apps/UiToolMeta.php
```diff
@@ -0,0 +1,63 @@
...
+ */
+final class UiToolMeta implements \JsonSerializable
+{
+    /**
+     * @param ?string   $resourceUri the ui:// URI of the linked UI resource
+     * @param ?string[] $visibility  who can see/call this tool: 'model', 'app', or both (default: ['model', 'app'])
+     */
+    public function __construct(
+        public readonly ?string $resourceUri = null,
+        public readonly ?array $visibility = null,
```
soyuka, line 33: 🟡 **risk** — `?string[] $visibility` forces callers to pass `[ToolVisibility::Model->value, ...]`. Verbose, typo-prone, defeats the existing `ToolVisibility` enum.

**Fix:** accept `array<ToolVisibility>|array<string>` and normalize in the constructor — catches typos at construction time.

src/Schema/Extension/Apps/McpApps.php
```diff
@@ -0,0 +1,46 @@
...
+ * Enable on the server via {@see \Mcp\Server\Builder::enableExtension()}.
+ *
+ * @see https://github.com/modelcontextprotocol/ext-apps
+ *
+ * @author Christopher Hertel <mail@christopher-hertel.de>
+ */
+final class McpApps implements ServerExtensionInterface
+{
+    public const EXTENSION_ID = 'io.modelcontextprotocol/ui';
+    public const MIME_TYPE = 'text/html;profile=mcp-app';
```
soyuka, line 31: ❓ **question** — MIME constant `text/html;profile=mcp-app`. RFC 6838 typically quotes the `profile` parameter (`profile="mcp-app"`). Is the unquoted form spec-mandated by `io.modelcontextprotocol/ui`, or just copied loosely?

src/Server/Builder.php
```diff
@@ -230,6 +236,31 @@ public function setCapabilities(ServerCapabilities $serverCapabilities): self
...
+    {
+        foreach ($extensions as $extension) {
+            if (\is_string($extension)) {
+                if (!is_subclass_of($extension, ServerExtensionInterface::class)) {
+                    throw new InvalidArgumentException(\sprintf('Extension class "%s" must implement "%s".', $extension, ServerExtensionInterface::class));
+                }
+                $extension = new $extension();
+            }
+
+            $this->extensions[$extension->getId()] = $extension->getCapabilities();
```
soyuka: 🟡 **risk** — Silent overwrite of an extension by id (`$this->extensions[$extension->getId()] = ...`). Two `enableExtension(McpApps::class, ...)` calls with the same id silently let the last win — no warning, no surfaceable error.

**Fix:** log a warning, or throw `LogicException` on duplicate id.

src/Server/Builder.php
```diff
@@ -587,6 +618,7 @@ public function build(): Server
             promptsListChanged: $this->eventDispatcher instanceof EventDispatcherInterface,
             logging: true,
             completions: true,
+            extensions: [] !== $this->extensions ? $this->extensions : null,
```
soyuka: 🟡 **risk** — When a caller uses `setCapabilities()`, the `$this->serverCapabilities ?? new ServerCapabilities(...)` branch never folds `$this->extensions` in. Extensions registered via `enableExtension()` are silently dropped. Documented in `docs/extensions.md` but no runtime hint.

**Fix:** `$logger->warning(...)` when both are set, or merge `$this->extensions` into the caller-provided capabilities.

src/Server/Builder.php
```diff
@@ -230,6 +236,31 @@ public function setCapabilities(ServerCapabilities $serverCapabilities): self
...
+     * @param class-string<ServerExtensionInterface>|ServerExtensionInterface ...$extensions
+     */
+    public function enableExtension(string|ServerExtensionInterface ...$extensions): self
+    {
+        foreach ($extensions as $extension) {
+            if (\is_string($extension)) {
+                if (!is_subclass_of($extension, ServerExtensionInterface::class)) {
+                    throw new InvalidArgumentException(\sprintf('Extension class "%s" must implement "%s".', $extension, ServerExtensionInterface::class));
+                }
+                $extension = new $extension();
```
soyuka: 🔵 **nit** — `new $extension()` assumes a zero-arg constructor. The PHPDoc says "instantiated with no arguments" but `ServerExtensionInterface` does not enforce it — runtime error if an extension class requires constructor args.

**Fix:** add `__construct()` to `ServerExtensionInterface`, or restrict the `class-string` variant to a dedicated marker interface that documents the contract.

src/Schema/Extension/ServerExtensionInterface.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * A server-side MCP protocol extension advertised during capability negotiation.
+ *
+ * Implementations are typically zero-config — they expose a stable identifier and the
+ * capability payload announced under `capabilities.extensions[<id>]` in the initialize
+ * response.
+ *
+ * @author Christopher Hertel <mail@christopher-hertel.de>
+ */
+interface ServerExtensionInterface
```
soyuka, line 23: 🔵 **nit** — `getCapabilities(): array<string, mixed>` is too loose. Most concrete extensions will have structured payloads (e.g. `mimeTypes`, `uriSchemes`).

**Fix:** keep the generic interface but expose a typed sub-interface for well-known extensions, or at least document the JSON-serializability contract.

src/Schema/ServerCapabilities.php
```diff
@@ -107,6 +110,7 @@ public static function fromArray(array $data): self
             logging: $loggingEnabled,
             completions: $completionsEnabled,
             experimental: $data['experimental'] ?? null,
+            extensions: $data['extensions'] ?? null,
```
soyuka: 🟡 **risk** — `fromArray` accepts `extensions` as raw `$data['extensions'] ?? null` — no shape validation. A malformed payload (e.g. a string or scalar) silently propagates into the readonly property and only blows up later during serialization.

**Fix:** add an `is_array($data['extensions'])` guard, mirroring how `experimental` and friends are handled.

src/Schema/ClientCapabilities.php
```diff
@@ -68,7 +71,8 @@ public static function fromArray(array $data): self
             $rootsListChanged,
             $sampling,
             $elicitation,
-            $data['experimental'] ?? null
+            $data['experimental'] ?? null,
+            $data['extensions'] ?? null,
```
soyuka: 🟡 **risk** — Same shape-validation gap as `ServerCapabilities::fromArray`: `extensions` is taken verbatim from `$data` with no `is_array` guard.

**Fix:** validate before assignment.

examples/server/mcp-apps/server.php
```diff
@@ -0,0 +1,51 @@
...
+    ->setServerInfo('MCP Apps Weather Example', '1.0.0')
+    ->setLogger(logger())
+    ->enableExtension(McpApps::class)
+    ->addResource(
+        [WeatherApp::class, 'getWeatherApp'],
+        'ui://weather-app',
+        'weather-app',
+        description: 'Interactive weather dashboard',
+        mimeType: McpApps::MIME_TYPE,
+        meta: ['ui' => new stdClass()],
```
soyuka: 🔵 **nit** — `meta: ['ui' => new stdClass()]` is a magic presence marker. Future readers will not know why an empty `stdClass` is required vs. just omitting the key.

**Fix:** introduce a typed marker (e.g. `McpApps::resourceMarker(): \stdClass`) so intent is discoverable and the empty-object protocol detail is hidden behind a named API.

examples/server/mcp-apps/weather-app.html
```diff
@@ -0,0 +1,230 @@
...
+            rainy:           { icon: '🌧️', key: 'rainy' },
+        };
+
+        // MCP Apps protocol: postMessage with JSONRPCMessage objects between iframe (View) and host.
+        // Spec: https://github.com/modelcontextprotocol/ext-apps
+        let requestId = 0;
+        let initialized = false;
+        const pending = new Map();
+
+        function post(msg) { window.parent.postMessage(msg, '*'); }
```
soyuka, line 134: 🔵 **nit** — `post(msg)` uses `'*'` as `postMessage` target origin. That's a spec-level limitation (the iframe doesn't know its parent origin upfront), but worth an inline comment here — the receiving host **must** validate sender origin on its side.

examples/server/mcp-apps/weather-app.html
```diff
@@ -0,0 +1,230 @@
...
+        // tool-input / tool-result notifications or tools/call requests.
+        (async () => {
+            try {
+                await sendRpc('ui/initialize', {
+                    protocolVersion: '2026-01-26',
+                    capabilities: {},
+                    clientInfo: { name: 'weather-app', version: '1.0.0' },
+                });
+            } catch { /* degrade gracefully on non-conforming hosts */ }
+            initialized = true;
```
soyuka: 🔵 **nit** — `initialized = true` is set but never read anywhere in the file. Dead variable.

docs/extensions.md
```diff
@@ -0,0 +1,103 @@
...
+            text: file_get_contents(__DIR__.'/app.html'),
+            meta: ['ui' => new UiResourceContentMeta(
+                csp: new UiResourceCsp(connectDomains: ['https://api.example.com']),
+                permissions: new UiResourcePermissions(geolocation: true),
+                prefersBorder: true,
+            )],
+        ),
+        'ui://my-app',
+        mimeType: McpApps::MIME_TYPE,
+        meta: ['ui' => new \stdClass()],
```
soyuka: 🔵 **nit** — The example uses `meta: ['ui' => new \stdClass()]` on the resource and `meta: ['ui' => new UiResourceContentMeta(...)]` inside `TextResourceContents` a few lines up. Two different `_meta.ui` shapes (presence marker vs. structured metadata) — readers will conflate them.

**Fix:** add one sentence clarifying when each shape applies.

src/Schema/Extension/Apps/UiResourceCsp.php
```diff
@@ -0,0 +1,80 @@
...
+        if (null !== $this->resourceDomains) {
+            $data['resourceDomains'] = $this->resourceDomains;
+        }
+        if (null !== $this->frameDomains) {
+            $data['frameDomains'] = $this->frameDomains;
+        }
+        if (null !== $this->baseUriDomains) {
+            $data['baseUriDomains'] = $this->baseUriDomains;
+        }
+
```
soyuka: 🔵 **nit** — Empty array `[]` for any of the `*Domains` fields serializes to `"connectDomains": []` etc. Spec question: is an empty list "allow none" or "omit field"? If "omit", change the guards to `if ($this->connectDomains)` (truthy check) instead of `null !== ...`.

src/Schema/Extension/Apps/McpApps.php
chr-hertel: It's how the spec documented it:
https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx

src/Schema/Extension/Apps/UiResourcePermissions.php
chr-hertel: this refers to the [actual spec](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx) and is mapped to concrete properties - if we go down the road of opening the spec for dynamics that would be quite a stunt and i'd rather argue that these DTOs should be exactly like the spec.

src/Server/Builder.php
chr-hertel: just dropped support for handing over class-strings

src/Schema/Extension/ServerExtensionInterface.php
chr-hertel: added a comment, but would guess that's enough?

soyuka (approved):

PR #282, opened 2026-04-11
CS update after release of v3.95

chr-hertel:
Due to https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/releases/tag/v3.95.0


Nyholm (approved):
Thank you

PR #288, opened 2026-04-19
[Schema][Server] Add top-level title to Tool

### LINKED ISSUE #287: [Server] Missing top-level `title` field in `Tool` schema
I noticed that the `Tool` schema does not expose the top-level `title` field introduced by the MCP specification revision `2025-06-18`. It seems that this was implemented for prompts in https://github.com/modelcontextprotocol/php-sdk/issues/276

e0ipso:
## Summary

- Adds the optional `title` field to `Tool` (mirrors the `title` addition to `Prompt` in #278) for MCP spec compliance.
- Threads the new `title` through the `McpTool` attribute, `Discoverer`, `ArrayLoader`, and `Server\Builder::addTool()`.
- `title` is serialized as the second key (between `name` and `inputSchema`) when non-null and omitted otherwise.
- Non-BC-breaking: `title` is a trailing optional parameter (default `null`) on `Tool::__construct`, `McpTool`, and `addTool()`; positional callers are unaffected.
- Fix #287 

## Test plan

- [x] `vendor/bin/phpunit` pass.
- [x] New unit coverage: `ToolTest` (construction, serialization key order, `fromArray`, round-trip) via data providers.
- [x] `DiscovererToolTitleTest` exercises `#[McpTool(title: ...)]` end-to-end against the shared `DiscoverableToolHandler` fixture.
- [x] `ArrayLoaderToolTitleTest` verifies the loader forwards `title` into the registered `Tool`.
- [x] `McpToolTest` covers the new attribute field and its default.


e0ipso:
The, unrelated, CI failures are fixed in https://github.com/modelcontextprotocol/php-sdk/pull/289

src/Capability/Attribute/McpTool.php
```diff
@@ -35,6 +36,7 @@ public function __construct(
         public ?array $icons = null,
         public ?array $meta = null,
         public ?array $outputSchema = null,
+        public ?string $title = null,
```
chr-hertel: I added this for prompt in #278 between `$name` and `$description` because i think of it as quite an essential property and we're okay with BC breaks in 0.x versions - would be more consistent from my point of view to place it in the same order here.

src/Schema/Tool.php
```diff
@@ -64,6 +66,7 @@ public function __construct(
         public readonly ?array $icons = null,
         public readonly ?array $meta = null,
         public readonly ?array $outputSchema = null,
+        public readonly ?string $title = null,
```
chr-hertel: same here, after `$name` would feel more consistent to me.

src/Server/Builder.php
```diff
@@ -387,6 +389,7 @@ public function addTool(
         ?array $icons = null,
         ?array $meta = null,
         ?array $outputSchema = null,
+        ?string $title = null,
```
chr-hertel: same here i guess :)

chr-hertel (changes requested):
Thanks for tackling this @e0ipso - only small consistency issue and please add an entry to the changelog for 0.5 - thanks already! :pray:

Rebase will solve the CS issue - thanks to you :D

chr-hertel (approved):

PR #289, opened 2026-04-19
style: apply php-cs-fixer and PHPStan fixes

e0ipso:
## Summary

- Applies `vendor/bin/php-cs-fixer fix` with the default project config to 13 pre-existing test files.
- The only change is adding missing `: void` return types to test methods (and a couple of spacing tweaks).

## Test plan

- [x] `vendor/bin/php-cs-fixer fix --dry-run` — clean after the commit.


e0ipso:
I am unsure about the policy around this stuff. Feel free to close it if you feel this is breaking expectations (even if we are in 0.4.0).

chr-hertel (approved):
Thank you @e0ipso - will merge this first and rebase #282!

PR #292, opened 2026-04-22
[Server] Allow overriding the default file name pattern for Discovery

johnhunt-lc:
This PR adds the option to pass in a `namePatterns` array of patterns to match against filenames, overriding the default `'*.php'` pattern that was previously hardcoded.

## Motivation and Context

It would be useful to be able to either limit further, or expand, the scope of files matched by discovery. The specific cases I've run into that require this functionality are:

* excluding some tools depending on environment context
* matching code convention for legacy projects that use alternative file extensions for php class files (`*.inc` for example)

## How Has This Been Tested?

I have tested this in a php application that uses `.inc` files to store classes, and have added additional unit tests. I haven't tried it in other contexts.

## Breaking Changes
This should not affect existing code.

## Types of changes
- [x] Documentation update

## Checklist


src/Capability/Discovery/Discoverer.php
```diff
@@ -65,11 +65,12 @@ public function __construct(
...
-     * @param string        $basePath    the base path for resolving directories
-     * @param array<string> $directories list of directories (relative to base path) to scan
-     * @param array<string> $excludeDirs list of directories (relative to base path) to exclude from the scan
+     * @param string             $basePath     the base path for resolving directories
+     * @param array<string>      $directories  list of directories (relative to base path) to scan
+     * @param array<string>      $excludeDirs  list of directories (relative to base path) to exclude from the scan
+     * @param array<string>|null $namePatterns list of file name patterns for the scan. Compatible with Finder->name()
      */
-    public function discover(string $basePath, array $directories, array $excludeDirs = []): DiscoveryState
+    public function discover(string $basePath, array $directories, array $excludeDirs = [], ?array $namePatterns = null): DiscoveryState
```
chr-hertel: Using a default here would ease the change:

```suggestion
    public function discover(string $basePath, array $directories, array $excludeDirs = [], array $namePatterns = ['*.php']): DiscoveryState
```

docs/server-builder.md
```diff
@@ -99,8 +99,9 @@ $server = Server::builder()
     ->setDiscovery(
         basePath: __DIR__,
         scanDirs: ['.', 'src', 'lib'],           // Where to look for MCP attributes
-        excludeDirs: ['vendor', 'tests'],        // Where NOT to look 
-        cache: $cacheInstance                    // Optional: cache discovered elements
+        excludeDirs: ['vendor', 'tests'],        // Where NOT to look
+        cache: $cacheInstance,                   // Optional: cache discovered elements
+        namePatterns: ['*.php', '*.inc'],         // Optional: list of filename patterns to match
```
chr-hertel: looks like a small whitespace inconsistency

```suggestion
        excludeDirs: ['vendor', 'tests'],        // Where NOT to look
        cache: $cacheInstance,                   // Optional: cache discovered elements
        namePatterns: ['*.php', '*.inc'],        // Optional: list of filename patterns to match
```

chr-hertel (changes requested):
Hey @johnhunt-lc, wasn't aware of that use case, but recall seeing that convention years ago. Thanks for the patch, minor comments tho.

src/Capability/Discovery/Discoverer.php
johnhunt-lc: I was batting around different ideas for this. $namePatterns gets passed through quite a few different functions, and I originally specified the `['*.php']` default for each one, but I realized that was scattering hardcoded defaults everywhere so I switched to default `null` in all the functions that pass it along, and just set the default in the ternary below.

I agree that I prefer having the default in the params instead. Would you rather I:

* set a class const and use it for the default in all functions that use this argument?
* hardcode this in all of them?
* set it here, make the rest default to null, make this nullable, and handle nulls in the function logic?
* some other 4th thing I haven't thought of?

johnhunt-lc: @chr-hertel hi! I ended up adding a const for the default value on the Interface, and then using that const as the default value for that param on functions that use it. Does this work better or would you prefer a different method?

docs/server-builder.md
```diff
@@ -109,6 +110,7 @@ $server = Server::builder()
 - `$scanDirs` (array): Directories to recursively scan for `#[McpTool]`, `#[McpResource]`, etc. All subdirectories are included. (default: `['.', 'src']`)
 - `$excludeDirs` (array): Directory names to exclude **within** the scanned directories during recursive scanning
 - `$cache` (CacheInterface|null): Optional PSR-16 cache to store discovered elements for performance
+- `$namePatterns` (array): Optional list of Finder->name() compatible patterns to match against file names (default: `['*.php']`)
```
chr-hertel: Exposing `Finder->name()` here feels a bit internal to me, let's focus on the behavior:
```suggestion
- `$namePatterns` (array): Optional list of patterns (regexp, glob, or string) for file names (default: `['*.php']`)
```

src/Capability/Discovery/Discoverer.php
chr-hertel: Ah okay, thanks for that explanation - sure you need to duplicate that now. works for me like this :+1:

johnhunt-lc: I think I got this everywhere, are we good to resolve this thread?

chr-hertel (dismissed):
Thank you @johnhunt-lc!

chr-hertel (approved):

PR #294, opened 2026-04-29
[Server] Allow runtime-resolved element handlers

e0ipso:
## Summary

Adds `Builder::add(Tool|ResourceDefinition|ResourceTemplate|Prompt $definition, ElementHandlerInterface $handler)` for cases where an element's name, schema, or description is only known at runtime. The existing `addTool/addResource/addResourceTemplate/addPrompt` methods, which derive metadata from reflection, keep working unchanged.

Four handler interfaces back the new entry point:

- `ToolHandlerInterface::execute(array $arguments, ClientGateway $gateway)`
- `ResourceHandlerInterface::read(string $uri, ClientGateway $gateway)`
- `ResourceTemplateHandlerInterface::read(string $uri, array $variables, ClientGateway $gateway)`
- `PromptHandlerInterface::get(array $arguments, ClientGateway $gateway)`

All four extend an empty `ElementHandlerInterface` marker so `Builder::add()` can accept any of them through one signature. Mismatched pairings (for example a `Tool` with a `PromptHandlerInterface`) raise `Mcp\Exception\InvalidArgumentException` at registration.

A new `ExplicitElementLoader` translates these pairs into Registry entries. It is wired in `Builder::build()` alongside `ReflectedElementLoader` (renamed from `ArrayLoader`), so the reflection-based loader is unchanged. `ReferenceHandler::handle()` builds a `ClientGateway` from the active session and injects it into handlers that declare a `ClientGateway` parameter.

## Motivation and Context

Reflection-based discovery does not fit config-driven integrations where parameter names, schemas, and arguments are not known at the time the code is written. Two concrete blockers:

- `ReferenceHandler::prepareArguments()` binds incoming arguments to handler parameters by name via reflection, so the handler signature has to mirror the input schema field names.
- `HandlerResolver::resolve()` only accepts `[ClassName::class, 'method']` strings, which rules out DI-constructed handler instances with constructor arguments.

The new entry point gives those integrations a typed registration path without changing the existing one. Standalone usage with attributes or callables is unaffected.

## How Has This Been Tested?

- `ExplicitElementLoaderTest` covers tool, resource, resource template, and prompt registration through `Builder::add()`, plus type-mismatch failures.
- `ReferenceHandlerTest` covers the explicit-handler dispatch branch in `ReferenceHandler::handle()` for all four element kinds.
- Existing tests still pass.

## Breaking Changes

- Renamed `Mcp\Schema\Resource` to `Mcp\Schema\ResourceDefinition`. No alias.
- Renamed `Mcp\Capability\Registry\Loader\ArrayLoader` to `Mcp\Capability\Registry\Loader\ReflectedElementLoader`.

Existing closure/array/string handler forms keep working through `ReflectedElementLoader`.

## Types of changes

- [ ] Bug fix
- [x] New feature (non-breaking)
- [x] Breaking change
- [x] Documentation update

## Checklist

- [x] My code follows the repository style guidelines


src/Capability/Registry/Loader/ArrayLoader.php
```diff
@@ -98,6 +105,37 @@ public function load(RegistryInterface $registry): void
...
+                        annotations: $data['annotations'] ?? null,
+                        icons: $data['icons'] ?? null,
+                        meta: $data['meta'] ?? null,
+                        outputSchema: $outputSchema,
+                    );
+                    $registry->registerTool($tool, $data['handler'], true);
+
+                    $handlerDesc = $this->getHandlerDescription($data['handler']);
+                    $this->logger->debug("Registered manual runtime tool {$data['name']} from handler {$handlerDesc}");
+                    continue;
```
mglaman: Is there a way we can prevent duplicating the `$tool = new Tool(` below? There should be a logic branch we can do like


```
if ($data['handler'] instanceof RuntimeToolHandlerInterface) {
} else {
  $reflection = HandlerResolver::resolve($data['handler']);
```

Then we can have the required variables for just one `$tool = new Tool(`

mglaman (commented):
I wonder if we can simplify the array loader

mglaman:
Actually I was looking at `\Mcp\Capability\Registry\Loader\ArrayLoader::load`:

It respects everything from `$data`, which allows dynamic `inputSchema` and `outputSchema` if you use `\Mcp\Server\Builder::addTool`. The reflection is still made, but can be ignored. Do we need a new interface for this? Because the runtime information can be generated when the server is built

src/Capability/Registry/Loader/ArrayLoader.php
e0ipso: This was my way of trying to keep the diff simple. But you are right, that was not the right approach. I will introduce separate helper methods to prepare each provider so we can have a clean:

```php
$tool = $data['handler'] instanceof RuntimeHandlerInterface
  : $this->prepareRuntimeHandler($data)
  ? $this->prepareReflectionHandler($data)
```

e0ipso:
Thanks for the review @mglaman! I have updated the code based on your suggestions. I like it much better now!

---

Here is my answer to your question above on "what is the point of this PR?". I should have addressed that more clearly in the PR description.

Dispatch binds by parameter name, not by schema. `ReferenceHandler::prepareArguments()` iterates the handler's `ReflectionParameters` and looks up `$arguments[$paramName]`. Variadic is explicitly TODO. So if your tool's inputSchema is `{title: string, body: string}`, the handler method has to be `fn(string $title, string $body)`. With a config-driven schema you don't know those parameter names at the time of writing the code.

Additionally `HandlerResolver::resolve()` requires both elements to be strings, so `[$instance, 'execute']` is rejected. Only `[ClassName::class, 'method']` works. For DI-constructed objects with constructor args, that doesn't work. That could be addressed differently, but this PR removes that need.

Finally, prompts/templates have their `PromptArguments` built from reflection only. There is no field in the manual-prompt array for arguments or completion providers.

I am not saying the interface isn't the only way, but it is typically the way to write a contract for code integration.

mglaman:
>  So if your tool's inputSchema is {title: string, body: string}, the handler method has to be fn(string $title, string $body). With a config-driven schema you don't know those parameter names at the time of writing the code.

Perfect, that's what I was missing. Handler signature must match the schema. But in my opinion that would always be the case. I guess it's just hard for me to visualize the contract of something using this. The tests and docs don't cover `array $arguments`. 

I still don't understand what a "from configuration" tool would look like. I know Drupal does fancy stuff but I am failing to visualize an example of the code to help me grok this change. I am still on the "I'd have a plugin that maps Drupalisms to the SDK"

chr-hertel:
Hi @e0ipso, interesting approach and quite valid use case I understand.
I'm not sure I get what's the benefit of reusing the `ArrayLoader`?

With `api-platform/mcp` @soyuka had quite a similar problem - that's why we have that loader extension point IIRC. however, he needed to go with an extra [handler](https://github.com/api-platform/mcp/blob/main/Server/Handler.php) as well and I feel like it is desirable to solve that use case on SDK level already :thinking: 

Maybe we can even think of inverting the default - for standalone usage the discovery feature is great, but in most other cases i saw, it's basically worked around - those `RuntimeXyzInterface` implementation feel similarly to me, WDYT?

---

Edit, second thought:
It could be convenient to have a `Builder::add(new Prompt(...), new MyPromptHandler() implements PromptHandlerInterface)` maybe

docs/server-builder.md
```diff
@@ -362,6 +362,65 @@ the handler's method name and docblock.
...
+integrations), reflection-based discovery does not apply. Implement an
+element-specific runtime interface instead and pass the instance to the
+Builder.
+
+| Element kind     | Interface                                |
+|------------------|------------------------------------------|
+| Tool             | `RuntimeToolHandlerInterface`            |
+| Prompt           | `RuntimePromptHandlerInterface`          |
+| Resource template| `RuntimeResourceTemplateHandlerInterface`|
+| Resource         | `RuntimeHandlerInterface`                |
```
chr-hertel:
```suggestion
| Resource         | `RuntimeResourceHandlerInterface`        |
```

e0ipso:
@chr-hertel if I understand correctly your preference is:

1. We add a new helper `Builder::add(ElementInterface $element, ElementHandlerInterface $handler)` that is documented as the primary registration.
2. The reflection-based method is documented as sugar syntax.
3. This `Builder::add(...)` registers using a dedicated, and new, loader. This leaves the `ArrayLoader` alone (instead of branching inside and sniffing interfaces).

chr-hertel:
Might be worth exploring if working, but would like to have more opinions here @soyuka @Nyholm @CodeWithKyrian

e0ipso:
@chr-hertel I think implemented your suggestion. I learned more about the project in the process :relaxed:

e0ipso:
@chr-hertel I had an LLM update the PR description to reflect the conversation so far, and the new direction. I'm notifying you in case you want to revise your :+1:

src/Capability/Registry/ElementReference.php
```diff
@@ -11,8 +11,11 @@
 
 namespace Mcp\Capability\Registry;
 
+use Mcp\Server\Handler\ElementHandlerInterface;
+
 /**
- * @phpstan-type Handler \Closure|array{0: object|string, 1: string}|string
+ * @phpstan-type CallableHandler \Closure|array{0: object|string, 1: string}|string
+ * @phpstan-type Handler CallableHandler|ElementHandlerInterface
```
e0ipso: I am splitting this in two for convenience and documentation brevity.

src/Capability/Registry/ReferenceHandler.php
```diff
@@ -33,33 +39,54 @@ public function __construct(
...
-            return \call_user_func($reference->handler, ...$arguments);
+            return \call_user_func($handler, ...$arguments);
         }
 
-        if (\is_array($reference->handler)) {
-            [$className, $methodName] = $reference->handler;
+        if (\is_array($handler)) {
+            [$className, $methodName] = $handler;
             $reflection = new \ReflectionMethod($className, $methodName);
             $instance = $this->getClassInstance($className);
```
e0ipso, line 74: This looks disruptive at first glance but it's just _using the new variable for `$reference->handler`_

src/Capability/Registry/Loader/ExplicitElementLoader.php
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Translates `Builder::add()` definition+handler pairs into Registry entries.
+ *
+ * Each pair is registered with `isManual: true`, matching the precedence used
+ * by the `ArrayLoader` for closure/array/string handlers.
+ *
+ * @author Mateu Aguiló Bosch <mateu.aguilo.bosch@gmail.com>
+ */
+final class ExplicitElementLoader implements LoaderInterface
```
e0ipso, line 35: IMHO this feels a bit excessive indirection, but I believe this is the idiomatic way to accomplish this.

Perhaps we can avoid this if we add the elements to the registry directly in the builder. However, that approach will break the current responsibility encapsulation.

Guidance/validation here is appreciated.

src/Server/Handler/ElementHandlerInterface.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Mcp\Server\Handler;
+
+/**
+ * Marker contract shared by all explicit element handler interfaces (tool,
+ * resource, resource template, prompt). Used as a single type hint where any
+ * element handler is acceptable.
+ *
+ * @author Mateu Aguiló Bosch <mateu.aguilo.bosch@gmail.com>
+ */
+interface ElementHandlerInterface
```
mglaman, line 21: Devils advocate here: is an empty interface here better than explicit unions of `PromptHandlerInterface|ResourceHandlerInterface|ResourceTemplateHandlerInterface|ToolHandlerInterface`

Edit: had to scroll back up and see `callable|array|string|ElementHandlerInterface $handler,`... yes, yes it is.

mglaman (commented):
This seems much better

e0ipso:
@chr-hertel @Nyholm et al. Please let me know if you prefer that I take this in a different direction instead.

chr-hertel:
Will come back to this in a minute :+1: first sorting small stuff right now, thanks already!

docs/mcp-elements.md
```diff
@@ -42,6 +42,11 @@ Each capability can be registered using two methods:
 
 For manual registration details, see [Server Builder Manual Registration](server-builder.md#manual-capability-registration).
 
+For runtime, config-driven elements whose shape is not known at compile time
+(e.g. bridging configuration entities into MCP elements), see [Explicit
+element registration](server-builder.md#explicit-element-registration) in the
+Server Builder docs.
```
chr-hertel: can use more width, i basically default to 120 chars

```suggestion
For runtime, config-driven elements whose shape is not known at compile time (e.g. bridging configuration entities into
MCP elements), see [Explicit element registration](server-builder.md#explicit-element-registration) in the Server Builder docs.
```

docs/server-builder.md
```diff
@@ -362,6 +362,67 @@ the handler's method name and docblock.
 
 For more details on MCP elements, handlers, and attribute-based discovery, see [MCP Elements](mcp-elements.md).
 
+### Explicit element registration
+
+When an element's name, schema, or description is only known at runtime
+(for example, a Drupal module bridging configuration entities into MCP
```
chr-hertel: I would favor to name drop the Drupal mcp module in the front readme instead:


```suggestion
(for example, a configuration or database entities into MCP
```

docs/server-builder.md
chr-hertel, line 1: please default line length to 120 chars

src/Capability/Registry/Loader/ExplicitElementLoader.php
chr-hertel: I like having this as a basic loader, I wonder tho if this should be the `ArrayLoader` or `InMemoryLoader` - and the `ArrayLoader` deserves a different name 🤔

src/Capability/Registry/ReferenceHandler.php
chr-hertel: bloating the diff a bit, but fair enough to me

chr-hertel (changes requested):
I like this very much already - switched to nit-pick-mode while reviewing - first round of comments tho. Thanks @e0ipso!

chr-hertel:
Also needs a CHANGELOG.md entry please :+1:

docs/server-builder.md
```diff
@@ -362,6 +362,67 @@ the handler's method name and docblock.
...
+);
+
+$server = Server::builder()
+    ->add($tool, new WeatherHandler())
+    ->build();
+```
+
+`Builder::add()` validates the pairing at registration time. Pairing a
+`Tool` definition with, for example, a `PromptHandlerInterface` raises
+`Mcp\Exception\ConfigurationException`. The schema value object validates
```
chr-hertel: this is actually implemented differently => raising a `TypeError`, but I'd actually favor to throw a package specific exception over a `TypeError`, e.g. `Mcp\Exception\InvalidArgumentException`

src/Server/Builder.php
```diff
@@ -139,15 +151,36 @@ final class Builder
...
      */
     private array $prompts = [];
 
+    /**
+     * @var list<array{definition: Tool, handler: ToolHandlerInterface}>
+     */
+    private array $explicitTools = [];
+
+    /**
+     * @var list<array{definition: \Mcp\Schema\Resource, handler: ResourceHandlerInterface}>
```
chr-hertel:
```suggestion
     * @var list<array{definition: Resource, handler: ResourceHandlerInterface}>
```

chr-hertel: oh, that gets you into a mess with `resource` type i guess?

src/Capability/Registry/Loader/ExplicitElementLoader.php
soyuka: IMO this is also in-memory, lets keep this name but rename `ArrayLoader` to `ReflectedElementLoader`?

src/Capability/Registry/Loader/ExplicitElementLoader.php
```diff
@@ -0,0 +1,65 @@
...
+use Mcp\Schema\Tool;
+use Mcp\Server\Handler\PromptHandlerInterface;
+use Mcp\Server\Handler\ResourceHandlerInterface;
+use Mcp\Server\Handler\ResourceTemplateHandlerInterface;
+use Mcp\Server\Handler\ToolHandlerInterface;
+
+/**
+ * Translates `Builder::add()` definition+handler pairs into Registry entries.
+ *
+ * Each pair is registered with `isManual: true`, matching the precedence used
```
soyuka: probably revise this if we remove `isManual` (https://github.com/modelcontextprotocol/php-sdk/pull/302)

src/Capability/Registry/ReferenceHandler.php
```diff
@@ -33,33 +39,54 @@ public function __construct(
...
+                $handler instanceof ResourceHandlerInterface => $handler->read(
+                    $arguments['uri'] ?? throw new InvalidArgumentException('Resource dispatch requires a "uri" argument.'),
+                    $client,
+                ),
+                $handler instanceof ResourceTemplateHandlerInterface => $handler->read(
+                    $arguments['uri'] ?? throw new InvalidArgumentException('Resource template dispatch requires a "uri" argument.'),
+                    array_diff_key($callArgs, ['uri' => null]),
+                    $client,
+                ),
+                default => throw new InvalidArgumentException(\sprintf('Unsupported %s implementation: %s.', ElementHandlerInterface::class, $handler::class)),
```
soyuka: I kinda dislike this part of the PR. Reference handler should not be aware of all these implementations its responsibility is to handle `callable` resolver. Also if a `ElementHandler` implements both `ToolHandlerInterface` and `PromptHandlerInterface` we'd just pick the first one?  I'd rather see the `ExplicitElementLoader::load` do something like this:

```
  foreach ($this->tools as $entry) {
      $handler  = $entry['handler'];
      $callable = static fn (array $_arguments, ClientGateway $_gateway): mixed
          => $handler->execute($_arguments, $_gateway);
      $registry->registerTool($entry['definition'], $callable, true);
  }
  // same shape for resources / resource templates / prompts
```

Quite close to what we do in API platform actually. Also this avoids coupling ReferenceHandler with all the types of handler.

src/Server/Builder.php
soyuka: resource is a reserved php word, even in a class name we should rather rename this class.

src/Capability/Registry/ResourceReference.php
```diff
@@ -27,7 +28,7 @@ class ResourceReference extends ElementReference
      */
     public function __construct(
         public readonly Resource $resource,
-        callable|array|string $handler,
+        callable|array|string|ElementHandlerInterface $handler,
```
soyuka: I'd revert these as well and keep the use of callables.

soyuka (changes requested):

e0ipso:
Thanks for all the feedback so far. I am quietly giving space for others to chime in before addressing the feedback.

src/Capability/Registry/ResourceReference.php
e0ipso: @soyuka can you expand a bit more on this? I am not sure what you mean exactly. I don't want to implement your feedback erroneously and have you go through an additional round of review/feedback.

soyuka: if the code gets moved to `ExplicitElementLoader` as I suggested above we only need `callable` and don't need this method signature change

chr-hertel:
Just integrated #302 into main, rebase unlocked here :muscle:

docs/server-builder.md
e0ipso: Fixed. The table was rewritten and now lists the four `*HandlerInterface` (Tool / Resource / ResourceTemplate / Prompt), so the inconsistency is gone.

docs/mcp-elements.md
e0ipso: Fixed. Rewrapped `docs/mcp-elements.md` to 120 chars and applied your exact suggestion.

docs/server-builder.md
e0ipso: Fixed. The section now reads exactly as you suggested.

docs/server-builder.md
e0ipso: Fixed. New docs wrap at 120.

docs/server-builder.md
e0ipso: Fixed. `Builder::add()` now throws `Mcp\Exception\InvalidArgumentException` on a mismatched pairing instead of letting PHP raise a `TypeError`.

src/Capability/Registry/Loader/ExplicitElementLoader.php
e0ipso: Done. `ArrayLoader` is now `ReflectedElementLoader`.

src/Capability/Registry/Loader/ExplicitElementLoader.php
e0ipso: Nothing to revise — `isManual` is already gone from `src/`. Whichever of #294 or #302 lands second won't have a flag to remove.

src/Capability/Registry/ReferenceHandler.php
e0ipso: Fixed along the lines you sketched. `ReferenceHandler` no longer imports or branches on any handler interface; `ExplicitElementLoader::load()` wraps each handler in a `Closure` and registers it through the existing `registerTool` / `registerResource` / `registerResourceTemplate` / `registerPrompt`. The "implements both" ambiguity goes away too.

src/Server/Builder.php
e0ipso: Fixed. `Mcp\Schema\Resource` is now `Mcp\Schema\ResourceDefinition` (no alias).

src/Capability/Registry/ResourceReference.php
e0ipso: Got it, thanks. The handler-type widening on the *Reference classes and on `Registry` / `RegistryInterface` is reverted, so those still take a plain `callable`. The interface-aware dispatch lives in `ExplicitElementLoader` now.

e0ipso:
I _think_ this should be pretty close. The Drupal implementation is quite clean when using this patch, I am so happy about this one!!

src/Capability/Registry/ReferenceHandler.php
```diff
@@ -113,6 +124,16 @@ private function prepareArguments(\ReflectionFunctionAbstract $reflection, array
...
+            } elseif (
+                $type instanceof \ReflectionNamedType
+                && 'array' === $type->getName()
+                && \in_array($paramName, self::ARGUMENT_BAG_PARAMETERS, true)
+            ) {
+                $skipKeys = array_merge(
+                    self::RESERVED_ARGUMENT_KEYS,
+                    array_values(array_diff($allParamNames, [$paramName])),
+                );
+                $finalArgs[$paramPosition] = array_diff_key($arguments, array_flip($skipKeys));
```
e0ipso: This is the ugly part. I am open to suggestions.

src/Capability/Registry/Loader/ExplicitElementLoader.php
```diff
@@ -0,0 +1,80 @@
...
+                static fn (string $uri, ClientGateway $client) => $handler->read($uri, $client),
+            );
+        }
+
+        foreach ($this->resourceTemplates as $entry) {
+            $handler = $entry['handler'];
+            $registry->registerResourceTemplate(
+                $entry['definition'],
+                static fn (string $uri, array $variables, ClientGateway $client) => $handler->read($uri, $variables, $client),
+            );
```
chr-hertel: Third argument `completionProviders` basically lost, not sure how to tackle that - would accept that for now. Same for prompts below.

src/Server/Builder.php
```diff
@@ -556,11 +614,17 @@ public function build(): Server
             $this->gcDivisor,
         );
 
-        // ArrayLoader runs before DiscoveryLoader so manual entries are seen first; DiscoveryLoader's
-        // identity check then preserves them against same-name discovered entries.
+        // ReflectedElementLoader runs before DiscoveryLoader so manual entries are seen first;
```
chr-hertel:
```suggestion
        // ExplicitElementLoader and ReflectedElementLoader run before DiscoveryLoader so manual entries are seen first;
```

docs/server-builder.md
```diff
@@ -398,6 +398,62 @@ the handler's method name and docblock.
 
 For more details on MCP elements, handlers, and attribute-based discovery, see [MCP Elements](mcp-elements.md).
 
+### Explicit element registration
+
+When an element's name, schema, or description is only known at runtime (for example, the Drupal `mcp` module exposing
```
chr-hertel: Let's move the Drupal reference to README section "PHP Libraries Using the MCP SDK"

src/Capability/Registry/ReferenceHandler.php
chr-hertel: i'm a bit worried about the impact to tools, that actually have `array $argument`, but we don't know at this level if we're dealing with an interface implementation, right?

just copying your concern i guess ... 🤔

chr-hertel (commented):
Nit pick level, only the "ugly part" got me thinking as well ...
Need to let that sink in, but those small items is actionable already.
Thanks @e0ipso!

src/Capability/Registry/Loader/ExplicitElementLoader.php
soyuka: can't we just add the completionProvider? there'd be a feature gap between the two loaders if not?

chr-hertel: was my thought as well, but we only have `add(Prompt, PromptHandler)` at the moment.

i don't think it should be part of the `Mcp\Schema\Prompt` instance, so it could be an optional third argument?
`add(Prompt, PromptHandler, ?CompletionProvider)`

src/Capability/Registry/ReferenceHandler.php
soyuka: Proposed fix pushed at [soyuka/php-sdk@c214a5d](https://github.com/soyuka/php-sdk/commit/c214a5d10ab795c81f18dde9e778c9dc0172f7a1).

The idea is to bind `ExplicitElementLoader` closures to `ReferenceHandler::class` as their closure scope. `ReferenceHandler::handle()` short-circuits when `getClosureScopeClass()?->getName() === self::class` and invokes the closure with the raw argument bag. This drops the magic-name heuristic (`array $arguments` / `array $variables` matched by parameter name) and the `RESERVED_ARGUMENT_KEYS` / `ARGUMENT_BAG_PARAMETERS` consts.


An alternative would be a wrapper class + widening, but its really nitpicking and probably not worth it.

Feel free to cherry-pick the commit.

soyuka:
Nice I like this implementation better! We could've left the `Resource` rename in another PR to lower the diff but we should've taken care of that anyways...

chr-hertel (approved):
Thanks @e0ipso - works for me like this 👍

soyuka (approved):

soyuka:
Thanks!

e0ipso:
Thank you!

e0ipso:
Crossing fingers for a `v0.6.0` soon :crossed_fingers:

PR #297, opened 2026-05-02
[Schema] fix: allow URIs without double slash in Resource and ResourceTemplate (RFC 3986)

### LINKED ISSUE #293: [Schema] URI Regex doesn't seem to be correct
https://github.com/modelcontextprotocol/php-sdk/blob/fb2c8c2ee4ab2791239c5f534bb07bfb7589d4e8/src/Schema/Resource.php#L46

If a Resource is using a [custom scheme for its URI](https://modelcontextprotocol.io/specification/2025-11-25/server/resources), it must follow [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986). However the part after the schema: (the 'heir-part') does not have to start with two slashes. See the various productions in [the path documentation](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) (including path-empty), and [some examples](https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.2).

Please consider removing the two required slashes in the path component of URI_PATTERN. Thanks!

blackwell-systems:
## Summary

The URI regex in `Resource` and `ResourceTemplate` requires `://` after the scheme, rejecting valid URIs where the hier-part does not start with `//`. Per [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3), the authority component (and its `//` prefix) is optional.

## What breaks

```php
// All of these throw InvalidArgumentException:
new Resource(uri: 'urn:isbn:0451450523', name: 'book');
new Resource(uri: 'mailto:user@example.com', name: 'contact');
new Resource(uri: 'data:text/plain;base64,SGVsbG8=', name: 'data');
new Resource(uri: 'config:myapp/settings', name: 'settings');

new ResourceTemplate(uriTemplate: 'config:{key}', name: 'config');
```

The [MCP spec](https://modelcontextprotocol.io/specification/2025-11-25/server/resources) allows custom URI schemes following RFC 3986.

## Fix

Changed `://` to `:` in both regex patterns:

```diff
- '/^[a-zA-Z][a-zA-Z0-9+.-]*:\/\/[^\s]*$/'
+ '/^[a-zA-Z][a-zA-Z0-9+.-]*:[^\s]*$/'

- '/^[a-zA-Z][a-zA-Z0-9+.-]*:\/\/.*{[^{}]+}.*/'
+ '/^[a-zA-Z][a-zA-Z0-9+.-]*:.*{[^{}]+}.*/'
```

URIs with `://` (like `file://path`) continue to match.

## Test plan

- [x] Added data-provider tests for `urn:`, `mailto:`, `data:`, and custom schemes without `//`
- [x] Added data-provider tests for resource templates without `//`
- [x] All 20 tests pass (`phpunit tests/Unit/Schema/`)
- [x] Existing tests unchanged

Fixes #293

### DIFF
--- a/tests/Unit/Schema/ResourceTemplateTest.php
+++ b/tests/Unit/Schema/ResourceTemplateTest.php
@@ -46,6 +46,25 @@ class ResourceTemplateTest extends TestCase
         );
     }
 
+    #[DataProvider('provideValidTemplatesWithoutDoubleSlash')]
+    public function testConstructorAcceptsTemplatesWithoutDoubleSlash(string $uriTemplate): void
+    {
+        $resource = new ResourceTemplate(
+            uriTemplate: $uriTemplate,
+            name: 'test-template',
+        );
+
+        $this->assertInstanceOf(ResourceTemplate::class, $resource);
+        $this->assertSame($uriTemplate, $resource->uriTemplate);
+    }
+
+    public static function provideValidTemplatesWithoutDoubleSlash(): iterable
+    {
+        yield 'custom scheme without slashes' => ['config:{key}'];
+        yield 'custom scheme with slashes' => ['config://{key}'];
+        yield 'urn-style template' => ['urn:resource:{id}'];
+    }
+
     public function testFromArrayValid(): void
     {
         $resource = ResourceTemplate::fromArray([
--- a/tests/Unit/Schema/ResourceTemplateTest.php
+++ b/tests/Unit/Schema/ResourceTemplateTest.php
@@ -46,8 +46,8 @@ class ResourceTemplateTest extends TestCase
         );
     }
 
-    #[DataProvider('provideValidTemplatesWithoutDoubleSlash')]
-    public function testConstructorAcceptsTemplatesWithoutDoubleSlash(string $uriTemplate): void
+    #[DataProvider('provideValidTemplates')]
+    public function testConstructorAcceptsTemplates(string $uriTemplate): void
     {
         $resource = new ResourceTemplate(
             uriTemplate: $uriTemplate,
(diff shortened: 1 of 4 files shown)


tests/Unit/Schema/ResourceTemplateTest.php
```diff
@@ -46,6 +46,25 @@ public function testConstructorInvalid(): void
...
+        $resource = new ResourceTemplate(
+            uriTemplate: $uriTemplate,
+            name: 'test-template',
+        );
+
+        $this->assertInstanceOf(ResourceTemplate::class, $resource);
+        $this->assertSame($uriTemplate, $resource->uriTemplate);
+    }
+
+    public static function provideValidTemplatesWithoutDoubleSlash(): iterable
```
chr-hertel:
```suggestion
    public static function provideValidTemplates(): iterable
```

tests/Unit/Schema/ResourceTest.php
```diff
@@ -46,6 +46,27 @@ public function testConstructorInvalid(): void
...
+        $resource = new Resource(
+            uri: $uri,
+            name: 'test-resource',
+        );
+
+        $this->assertInstanceOf(Resource::class, $resource);
+        $this->assertSame($uri, $resource->uri);
+    }
+
+    public static function provideValidUrisWithoutDoubleSlash(): iterable
```
chr-hertel:
```suggestion
    public static function provideValidUris(): iterable
```

chr-hertel (approved):
Thanks @blackwell-systems - briefly renamed some test methods while merging, JFYI.

Issue #298, opened 2026-05-04
Cannot pass null to nullable enum parameters


SBNTT:
**Describe the bug**
When an MCP tool is defined with a nullable Enum parameter (`?MyEnum $param = null`), the generated JSON schema includes "null" in the type array, but does not include null in the enum array.

Because JSON Schema requires the value to strictly match one of the elements in the enum array when the keyword is present, validation fails if a client sends an explicit null payload for this parameter.

**To Reproduce**
```php
enum MyEnum: string
{
    case FOO = 'FOO';
    case BAR = 'BAR';
}

#[McpTool]
public function myTool(?MyEnum $myEnum = null): int
{
    return 42;
}
```
**Actual behavior**
The `tools/list` endpoint returns the following schema:
```json
{
  "tools": [
    {
      "name": "myTool",
      "inputSchema": {
        "type": "object",
        "properties": {
          "myEnum": {
            "type": [
              "null",
              "string"
            ],
            "default": null,
            "enum": [
              "FOO",
              "BAR"
            ]
          }
        }
      }
    }
  ]
}
```

When calling the tool with an explicit null 
```json
{"name": "myTool", "arguments": {"myEnum": null}}
```
the following error is thrown:
```json
{
  error: "MCP error -32602: Invalid parameters for tool 'myTool': Property '/myEnum': Value must be one of the allowed values: "FOO", "BAR"."
}
```

**Expected behavior**
The generated schema should accept explicit null payloads when the enum type is nullable

chr-hertel:
Fixed with #299 - thanks for the report!

PR #299, opened 2026-05-04
fix: Cannot pass null to nullable enum parameters

SBNTT:
fix proposal for #298


chr-hertel (approved):
Thanks for reporting & patching @SBNTT! 👍

PR #300, opened 2026-05-05
Add PSR-3 compliant `exception` context when logging exceptions

svenmuennich:
This PR consistently adds caught exceptions to the logging context as `exception`, following the standard defined by [PSR-3](https://www.php-fig.org/psr/psr-3/).

## Motivation and Context

Previously almost no caught exceptions where included in the context passed to respective PSR logging calls, making it very hard for downstream tooling to e.g. filter certain errors from reporting.

## How Has This Been Tested?

Added unit test coverage for inclusion of exceptions in logging.

## Breaking Changes

n/a

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

I decided to include objects of type `Throwable` in the logging, even though PSR-3 only covers objects of types `Exception`. My reasoning was based on the following statement in the spec:
>Implementors MUST still verify that the 'exception' key is actually an Exception before using it as such, as it MAY contain anything.

### DIFF
--- a/src/Capability/Discovery/DocBlockParser.php
+++ b/src/Capability/Discovery/DocBlockParser.php
@@ -47,9 +47,7 @@ class DocBlockParser
         } catch (\Throwable $e) {
             // Log error or handle gracefully if invalid DocBlock syntax is encountered
             $this->logger->warning('Failed to parse DocBlock', [
-                'error' => $e->getMessage(),
                 'exception' => $e,
-                'exception_trace' => $e->getTraceAsString(),
             ]);
 
             return null;
--- a/src/Capability/Discovery/DocBlockParser.php
+++ b/src/Capability/Discovery/DocBlockParser.php
@@ -48,6 +48,7 @@ class DocBlockParser
             // Log error or handle gracefully if invalid DocBlock syntax is encountered
             $this->logger->warning('Failed to parse DocBlock', [
                 'error' => $e->getMessage(),
+                'exception' => $e,
                 'exception_trace' => $e->getTraceAsString(),
             ]);
 
--- a/src/Capability/Discovery/SchemaValidator.php
+++ b/src/Capability/Discovery/SchemaValidator.php
@@ -88,8 +88,6 @@ class SchemaValidator
         } catch (\Throwable $e) {
             $this->logger->error('MCP SDK: JSON Schema validation failed internally.', [
                 'exception' => $e,
-                'exception_message' => $e->getMessage(),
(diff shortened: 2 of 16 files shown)


src/Capability/Discovery/DocBlockParser.php
```diff
@@ -48,6 +48,7 @@ public function parseDocBlock(string|false|null $docComment): ?DocBlock
             // Log error or handle gracefully if invalid DocBlock syntax is encountered
             $this->logger->warning('Failed to parse DocBlock', [
                 'error' => $e->getMessage(),
+                'exception' => $e,
                 'exception_trace' => $e->getTraceAsString(),
```
chr-hertel: we can skip the others then I'd guess:

```suggestion
                'exception' => $e,
```

src/Capability/Discovery/SchemaValidator.php
```diff
@@ -87,6 +87,7 @@ public function validateAgainstJsonSchema(mixed $data, array|object $schema): ar
             $result = $validator->validate($dataToValidate, $schemaObject);
         } catch (\Throwable $e) {
             $this->logger->error('MCP SDK: JSON Schema validation failed internally.', [
+                'exception' => $e,
                 'exception_message' => $e->getMessage(),
                 'exception_trace' => $e->getTraceAsString(),
```
chr-hertel:
```suggestion
                'exception' => $e,
```

src/Client/Transport/HttpTransport.php
```diff
@@ -170,7 +170,7 @@ public function close(): void
                 $this->httpClient->sendRequest($request);
                 $this->logger->info('Session closed', ['session_id' => $this->sessionId]);
             } catch (\Throwable $e) {
-                $this->logger->warning('Failed to close session', ['error' => $e->getMessage()]);
+                $this->logger->warning('Failed to close session', ['error' => $e->getMessage(), 'exception' => $e]);
```
chr-hertel:
```suggestion
                $this->logger->warning('Failed to close session', ['exception' => $e]);
```

chr-hertel (changes requested):
Thanks for that impulse @svenmuennich - nice house-keeping. I'd say we could go one step further, WDYT?

svenmuennich:
> I'd say we could go one step further, WDYT?

Sure! I just did another sweep and made sure that only complete `exception`s are logged 👍

chr-hertel (approved):
Thank you @svenmuennich!

soyuka (approved):

PR #302, opened 2026-05-11
[Server] Untangle origin tracking from Registry

chr-hertel:
## Summary

- Drops `isManual` from `ElementReference` (and the four subclasses); origin is no longer a property of the element.
- Strips `clear()` / `getDiscoveryState()` / `setDiscoveryState()` from `RegistryInterface`; the Registry is now a flat last-write-wins map.
- Adds idempotent `unregisterTool/Resource/ResourceTemplate/Prompt(string)` to `RegistryInterface`.
- Rewrites `DiscoveryLoader` with a private owned-set diff so re-running `load()` only removes elements this loader itself contributed — preserves runtime registrations between discovery passes.
- Introduces `ChainLoader` for explicit composition; `Builder::build()` composes loaders internally so the user-facing API is unchanged.
- Manual-over-discovered precedence is preserved via loader ordering (`[user, discovery, manual]` — last-write-wins).

Supersedes #251: the runtime-registration regression that PR was fixing dissolves by construction, without adding a second origin flag (`isDiscovered`).

## Test plan

- [x] `make ci` — phpstan clean, php-cs-fixer clean, 652 unit tests / 2146 assertions pass.
- [x] `make inspector-tests` — 87 tests pass (7 environmental skips, same as `main`).
- [x] New `DiscoveryLoaderTest::testLoadDoesNotUnregisterRuntimeAdditions` locks in the regression that #251 was after.
- [x] `combined-registration` example still works end-to-end; the `config://priority` resource correctly resolves to the manual override (3 inspector snapshots updated to reflect this — they were previously showing the discovered version winning, which was a latent bug).

## Breaking changes

- `ElementReference->isManual` property removed.
- `RegistryInterface::register*()` signatures lose the `bool $isManual = false` parameter.
- `RegistryInterface::clear()`, `getDiscoveryState()`, `setDiscoveryState()` removed (only correct caller was `DiscoveryLoader`, which no longer needs them).

Pre-1.0, internal area — no userland impact known beyond what's documented above.

### DIFF
--- a/src/Capability/Registry/Loader/DiscoveryLoader.php
+++ b/src/Capability/Registry/Loader/DiscoveryLoader.php
@@ -12,7 +12,10 @@
 namespace Mcp\Capability\Registry\Loader;
 
 use Mcp\Capability\Discovery\DiscovererInterface;
+use Mcp\Capability\Discovery\DiscoveryState;
 use Mcp\Capability\RegistryInterface;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 
 /**
  * @internal
@@ -21,10 +24,12 @@ use Mcp\Capability\RegistryInterface;
  */
 final class DiscoveryLoader implements LoaderInterface
 {
+    private DiscoveryState $owned;
+
     /**
-     * @param string[]       $scanDirs
-     * @param array|string[] $excludeDirs
-     * @param string[]       $namePatterns
+     * @param string[] $scanDirs
+     * @param string[] $excludeDirs
+     * @param string[] $namePatterns
      */
     public function __construct(
         private string $basePath,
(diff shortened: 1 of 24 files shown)


chr-hertel:
@soyuka that could be an alternative approach to #251, removing the leaky `isManual` and isolating it in that `DiscoveryLoader`

soyuka:
indeed, its a bit more complex then just a boolean though but I like the approach let me review.

src/Capability/Registry/Loader/DiscoveryLoader.php
```diff
@@ -21,24 +24,162 @@
  */
 final class DiscoveryLoader implements LoaderInterface
 {
+    private DiscoveryState $owned;
```
soyuka, line 27:
```suggestion
    private DiscoveryState $owned = new DiscoveryState();
```

src/Capability/Registry/Loader/DiscoveryLoader.php
```diff
@@ -21,24 +24,162 @@
...
      */
     public function __construct(
         private string $basePath,
         private array $scanDirs,
         private array $excludeDirs,
         private DiscovererInterface $discoverer,
         private array $namePatterns = DiscovererInterface::DEFAULT_NAME_PATERNS,
+        private LoggerInterface $logger = new NullLogger(),
     ) {
+        $this->owned = new DiscoveryState();
```
soyuka, line 42:
```suggestion
```

src/Server/Builder.php
```diff
@@ -526,28 +527,29 @@ public function build(): Server
...
+                $loaders[] = new DiscoveryLoader($this->discoveryBasePath, $this->discoveryScanDirs, $this->discoveryExcludeDirs, $discoverer, $this->discoveryNamePatterns, $logger);
             } else {
                 $logger->warning('File-based discovery requires symfony/finder. Skipping automatic discovery. Run: composer require symfony/finder');
             }
         }
 
-        foreach ($loaders as $loader) {
-            $loader->load($registry);
-        }
+        $loader = 1 === \count($loaders) ? $loaders[0] : new ChainLoader($loaders);
```
soyuka:
```suggestion
        $loader = new ChainLoader($loaders);
```
is probably fine

soyuka (dismissed):
Indeed this can replace my monkey patch its cleaner :)

src/Capability/Registry/Loader/DiscoveryLoader.php
chr-hertel: won't work, see https://3v4l.org/i43rH

src/Capability/Registry/Loader/DiscoveryLoader.php
chr-hertel: we could shift it to a constructor argument, but if we only do it for the way of how to instantiate the state, i'd rather keep it like this

soyuka (approved):

PR #304, opened 2026-05-12
[CI] Add weekly conformance run against latest upstream release

soyuka:
The PR pipeline only runs conformance when someone opens a PR. Upstream `@modelcontextprotocol/conformance` releases between PRs go untested.

This workflow:
- runs the same conformance commands on a weekly cron (Mondays 06:00 UTC),
- can be triggered manually from the Actions tab (`workflow_dispatch`),
- pins to `@modelcontextprotocol/conformance@latest` to pick up new releases,
- opens (or comments on) a tracking issue labelled `conformance-weekly` on scheduled failures.

## Test plan

- [ ] Manual run from the Actions tab succeeds (or fails cleanly with an issue filed)

### DIFF
--- a/.github/workflows/conformance-weekly.yaml
+++ b/.github/workflows/conformance-weekly.yaml
@@ -1,0 +1,98 @@
+name: conformance-weekly
+
+# Runs the MCP conformance suite weekly against the latest
+# @modelcontextprotocol/conformance release. The on:pull_request pipeline
+# pins to whatever version is available at PR time; this schedule catches
+# upstream releases that add scenarios between PRs.
+on:
+  schedule:
+    - cron: '0 6 * * 1' # Mondays 06:00 UTC
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  server:
+    name: conformance / server (latest)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version: '22'
+      - run: composer install --prefer-dist --no-progress --no-interaction
+      - name: Start conformance server
+        run: |
+          mkdir -p tests/Conformance/sessions tests/Conformance/logs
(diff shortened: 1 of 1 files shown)


.github/workflows/conformance-weekly.yaml
```diff
@@ -0,0 +1,163 @@
...
+    name: Open issue on failure
+    runs-on: ubuntu-latest
+    needs: [conformance-server, conformance-client]
+    if: failure() && github.event_name == 'schedule'
+    steps:
+      - name: File or update tracking issue
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const marker = '<!-- conformance-weekly-failure -->';
```
soyuka: not sure we really need this notify section, what do you think about running this weekly?

.github/workflows/conformance-weekly.yaml
```diff
@@ -0,0 +1,163 @@
...
+        uses: "ramsey/composer-install@v4"
+
+      - name: Show conformance version
+        run: npx --yes @modelcontextprotocol/conformance@latest --version
+
+      - name: Start conformance server
+        run: |
+          mkdir -p tests/Conformance/sessions tests/Conformance/logs
+          chmod -R 777 tests/Conformance/sessions tests/Conformance/logs
+          docker compose -f tests/Conformance/Fixtures/docker-compose.yml up -d
```
chr-hertel, line 30: I know it's copy&paste mostly, but we could go with `--wait` over sleep maybe?

catched my eye only now => #305

soyuka:
global opinion about this? should I clean it up?

chr-hertel:
+1 from my end, yes :)
was wondering about creating an issue, but it's the best fit for our tools

.github/workflows/conformance-weekly.yaml
chr-hertel: i'm in favor for giving it a go - if we don't act on it and things just pile up, we can kill it easily again

chr-hertel (approved):

PR #305, opened 2026-05-12
ci: replace sleep with compose --wait + nginx healthcheck

chr-hertel:
## Summary
- Replace the blind `sleep 5` after `docker compose up -d` in the conformance job with `--wait`, so CI proceeds as soon as services are actually ready.
- Add an HTTP healthcheck on the nginx service (via `wget`) so `--wait` blocks on real readiness rather than just the `running` state.

## Test plan
- [ ] Conformance / server job passes on this PR


chr-hertel:
yeah, okay, the sleep is easier :joy:

PR #307, opened 2026-05-13
[Server] Compose StreamableHttpTransport security middleware via defaultMiddleware() factory

### LINKED ISSUE #280: Hardcoded Wildcard CORS (Access-Control-Allow-Origin: *)

### LINKED ISSUE #306: Streamable HTTP accepts unsupported or malformed MCP-Protocol-Version headers instead of returning 400
**Describe the bug**

After a normal Streamable HTTP initialization flow, the transport accepts requests with an unsupported or malformed `MCP-Protocol-Version` header and dispatches them normally with HTTP 200. The MCP Streamable HTTP specification for `2025-11-25` requires servers to respond with HTTP `400 Bad Request` for invalid or unsupported `MCP-Protocol-Version` values.

**To Reproduce**

1. Start a Streamable HTTP server using the PHP SDK.
2. Complete a normal `initialize` → `notifications/initialized` handshake.
3. Send a valid JSON-RPC request, such as `ping` or `tools/list`, with `MCP-Protocol-Version` set to `1900-01-01` (well-formed but unsupported) or `not-a-version` (malformed).
4. Observe the response.

**Expected behavior**

The server should reject the request before dispatching it:

```
HTTP/1.1 400 Bad Request
```

**Logs**

Unsupported version:
```text
POST /mcp HTTP/1.1
MCP-Protocol-Version: 1900-01-01
Content-Type: application/json
Accept: application/json, text/event-stream
Mcp-Session-Id: <session-id-from-initialize>

{"jsonrpc":"2.0","id":"bad-version-1","method":"tools/list","params":{}}

HTTP/1.1 200 OK

{"jsonrpc":"2.0","id":"bad-version-1","result":{"tools":[...]}}
```

Malformed version:
```text
POST /mcp HTTP/1.1
MCP-Protocol-Version: not-a-version
Content-Type: application/json
Accept: application/json, text/event-stream
Mcp-Session-Id: <session-id-from-initialize>

{"jsonrpc":"2.0","id":"bad-version-2","method":"tools/list","params":{}}

HTTP/1.1 200 OK

{"jsonrpc":"2.0","id":"bad-version-2","result":{...}}
```

sveneld:
## Summary

Replaces the implicit `instanceof + array_unshift` injection of CORS handling in `StreamableHttpTransport` with an explicit, composable PSR-15 stack exposed through a public `StreamableHttpTransport::defaultMiddleware()` factory. Adds three middleware that together cover the recommended HTTP-level hardening for MCP Streamable HTTP:

- **`CorsMiddleware`** — secure-by-default (no `Access-Control-Allow-Origin` header is set; cross-origin browser requests are blocked). Configurable `allowedOrigins` reflect a matching origin and automatically emit `Vary: Origin` so shared caches don't poison.
- **`DnsRebindingProtectionMiddleware`** — validates `Origin`/`Host` against an allowlist of hostnames (localhost variants by default).
- **`ProtocolVersionMiddleware`** — rejects requests carrying an unsupported `Mcp-Protocol-Version` header with `400 Bad Request` (spec compliance).

The transport itself becomes oblivious to the middleware list. The `$middleware` constructor argument is nullable:

| Passed value | Effect |
|---|---|
| omitted / `null` | Default secure stack from `StreamableHttpTransport::defaultMiddleware()` |
| `[]` | No middleware at all (e.g. when fronted by an application that already handles CORS/host validation) |
| explicit list | Used verbatim; mix and match by spreading `...StreamableHttpTransport::defaultMiddleware()` |

`SESSION_HEADER` and a new `PROTOCOL_VERSION_HEADER` are promoted to public constants on the transport so middleware can reuse them without duplicating string literals.

## Why this design

The three open items below all reach for the same shape — a piece of mandatory security logic that the transport should apply unless the user opts out. PRs #260 and #277 each implement this with their own `instanceof MyMiddlewareClass` check + `array_unshift` inside the transport constructor. That pattern scales linearly with the number of "mandatory" middleware and breaks once a user supplies a custom implementation of the same concern under a different class name. Adding `ProtocolVersionMiddleware` for #306 would mean a third copy of the same pattern.

This PR collapses all three into a single composition model: the transport is a thin PSR-15 host, and the recommended stack lives in a factory the user can spread, filter, or replace. No `instanceof` magic, no surprise prepending, no position arguments to bypass.

### Relation to other work

- **#277 (Extract CORS handling into `CorsMiddleware`)** — same goal of replacing the inline `withCorsHeaders()` post-hook with a PSR-15 middleware. This PR additionally drops the `instanceof CorsMiddleware` auto-prepend in favour of `defaultMiddleware()` composition (which is what @CodeWithKyrian suggested in [#277 review](https://github.com/modelcontextprotocol/php-sdk/pull/277#issuecomment-4229684297) and @chr-hertel agreed with). Also addresses #280 (wildcard CORS default) the same way #277 does — `Access-Control-Allow-Origin` is no longer set by default. Adds `Vary: Origin` when reflecting a specific origin, which #277 doesn't.
- **#260 (DNS rebinding protection)** — same middleware concept and default allowlist (`localhost`, `127.0.0.1`, `[::1]`, `::1`); avoids the `instanceof DnsRebindingProtectionMiddleware` auto-prepend pattern for the same scalability reason.
- **#306 (Unsupported `MCP-Protocol-Version` accepted)** — new `ProtocolVersionMiddleware` rejects with `400 Bad Request` per spec. Tolerates a missing header (initialize round-trip / legacy clients don't send it); strict per-session version-match validation is out of scope for an HTTP-layer middleware and can be layered on if maintainers want it.

If maintainers prefer to land #260 / #277 separately first, this PR can be rebased onto them and reduced to the composition refactor + `ProtocolVersionMiddleware`. Happy to do that.

## BC breaks

- `StreamableHttpTransport::__construct` — the `$corsHeaders` parameter is removed. The `$middleware` parameter consequently shifts one position. Positional callers passing the old `$corsHeaders` argument will get a TypeError at construction; switch to named arguments or drop the argument. `logger`/`middleware` are now positions 4/5 instead of 5/6.
- Default `Access-Control-Allow-Origin` is no longer `*`. Browser clients on a different origin will be blocked unless `CorsMiddleware` is configured with explicit `allowedOrigins`.

`CHANGELOG.md` entry added under 0.6.0.

## Test plan

- [x] `vendor/bin/phpunit --testsuite=unit` — 38 new/updated tests pass (3 middleware × ~10 tests each + transport integration). Full suite: 1 error + 2 failures are pre-existing on `upstream/main` (`DocBlockParserTest` missing `Composer\Semver\VersionParser` dev-dep, two `DiscoveryTest` fixtures) — verified by running the same tests against vanilla `main` with the branch stashed.
- [x] `vendor/bin/php-cs-fixer fix --dry-run` — clean
- [x] `vendor/bin/phpstan analyse` — clean for changed code; the one pre-existing error in `DocBlockParserTest` (missing dev-dep `Composer\Semver\VersionParser`) is unchanged
- [x] Both OAuth examples (`oauth-keycloak`, `oauth-microsoft`) updated to spread the default stack so their behaviour is preserved

### DIFF
--- a/src/Server/Transport/Http/JsonRpcErrorResponse.php
+++ b/src/Server/Transport/Http/JsonRpcErrorResponse.php
@@ -17,10 +17,9 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\StreamFactoryInterface;
 
 /**
- * Builds the canonical JSON-RPC error response used by the HTTP transport
- * and its middleware: a PSR-7 response with the given HTTP status, a
- * `Content-Type: application/json` header, and a body containing a single
- * `Error::forInvalidRequest($message)` payload.
+ * Builds a PSR-7 response with the given HTTP status and a JSON-RPC
+ * `Error` payload as body. Caller decides which `Error::for*` factory
+ * to use so the JSON-RPC error code matches the failure semantics.
  *
  * @internal
  */
@@ -30,9 +29,9 @@ final class JsonRpcErrorResponse
         ResponseFactoryInterface $responseFactory,
         StreamFactoryInterface $streamFactory,
         int $statusCode,
-        string $message,
+        Error $error,
     ): ResponseInterface {
-        $body = json_encode(Error::forInvalidRequest($message), \JSON_THROW_ON_ERROR);
+        $body = json_encode($error, \JSON_THROW_ON_ERROR);
 
         return $responseFactory
             ->createResponse($statusCode)
--- a/src/Server/Transport/Http/JsonRpcErrorResponse.php
+++ b/src/Server/Transport/Http/JsonRpcErrorResponse.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ * This file is part of the official PHP MCP SDK.
+ *
+ * A collaboration between Symfony and the PHP Foundation.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Mcp\Server\Transport\Http;
+
+use Mcp\Schema\JsonRpc\Error;
+use Psr\Http\Message\ResponseFactoryInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\StreamFactoryInterface;
+
+/**
+ * Builds the canonical JSON-RPC error response used by the HTTP transport
+ * and its middleware: a PSR-7 response with the given HTTP status, a
+ * `Content-Type: application/json` header, and a body containing a single
+ * `Error::forInvalidRequest($message)` payload.
+ *
+ * @internal
+ */
+final class JsonRpcErrorResponse
+{
+    public static function create(
+        ResponseFactoryInterface $responseFactory,
+        StreamFactoryInterface $streamFactory,
+        int $statusCode,
+        string $message,
+    ): ResponseInterface {
+        $body = json_encode(Error::forInvalidRequest($message), \JSON_THROW_ON_ERROR);
+
+        return $responseFactory
+            ->createResponse($statusCode)
+            ->withHeader('Content-Type', 'application/json')
+            ->withBody($streamFactory->createStream($body));
+    }
+}
--- a/src/Server/Transport/Http/Middleware/CorsMiddleware.php
+++ b/src/Server/Transport/Http/Middleware/CorsMiddleware.php
@@ -49,7 +49,7 @@ final class CorsMiddleware implements MiddlewareInterface
      */
     public function __construct(
         private readonly array $allowedOrigins = [],
-        array $allowedMethods = ['GET', 'POST', 'DELETE', 'OPTIONS'],
+        array $allowedMethods = ['GET', 'POST', 'DELETE'],
         array $allowedHeaders = [
             'Accept',
             'Authorization',
--- a/src/Server/Transport/Http/Middleware/CorsMiddleware.php
+++ b/src/Server/Transport/Http/Middleware/CorsMiddleware.php
@@ -11,6 +11,7 @@
 
 namespace Mcp\Server\Transport\Http\Middleware;
 
+use Mcp\Exception\InvalidArgumentException;
 use Mcp\Server\Transport\StreamableHttpTransport;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
@@ -18,14 +19,16 @@ use Psr\Http\Server\MiddlewareInterface;
 use Psr\Http\Server\RequestHandlerInterface;
 
 /**
- * Applies CORS headers to all responses produced by the inner pipeline.
+ * Applies CORS headers to responses produced by the inner pipeline.
  *
  * By default no `Access-Control-Allow-Origin` header is set, which effectively
  * blocks cross-origin browser requests (secure-by-default). Configure
  * `$allowedOrigins` with a concrete list, or `['*']` to allow any origin.
  *
- * Headers already set by inner middleware are preserved — this middleware only
- * adds defaults when they are absent.
+ * `Access-Control-Allow-Methods` and `Access-Control-Allow-Headers` are emitted
+ * only on preflight responses (`OPTIONS` with an `Access-Control-Request-Method`
+ * header), per the CORS specification. Headers already set by inner middleware
+ * are preserved — this middleware only adds defaults when they are absent.
  *
  * @author Volodymyr Panivko <sveneld300@gmail.com>
  */
@@ -38,10 +41,11 @@ final class CorsMiddleware implements MiddlewareInterface
     private readonly ?string $exposedHeadersHeader;
 
     /**
-     * @param list<string> $allowedOrigins Origins permitted for cross-origin requests. Empty disables `Access-Control-Allow-Origin`. Use `['*']` to allow any origin.
-     * @param list<string> $allowedMethods Methods advertised via `Access-Control-Allow-Methods`
-     * @param list<string> $allowedHeaders Headers advertised via `Access-Control-Allow-Headers`
-     * @param list<string> $exposedHeaders Headers advertised via `Access-Control-Expose-Headers`
+     * @param list<string> $allowedOrigins   Origins permitted for cross-origin requests. Empty disables `Access-Control-Allow-Origin`. Use `['*']` to allow any origin.
+     * @param list<string> $allowedMethods   Methods advertised via `Access-Control-Allow-Methods` (preflight only)
+     * @param list<string> $allowedHeaders   Headers advertised via `Access-Control-Allow-Headers` (preflight only)
+     * @param list<string> $exposedHeaders   Headers advertised via `Access-Control-Expose-Headers`
+     * @param bool         $allowCredentials Whether to emit `Access-Control-Allow-Credentials: true`. Incompatible with `allowedOrigins: ['*']` — combining them throws.
      */
     public function __construct(
         private readonly array $allowedOrigins = [],
@@ -55,8 +59,14 @@ final class CorsMiddleware implements MiddlewareInterface
             StreamableHttpTransport::SESSION_HEADER,
         ],
         array $exposedHeaders = [StreamableHttpTransport::SESSION_HEADER],
+        private readonly bool $allowCredentials = false,
     ) {
         $this->isWildcard = \in_array('*', $allowedOrigins, true);
+
+        if ($this->isWildcard && $allowCredentials) {
+            throw new InvalidArgumentException('Access-Control-Allow-Origin: * is incompatible with Access-Control-Allow-Credentials: true. Configure an explicit allowedOrigins list when credentialed requests are required.');
+        }
+
         $this->varyOnOrigin = [] !== $allowedOrigins && !$this->isWildcard;
         $this->allowedMethodsHeader = implode(', ', $allowedMethods);
         $this->allowedHeadersHeader = implode(', ', $allowedHeaders);
@@ -72,16 +82,22 @@ final class CorsMiddleware implements MiddlewareInterface
             $response = $response->withHeader('Access-Control-Allow-Origin', $allowedOrigin);
         }
 
+        if ($this->allowCredentials && null !== $allowedOrigin && !$response->hasHeader('Access-Control-Allow-Credentials')) {
+            $response = $response->withHeader('Access-Control-Allow-Credentials', 'true');
+        }
+
         if ($this->varyOnOrigin) {
             $response = $this->ensureVaryOrigin($response);
         }
 
-        if (!$response->hasHeader('Access-Control-Allow-Methods')) {
-            $response = $response->withHeader('Access-Control-Allow-Methods', $this->allowedMethodsHeader);
-        }
+        if ($this->isPreflight($request)) {
+            if (!$response->hasHeader('Access-Control-Allow-Methods')) {
+                $response = $response->withHeader('Access-Control-Allow-Methods', $this->allowedMethodsHeader);
+            }
 
-        if (!$response->hasHeader('Access-Control-Allow-Headers')) {
-            $response = $response->withHeader('Access-Control-Allow-Headers', $this->allowedHeadersHeader);
+            if (!$response->hasHeader('Access-Control-Allow-Headers')) {
+                $response = $response->withHeader('Access-Control-Allow-Headers', $this->allowedHeadersHeader);
+            }
         }
 
         if (null !== $this->exposedHeadersHeader && !$response->hasHeader('Access-Control-Expose-Headers')) {
@@ -89,6 +105,12 @@ final class CorsMiddleware implements MiddlewareInterface
         }
 
         return $response;
+    }
+
+    private function isPreflight(ServerRequestInterface $request): bool
+    {
+        return 'OPTIONS' === $request->getMethod()
+            && $request->hasHeader('Access-Control-Request-Method');
     }
 
     private function resolveAllowedOrigin(string $origin): ?string
@@ -116,7 +138,12 @@ final class CorsMiddleware implements MiddlewareInterface
             return $response->withHeader('Vary', 'Origin');
         }
 
-        if ('*' === trim($current) || false !== stripos($current, 'origin')) {
+        if ('*' === trim($current)) {
+            return $response;
+        }
+
+        $tokens = array_map('strtolower', array_map('trim', explode(',', $current)));
+        if (\in_array('origin', $tokens, true)) {
             return $response;
         }
 
--- a/src/Server/Transport/Http/Middleware/CorsMiddleware.php
+++ b/src/Server/Transport/Http/Middleware/CorsMiddleware.php
@@ -1,0 +1,125 @@
+<?php
+
+/*
+ * This file is part of the official PHP MCP SDK.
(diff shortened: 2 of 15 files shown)


chr-hertel:
Thanks for that proposal @sveneld - my solution was not on point, true.
i feel it is quite brittle for people to accidentally opt out of security features - the moment i would add my own custom middleware and forget to add the default ones, all protection is gone, right?

sveneld:
Yes, if you pass your own middleware list, the default protection middleware will not be applied automatically.

That is why in the examples I explicitly show that custom middleware should be merged with the default middleware, rather than replacing it entirely:

```
$transport = new StreamableHttpTransport(
    $request,
    logger: $logger,
    middleware: [
        ...StreamableHttpTransport::defaultMiddleware(),
        new AuthMiddleware($responseFactory),
    ],
);
```

This keeps the default security protections enabled while still allowing custom middleware to be added.

src/Server/Transport/Http/Middleware/CorsMiddleware.php
```diff
@@ -0,0 +1,125 @@
...
+
+    private function ensureVaryOrigin(ResponseInterface $response): ResponseInterface
+    {
+        $current = $response->getHeaderLine('Vary');
+
+        if ('' === $current) {
+            return $response->withHeader('Vary', 'Origin');
+        }
+
+        if ('*' === trim($current) || false !== stripos($current, 'origin')) {
```
soyuka: this could have false positives, suggestion: 

```php
$tokens = array_map('strtolower', array_map('trim', explode(',', $current)));
if ('*' === trim($current) || in_array('origin', $tokens, true)) {
    return $response;
}
```

src/Server/Transport/Http/Middleware/CorsMiddleware.php
```diff
@@ -0,0 +1,125 @@
...
+            $response = $this->ensureVaryOrigin($response);
+        }
+
+        if (!$response->hasHeader('Access-Control-Allow-Methods')) {
+            $response = $response->withHeader('Access-Control-Allow-Methods', $this->allowedMethodsHeader);
+        }
+
+        if (!$response->hasHeader('Access-Control-Allow-Headers')) {
+            $response = $response->withHeader('Access-Control-Allow-Headers', $this->allowedHeadersHeader);
+        }
```
soyuka: These make sense only if in a preflight request, suggestion: 

```php
$isPreflight = 'OPTIONS' === $request->getMethod()
    && $request->hasHeader('Access-Control-Request-Method');

if ($isPreflight && !$response->hasHeader('Access-Control-Allow-Methods')) {
    $response = $response->withHeader('Access-Control-Allow-Methods', $this->allowedMethodsHeader);
}
if ($isPreflight && !$response->hasHeader('Access-Control-Allow-Headers')) {
    $response = $response->withHeader('Access-Control-Allow-Headers', $this->allowedHeadersHeader);
}
```

src/Server/Transport/Http/Middleware/CorsMiddleware.php
```diff
@@ -0,0 +1,125 @@
...
+        array $allowedMethods = ['GET', 'POST', 'DELETE', 'OPTIONS'],
+        array $allowedHeaders = [
+            'Accept',
+            'Authorization',
+            'Content-Type',
+            'Last-Event-ID',
+            StreamableHttpTransport::PROTOCOL_VERSION_HEADER,
+            StreamableHttpTransport::SESSION_HEADER,
+        ],
+        array $exposedHeaders = [StreamableHttpTransport::SESSION_HEADER],
```
soyuka, line 61:
```suggestion
        array $exposedHeaders = [StreamableHttpTransport::SESSION_HEADER],
        bool $allowCredentials = false,
```

src/Server/Transport/Http/Middleware/CorsMiddleware.php
```diff
@@ -0,0 +1,125 @@
...
+    ) {
+        $this->isWildcard = \in_array('*', $allowedOrigins, true);
+        $this->varyOnOrigin = [] !== $allowedOrigins && !$this->isWildcard;
+        $this->allowedMethodsHeader = implode(', ', $allowedMethods);
+        $this->allowedHeadersHeader = implode(', ', $allowedHeaders);
+        $this->exposedHeadersHeader = [] === $exposedHeaders ? null : implode(', ', $exposedHeaders);
```
soyuka, line 73:
```php
if ($this->isWildcard && $allowCredentials) {
    throw new InvalidArgumentException(
        'Access-Control-Allow-Origin: * is incompatible with Access-Control-Allow-Credentials: true.',
    );
}
```

src/Server/Transport/Http/Middleware/CorsMiddleware.php
```diff
@@ -0,0 +1,125 @@
...
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $response = $handler->handle($request);
+
+        $allowedOrigin = $this->resolveAllowedOrigin($request->getHeaderLine('Origin'));
+        if (null !== $allowedOrigin && !$response->hasHeader('Access-Control-Allow-Origin')) {
+            $response = $response->withHeader('Access-Control-Allow-Origin', $allowedOrigin);
+        }
+
```
soyuka, line 84: add `Access-Control-Allow-Credentials: true`

Something like: 
```php
  if ($this->allowCredentials && null !== $allowedOrigin
      && !$response->hasHeader('Access-Control-Allow-Credentials')) {
      $response = $response->withHeader('Access-Control-Allow-Credentials', 'true');
  }
```

+ tests

src/Server/Transport/Http/Middleware/DnsRebindingProtectionMiddleware.php
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private function isAllowedOrigin(string $origin): bool
+    {
+        $parsed = parse_url($origin);
+        if (false === $parsed || !isset($parsed['host'])) {
+            return false;
+        }
+
+        return \in_array(strtolower($parsed['host']), $this->allowedHosts, true);
```
soyuka:
```suggestion
        $host = parse_url($origin, \PHP_URL_HOST);
        if (!is_string($host) || '' === $host) {
            return false;
        }
        return \in_array(strtolower($host), $this->allowedHosts, true);
```

src/Server/Transport/Http/Middleware/DnsRebindingProtectionMiddleware.php
```diff
@@ -0,0 +1,108 @@
...
+    /** @var list<string> */
+    private readonly array $allowedHosts;
+
+    /**
+     * @param list<string>                  $allowedHosts    Hostnames (without port) that are permitted. Defaults to localhost variants.
+     * @param ResponseFactoryInterface|null $responseFactory PSR-17 response factory (auto-discovered if null)
+     * @param StreamFactoryInterface|null   $streamFactory   PSR-17 stream factory (auto-discovered if null)
+     */
+    public function __construct(
+        array $allowedHosts = ['localhost', '127.0.0.1', '[::1]', '::1'],
```
soyuka:
```suggestion
        array $allowedHosts = ['localhost', '127.0.0.1', '[::1]'],
```

Useless value will never get matched below

src/Server/Transport/Http/Middleware/DnsRebindingProtectionMiddleware.php
```diff
@@ -0,0 +1,108 @@
...
+            $hostname = explode(':', $host, 2)[0];
+        }
+
+        return \in_array(strtolower($hostname), $this->allowedHosts, true);
+    }
+
+    private function createForbiddenResponse(string $message): ResponseInterface
+    {
+        return JsonRpcErrorResponse::create($this->responseFactory, $this->streamFactory, 403, $message);
+    }
```
soyuka, line 110:
```suggestion
    private function createForbiddenResponse(string $message): ResponseInterface
  {
      return $this->responseFactory->createResponse(403)
          ->withHeader('Content-Type', 'text/plain')
          ->withBody($this->streamFactory->createStream($message));
  }
```

I feel `Error::forInvalidRequest($message)` is wrong here.

src/Server/Transport/Http/JsonRpcErrorResponse.php
```diff
@@ -0,0 +1,42 @@
...
+        int $statusCode,
+        string $message,
+    ): ResponseInterface {
+        $body = json_encode(Error::forInvalidRequest($message), \JSON_THROW_ON_ERROR);
+
+        return $responseFactory
+            ->createResponse($statusCode)
+            ->withHeader('Content-Type', 'application/json')
+            ->withBody($streamFactory->createStream($body));
+    }
```
soyuka, line 40:
```suggestion
  public static function create(
      ResponseFactoryInterface $responseFactory,
      StreamFactoryInterface $streamFactory,
      int $statusCode,
      Error $error,
  ): ResponseInterface {
      $body = json_encode($error, \JSON_THROW_ON_ERROR);
      return $responseFactory->createResponse($statusCode)
          ->withHeader('Content-Type', 'application/json')
          ->withBody($streamFactory->createStream($body));
  }
```

src/Server/Transport/Http/Middleware/ProtocolVersionMiddleware.php
```diff
@@ -0,0 +1,97 @@
...
+                implode(', ', $this->supportedVersions),
+            )
+            : \sprintf(
+                'Unsupported %s header value: %s. Supported versions: %s.',
+                StreamableHttpTransport::PROTOCOL_VERSION_HEADER,
+                $headerValue,
+                implode(', ', $this->supportedVersions),
+            );
+
+        return JsonRpcErrorResponse::create($this->responseFactory, $this->streamFactory, 400, $message);
```
soyuka:
```suggestion
        return JsonRpcErrorResponse::create($this->responseFactory, $this->streamFactory, 400, Error::forInvalidParams($message));
```

soyuka (changes requested):

sveneld:
Thanks for the review, @soyuka — all inline comments addressed in the latest commit:

- **`CorsMiddleware`**: Methods/Headers emitted only on preflight; `Vary` check is token-based now; added `allowCredentials` flag (emits the header, throws when combined with wildcard).
- **`DnsRebindingProtectionMiddleware`**: plain-text 403 body; `parse_url(..., PHP_URL_HOST)`; dropped unreachable `'::1'` from the default allowlist.
- **`ProtocolVersionMiddleware`** + **`JsonRpcErrorResponse`**: helper now takes a `JsonRpc\Error`; the middleware passes `Error::forInvalidParams($message)`.

@chr-hertel — on the "easy to opt out" concern: passing `middleware: []` now emits a `warning`-level log through the transport's `LoggerInterface`, naming the disabled protections and pointing at the spread-defaults pattern. A visible signal in logs feels like the right weight for an explicit opt-out.

src/Server/Transport/Http/Middleware/CorsMiddleware.php
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * @param list<string> $allowedOrigins   Origins permitted for cross-origin requests. Empty disables `Access-Control-Allow-Origin`. Use `['*']` to allow any origin.
+     * @param list<string> $allowedMethods   Methods advertised via `Access-Control-Allow-Methods` (preflight only)
+     * @param list<string> $allowedHeaders   Headers advertised via `Access-Control-Allow-Headers` (preflight only)
+     * @param list<string> $exposedHeaders   Headers advertised via `Access-Control-Expose-Headers`
+     * @param bool         $allowCredentials Whether to emit `Access-Control-Allow-Credentials: true`. Incompatible with `allowedOrigins: ['*']` — combining them throws.
+     */
+    public function __construct(
+        private readonly array $allowedOrigins = [],
+        array $allowedMethods = ['GET', 'POST', 'DELETE', 'OPTIONS'],
```
soyuka:
```suggestion
        array $allowedMethods = ['GET', 'POST', 'DELETE'],
```

soyuka (dismissed):
Looks ok, we could add the `Access-Control-Max-Age` in another PR. Also wondering if warning isn't a bit excessive for the `middleware: []` option, defaults has middleware, explicit `[]` is because users really want this I'd be to remove the log (or use `info`).

chr-hertel:
i'm not that worried about that `middleware: []` case, but mostly about killing security features with `middleware: [new MyCustomMiddleware()]` - that's rather unintuitive

soyuka:
yeah but if you do that you're hopefully aware of the default middlewares

sveneld:
@chr-hertel — let me lay out how I framed the design here, so we can stay aligned on what we're actually trading off.

**Goals I tried to satisfy:**

- Make HTTP requests secure by default (CORS, DNS rebinding, protocol-version validation).
- Provide a simple, composable extension mechanism for the request pipeline.
- Let users override any individual protection with their own implementation.
- Validate the value the user passes in.

**Constraints I deliberately kept:**

- No \`instanceof\` checks against the middleware the user provides — that's exactly the auto-prepend / class-detect pattern we replaced from #260 / #277, and it doesn't scale past 2–3 defaults.
- No content-inspection iteration over the user's list to guess intent.
- No boolean flags toggling individual security features on/off — those age into a long signature of \`enableX/disableY/bypassZ\` knobs.
- Extension is plain list extension — no DSL, no named slot registry, no auxiliary \`bypassDefaults: true\`.

Under those constraints, the only lever the transport has is the list itself. That gives us a small, predictable surface:

- omitted / \`null\` → secure defaults from \`StreamableHttpTransport::defaultMiddleware()\`.
- explicit list → exactly what the user provided, verbatim, in their order.
- empty list (\`[]\`) → defaults dropped, **and a warning is logged** so the bypass is visible.

**On the remaining concern — \`middleware: [new MyAuth()]\` silently losing the defaults.**

I think this risk reads as bigger under pre-AI workflows than under how PHP / SDK integrations are written today. Configuring a constructor in 2026 is heavily AI-assisted: an IDE or coding agent reading the constructor PHPDoc + \`docs/transports.md\` will spread \`...StreamableHttpTransport::defaultMiddleware()\` for the user, or at minimum surface the omission as a suggestion. Combined with:

- the empty-array warning at runtime (catches the most degenerate case),
- the public \`defaultMiddleware()\` factory being the obvious thing to land on in autocompletion,
- and the doc example explicitly showing the spread pattern as the canonical way to add middleware,

I think we get adequate coverage without re-introducing the validation styles we agreed to drop.

soyuka (approved):

chr-hertel (approved):
Let's bring this in - thanks @sveneld!

PR #314, opened 2026-05-19
[Server] Add PropertyDescriberInterface extension point for SchemaGenerator

peter-si:
## Summary

Adds a `PropertyDescriberInterface` extension point to `SchemaGenerator` so callers can teach it how to render specific class types (`DateTime`, `Uuid`, domain value objects) as targeted JSON Schema fragments rather than the generic `{type: "object"}` fallback.

Describers are consulted, in order, for class-typed parameters before generic class inspection. The first non-null result wins; parameter-level metadata (description, default, nullable) is layered onto the described schema without overwriting fields the describer already set.

Ships two default describers:

- `Mcp\Capability\Discovery\PropertyDescriber\DateTimePropertyDescriber` — any `\DateTimeInterface` implementation → `{type: "string", format: "date-time"}`
- `Mcp\Capability\Discovery\PropertyDescriber\UuidPropertyDescriber` — `Symfony\Component\Uid\Uuid` and subclasses → `{type: "string", format: "uuid"}`

## Why

If an MCP tool method takes e.g. `\DateTimeInterface \$until`, the generator currently falls back to `{type: "object"}`, which tells the LLM nothing about the expected shape and forces ad-hoc workarounds in every tool. The describer chain gives a clean extension point for the long tail of value-object types every project has, without needing to subclass or fork \`SchemaGenerator\`.

Downstream we extend the chain further with app-specific types (Money, PhoneNumber, ...), but the two shipped here are general enough that they feel like they belong upstream.

## Backwards compatibility

- New constructor parameter defaults to an empty iterable.
- Behavior is unchanged for callers that don't pass any describers.
- No public API removed or modified.

## Tests

- New \`DateTimePropertyDescriberTest\` and \`UuidPropertyDescriberTest\` cover per-describer happy paths and pass-through.
- New integration tests in \`SchemaGeneratorTest\` cover: fallback when no describer claims the type, describer override of generic inference, docblock-description layering, nullable + default handling, first-non-null-wins ordering, and that describers don't intercept unrelated class types.
- Full \`unit\` suite: 690 tests, 2215 assertions, all green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Capability/Discovery/PropertyDescriberInterface.php
+++ b/src/Capability/Discovery/PropertyDescriberInterface.php
@@ -14,18 +14,26 @@ namespace Mcp\Capability\Discovery;
 /**
  * Translates a PHP class type into a JSON Schema fragment.
  *
- * The {@see SchemaGenerator} consults registered describers, in order, before
- * falling back to generic class inspection. The first describer that returns
- * a non-null schema wins. Implementations let callers teach the generator
- * about value-object types (DateTime, Uuid, etc.) whose JSON Schema
+ * A describer declares the class (or base class/interface) it handles via
+ * {@see self::supportedClass()}. The {@see SchemaGenerator} matches a
+ * parameter's concrete class against that type — directly or through its
+ * parents and interfaces — and, when several describers are registered,
+ * consults them in priority order. Implementations let callers teach the
+ * generator about value-object types (DateTime, Uuid, etc.) whose JSON Schema
  * representation is more specific than a generic `{type: "object"}`.
  */
 interface PropertyDescriberInterface
 {
     /**
-     * @param class-string $className
+     * The class or interface this describer handles. Parameters whose type is
+     * the class itself or any subtype of it are dispatched to this describer.
      *
-     * @return array<string, mixed>|null Schema fragment, or null to pass to the next describer
+     * @return class-string
      */
-    public function describe(string $className): ?array;
+    public static function supportedClass(): string;
+
+    /**
+     * @return array<string, mixed> Schema fragment for the supported type
+     */
+    public function describe(): array;
 }
--- a/src/Capability/Discovery/PropertyDescriberInterface.php
+++ b/src/Capability/Discovery/PropertyDescriberInterface.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of the official PHP MCP SDK.
+ *
+ * A collaboration between Symfony and the PHP Foundation.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Mcp\Capability\Discovery;
+
+/**
+ * Translates a PHP class type into a JSON Schema fragment.
+ *
+ * The {@see SchemaGenerator} consults registered describers, in order, before
+ * falling back to generic class inspection. The first describer that returns
+ * a non-null schema wins. Implementations let callers teach the generator
+ * about value-object types (DateTime, Uuid, etc.) whose JSON Schema
+ * representation is more specific than a generic `{type: "object"}`.
+ */
+interface PropertyDescriberInterface
+{
+    /**
+     * @param class-string $className
+     *
+     * @return array<string, mixed>|null Schema fragment, or null to pass to the next describer
+     */
+    public function describe(string $className): ?array;
+}
--- a/src/Capability/Discovery/PropertyDescriberInterface.php
+++ b/src/Capability/Discovery/PropertyDescriberInterface.php
@@ -15,23 +15,15 @@ namespace Mcp\Capability\Discovery;
  * Translates a PHP class type into a JSON Schema fragment.
  *
  * A describer declares the class (or base class/interface) it handles via
- * {@see self::supportedClass()}. The {@see SchemaGenerator} matches a
- * parameter's concrete class against that type — directly or through its
- * parents and interfaces — and, when several describers are registered,
- * consults them in priority order. Implementations let callers teach the
- * generator about value-object types (DateTime, Uuid, etc.) whose JSON Schema
- * representation is more specific than a generic `{type: "object"}`.
+ * {@see PropertyHandlerInterface::supportedClass()}. The {@see SchemaGenerator}
+ * matches a parameter's or return type's concrete class against that type —
+ * directly or through its parents and interfaces — and, when several describers
+ * are registered, consults them in priority order. Implementations let callers
+ * teach the generator about value-object types (DateTime, Uuid, etc.) whose JSON
+ * Schema representation is more specific than a generic `{type: "object"}`.
  */
-interface PropertyDescriberInterface
+interface PropertyDescriberInterface extends PropertyHandlerInterface
 {
-    /**
-     * The class or interface this describer handles. Parameters whose type is
-     * the class itself or any subtype of it are dispatched to this describer.
-     *
-     * @return class-string
-     */
-    public static function supportedClass(): string;
-
     /**
      * @return array<string, mixed> Schema fragment for the supported type
      */
--- a/src/Capability/Discovery/SchemaGenerator.php
+++ b/src/Capability/Discovery/SchemaGenerator.php
@@ -60,33 +60,18 @@ use phpDocumentor\Reflection\DocBlock\Tags\Param;
  */
 final class SchemaGenerator implements SchemaGeneratorInterface
 {
-    /**
-     * @var list<PropertyDescriberInterface> Consulted in registration order before generic class
-     *                                       inspection; the first matching describer wins
-     */
-    private readonly array $propertyDescribers;
+    private readonly PropertyHandlerResolver $handlerResolver;
 
     /**
-     * Memoizes describer resolution per concrete class. Holds the matching
-     * describer or `false` when none matched, so repeated parameter types —
-     * and the common "no describer" case — are resolved at most once.
-     *
-     * @var array<class-string, PropertyDescriberInterface|false>
-     */
-    private array $describerCache = [];
-
-    /**
-     * @param iterable<PropertyDescriberInterface> $propertyDescribers Consulted in order before
-     *                                                                 generic class inspection; the
-     *                                                                 first matching describer wins
+     * @param iterable<PropertyHandlerInterface> $propertyHandlers Consulted in registration order;
+     *                                                             the first whose supported class
+     *                                                             matches a type wins
      */
     public function __construct(
         private readonly DocBlockParser $docBlockParser,
-        iterable $propertyDescribers = [],
+        iterable $propertyHandlers = [],
     ) {
-        $this->propertyDescribers = \is_array($propertyDescribers)
-            ? array_values($propertyDescribers)
-            : iterator_to_array($propertyDescribers, false);
+        $this->handlerResolver = new PropertyHandlerResolver($propertyHandlers);
     }
 
     /**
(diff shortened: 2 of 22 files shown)


src/Capability/Discovery/PropertyDescriberInterface.php
```diff
@@ -0,0 +1,31 @@
...
+ * representation is more specific than a generic `{type: "object"}`.
+ */
+interface PropertyDescriberInterface
+{
+    /**
+     * @param class-string $className
+     *
+     * @return array<string, mixed>|null Schema fragment, or null to pass to the next describer
+     */
+    public function describe(string $className): ?array;
```
chr-hertel: since we're only relying on the `$className` and don't have any dynamic, we could be more explicit while designing this interface.

think of sth like `public static function supportedClass(): class-string`
=> no implicit nullable as encoded non-support
=> static index+lookup instead of iterating over all describers for a property

src/Capability/Discovery/SchemaGenerator.php
```diff
@@ -349,6 +364,65 @@ private function inferParameterTypes(array $paramInfo): array
...
+     */
+    private function describeClassType(array $paramInfo): ?array
+    {
+        $reflectionType = $paramInfo['reflection_type_object'];
+        if (!$reflectionType instanceof \ReflectionNamedType || $reflectionType->isBuiltin()) {
+            return null;
+        }
+
+        $className = $reflectionType->getName();
+        foreach ($this->propertyDescribers as $describer) {
```
chr-hertel: with multiple properties and a generator injected, we'd need to rewind here - if possible. but see comment above on the interface - maybe we don't need this here.

tests/Unit/Capability/Discovery/SchemaGeneratorFixture.php
```diff
@@ -438,6 +438,37 @@ public function withParameterNamedRequest(string $_request): void
...
+     */
+    public function dateTimeWithDescription(\DateTimeInterface $until): void
+    {
+    }
+
+    public function nullableDateTimeParam(?\DateTimeImmutable $finishedAt = null): void
+    {
+    }
+
+    public function uuidParam(\Symfony\Component\Uid\Uuid $bookingId): void
```
chr-hertel: please use an import here for the type

chr-hertel (changes requested):
Hi @peter-si,
thanks for this proposal - i like the general idea, but left some comments about the design.
also, i think it would be good to add docs and think about higher-level usage => how to add describers on the `Mcp\Server\Builder`.
Thanks already!

src/Capability/Discovery/PropertyDescriberInterface.php
peter-si: Changed the interface to `static supportedClass(): class-string` plus a non-nullable `describe(): array`, so support is declared explicitly rather than encoded as a `null` return. The generator matches a parameter's class against `supportedClass()` via `is_a()` (so `\DateTimeInterface` still covers `\DateTimeImmutable`, `Uuid` covers `UuidV4`, etc.) and memoizes the resolution per concrete class, so describers aren't re-scanned per property.

I kept an ordered scan + per-class cache rather than a flat exact-class index, because the shipped describers need to match subclasses/interfaces, which an exact-key map can't express.

src/Capability/Discovery/SchemaGenerator.php
peter-si: Resolved by the interface change plus materializing the describers once in the constructor (`iterator_to_array`), so an injected `Generator` is no longer exhausted across parameters — no rewind needed.

tests/Unit/Capability/Discovery/SchemaGeneratorFixture.php
peter-si: Done - added `use Symfony\Component\Uid\Uuid;`; the fixture parameter is now `Uuid $bookingId`.

peter-si:
Hi @chr-hertel 

thanks for the review. I've pushed_ cfad5b4 addressing the feedback:

- **Interface** - explicit `static supportedClass(): class-string` + non-nullable `describe(): array`; support is no longer an implicit `null`. Matching against `supportedClass()` (subtypes included) lives in `SchemaGenerator`, with the describers materialized once and resolution memoized per concrete class.
- **Builder** - added `addPropertyDescriber()` (opt-in; consulted in registration order, first match wins). Mutually exclusive with `setSchemaGenerator()`.
- **Docs** - new *Custom Type Describers* section under Schema Generation and Validation, plus a Method Reference entry in `server-builder.md`.
- Import nit in the fixture fixed.

Kept it simple for now - no priority system; happy to add one if you'd prefer it over registration order.

chr-hertel:
Thanks for the update - i wonder tho if this is usable already or rather incomplete - how would you use this feature?

peter-si:
When you say you think it's incomplete what specifically do you have in mind? What do you think is missing?

This is the ground work to make the schema generation extendable. In our Symfony project we had to create decorators, because there was no entry point to modify how the schema is generated. I took the inspiration from Symfony Serializer (with it's normalizer approach). Whether it is usable today - sure you just need to make your own describers for your own specific VOs

Of course we could ship more standard describers (https://github.com/symfony/serializer/tree/8.1/Normalizer), but that can maybe be another PR once this one is more adopted.

chr-hertel:
let's talk features, so you have a tool that uses `Uuid` or `DateTime` as argument or output schema or how do I understand? just want to follow your use case here :)

peter-si:
It's a bit related to what I'm suggesting here https://github.com/symfony/symfony/issues/64279. Our usecase is that we have a Symfony app. We already have a scaffolding (using Nelmio api doc) which describes our controllers into openapi spec using only strict typed objects as a source (no OA attributes needed). This way the docs never drift from the actual code

Now that we implemented the MCP protocol in our app I want to achieve the same thing. We have a code which defines the tools and uses all our already existing DTOs as input/output parameters. Similar to standard controllers. Something like this:
```php
#[McpTool(
    name: 'location_town_shop_list',
    description: 'List shops available in a town by slug.',
    annotations: new ToolAnnotations(
        title: 'Town Shop List',
        readOnlyHint: true,
        destructiveHint: false,
        idempotentHint: true,
        openWorldHint: false
    ),
 )]
#[McpToolAccess(roles: [Role::ROLE_INTERNAL])]
public function getTownShopList(string $slug): GetTownShopListV1Dto
{
    return $this->guard->run(fn (): GetTownShopListV1Dto => $this->locationService->getTownShopList($slug));
}
```

and yes the input parameter could be an UUID of a town (we like strict types), but also on the return side since we serialize/return whole DTOs we need to be able to give them proper format.

chr-hertel:
> and yes the input parameter could be an UUID of a town

this was also what i was thinking and I don't think this PR addresses this fully. yes, we can generate an UUID hint in the schema now, but we cannot upcast what is coming in from the client to a real Uuid instance to be handed to the tool as argument:

talking about:
```php
public function getTownShopList(Uuid $id)
```

peter-si:
yes, you are right. I thought it could be separate topic, but yes we can put it all together since it's touching the same concerns. There is also the output side concern. Because currently it's just a simple json_encode(). But doing so would mean a much bigger scope for this PR - new property-reflection engine (big, absent on input side too). Do you think we should address that as well?

peter-si:
Pushed an update that closes the gap you flagged — `getTownShopList(Uuid $id)` now receives a real `Uuid`, not the raw client string.

The describer extension point is now three small interfaces sharing `supportedClass()`, and a single class may implement any combination:

- `PropertyDescriberInterface` — `describe()`: type → JSON Schema (unchanged)
- `PropertyDenormalizerInterface` — `denormalize($value, $class)`: client input → instance, applied in `ReferenceHandler` for class-typed arguments
- `PropertyNormalizerInterface` — `normalize($value)`: instance → JSON, applied to a class-typed tool result before it's returned

The shipped `Uuid`/`DateTime` describers implement all three, and the `is_a` lookup that was inline in `SchemaGenerator` is now a shared `PropertyHandlerResolver` used by the schema, input and output paths.

The important part is that there are now real extension points where you can plug the Symfony components in: register one handler that delegates `denormalize`/`normalize` to `symfony/serializer` (and `describe` to your schema source) and your value objects and DTOs round-trip end-to-end, without subclassing the generator. Examples in the docs: [Custom Type Describers](https://github.com/peter-si/php-sdk/blob/feat/property-describer-extension-point/docs/mcp-elements.md#custom-type-describers).

Output: class-typed results are normalized to JSON via `PropertyNormalizerInterface`, and an `outputSchema` is inferred from the return type when the describer yields an `object` schema.

`addPropertyDescriber()` stays mutually exclusive with `setSchemaGenerator()` / `setReferenceHandler()` (use those if you'd rather configure the components yourself). Docs + tests updated; unit suite green.

chr-hertel:
Ouch, that's quite some change, but makes sense that this is quite some effort. But that's the only way the first step makes sense ... not sure tho, tbh, if we want that entire layer in here :thinking: 

need to give that some thoughs - and some feedback of others, ofc

soyuka:
Really not a huge fan, its not the responsibility of the PHP SDK to generate JSON Schemas, use any of the available library out there and use in in the tool no?

peter-si:
Thanks @soyuka, fair pushback - let me try to separate a few things, because I think there are three different concerns getting bundled together:

**1. JSON Schema isn't a side-feature of MCP, it _is_ the tool contract.**

The MCP spec requires `inputSchema` (and optionally `outputSchema`) as JSON Schema for every tool - that's the only thing the model sees when deciding whether/how to call a tool. If the PHP SDK punts schema generation entirely, every consumer has to wire up their own pipeline before they can ship a single tool. At that point what is the SDK doing for them on the protocol side? I'd argue schema generation belongs in the SDK by definition, the same way request/response handling does.

**2. "Just use a library in the tool" works in theory, but the SDK doesn't actually let you.**

Today the only escape hatch is `setSchemaGenerator()` - replace the whole thing. That makes the bundled `SchemaGenerator` more or less decorative: any non-trivial app has to throw it out and ship its own. That's strictly worse than not shipping one, because it sets an expectation the SDK doesn't follow through on.

What this PR is going for is the opposite - keep the default generator useful for the common case, and give downstream libraries a small, well-defined seam to teach it about new types. A Symfony integration can plug `symfony/serializer` + `symfony/type-info` in through three small interfaces without forking anything. An ApiPlatform integration could do the same with its own metadata. That's the point of the extension hooks.

**3. The size of this PR is itself the signal.**

I take the "this is a lot of change" point - I'd flip it, though. The fact that _this_ is what it takes to make `SchemaGenerator` cleanly extensible is the signal that the SDK isn't really ready for downstream integration yet. If adding a seam for the most basic surface (class → schema, value ↔ instance) needs a change this size, how is a third-party library supposed to integrate cleanly today without subclassing or replacing the generator?

I'm explicitly not trying to grow the SDK's surface or pull maintenance burden upstream - the goal is the opposite. Make it cheap enough to extend that integrations (Symfony, ApiPlatform, whatever else) can live _downstream_ and bring their own serializer / type-info / validator. The two default describers shipped here (`DateTime`, `Uuid`) are just enough to make the common case work out of the box - the real value is the seam, not those classes.

**Open question for you both**

You know the SDK's direction and roadmap better than I do. If the intent is "the SDK shouldn't ship a schema generator at all, downstream owns it end-to-end" - that's a coherent position, and I'd want to know that before investing more here. If the intent is "we ship a generator and want it to be extensible," I think something shaped like this PR is roughly the minimum.

Either of those is fine by me - I just need a steer on which way you'd like to go. Happy to rework, split, or shelve depending on the answer.

soyuka:
hi @peter-si fair, and honestly I should have been clearer earlier — full disclosure, I wasnt around when the SchemaGenerator landed in the SDK in the first place, so I have no strong attachment to keeping it as-is. happy to revisit the whole thing.

I went and looked at what the other SDKs actually do before answering you, because you raised a real question. roughly:

- typescript SDK ships zero reflection (cant, types are erased). user hands the SDK a Standard Schema validator (zod / valibot / arktype) or raw JSON Schema via `fromJsonSchema()`. SDK is glue.
- java SDK ships zero generation. `Tool` is built from a `Map<String, Object>` the user writes. no `@Tool` annotation in core. spring-ai is a *separate project* that layers annotations + `victools/jsonschema-generator` on top — and even it uses an external lib, doesnt roll its own.
- go SDK does reflect by default, but delegates to `github.com/google/jsonschema-go`. ~100 lines of glue in the SDK, the rest is the external lib. extension point (custom type → schema) is the libs own `TypeSchemas` map, not something the SDK invents.
- python SDK reflects via pydantic. zero MCP-specific schema API. custom types → pydantics own hooks.

so the pattern across every SDK that ships generation is: pick an external library, dont reimplement. the bare java SDK and the typescript SDK go further and ship nothing at all.

PHP today is the outlier — 842 lines of hand-rolled SchemaGenerator + 125 of DocBlockParser, no external dep, nothing to plug into. thats why this PR exists, and I get why. but I dont think adding a describer interface to the walker you propose is the right move. it locks us into maintaining the walker, and we still wouldnt have what symfony/serializer or symfony/type-info already give for free.

my preference: SDK takes a plain array, thats it. `#[McpTool(inputSchema: [...])]` already works, we keep that. for users who want generation, mcp-bundle can ship one (probably on top of symfony/type-info, mirroring what you propose in symfony/symfony#64279), or pick whatever lib they like — api-platform/json-schema, cognesy/instructor-php, neuron-core, whatever. not the SDKs job, same as it isnt the java SDKs job.

concretely that means: drop `SchemaGenerator` + `DocBlockParser` + `setSchemaGenerator()` from core (not stable yet, no BC promise), move them out (mcp-bundle owns the symfony-flavored generator). the protocol-level surface shrinks, the framework integration grows in the right place. closer to where java + spring-ai drew the line.

chr-hertel:
Closing here for now - an alternative approach could be to just use the extension point the SDK provides in userland. If that's not sufficient, let's discuss what's missing.
Thanks anyhow @peter-si!

PR #372, opened 2026-06-01
[Schema][Server] Add Skills extension (io.modelcontextprotocol/skills) support

wachterjohannes:
## Skills extension (`io.modelcontextprotocol/skills`, SEP-2640)

Implements the Skills extension. [SEP-2640](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2640) reached Final status and merged into the specification on 2026-09-13. A skill is a directory of files, at minimum a `SKILL.md` with YAML frontmatter, served through the existing Resources primitive under `skill://<skill-path>/<file-path>`. `Builder::addSkillsFromDirectory()` walks a directory and registers every skill's files as resources, and enables the extension's two mandatory methods: `skills/list`, which enumerates the skills a server serves with a complete digest-and-size manifest per skill, and `skills/get`, which returns a single skill's entry by URI, including skills absent from the listing.

The manifest lets a host build its registry, present a skill for approval, and verify every later read without a follow-up call. `skills/list` results carry the base protocol's `ttlMs`/`cacheScope` caching attributes on protocol version `2026-07-28` and later, per the base list-caching mechanism (SEP-2549). Nested skills, a `SKILL.md` inside another skill's directory, are supported per the spec: a nested skill gets its own top-level entry under its own frontmatter, and still appears in the enclosing skill's resource manifest.

This PR was reworked from an earlier version built against a draft of the spec that used a discovery-index resource and packed archive distribution. The Core Maintainers removed both during review in favor of a complete resources manifest on every skill entry (see the SEP's Deferred Features appendix for the rationale); this implementation matches the Final spec text as merged.

Not implemented: the optional `resources/directory/read` method and its `directoryRead` capability flag, which this PR does not declare. Adding it later is backward compatible.

Adds `symfony/yaml` for frontmatter parsing. Docs are in `docs/advanced/extensions.md`, and a runnable example server is at `examples/server/skills/`. phpstan, php-cs-fixer, and the full test suite (1822 tests) are green, including an inspector stdio snapshot test and a dispatch-level test that sends real JSON-RPC requests through the server and asserts the caching envelope.


wachterjohannes:
For cross-reference, the downstream pieces are now open:

- **Consumer:** symfony/ai#2132 — ships the first skills built on this PR in the Symfony "Mate" MCP server, colocated with the tools they orchestrate (two skills, one with a `references/` subdirectory exercising the multi-file model).
- **Working-group findings:** modelcontextprotocol/experimental-ext-skills#95 — records this as the first PHP-ecosystem implementation of SEP-2640, including the spec-text frictions surfaced here (undefined `_meta` namespace, the resource-`name` charset constraint vs. the "name SHOULD equal frontmatter name" recommendation, and the empty-payload `{}`-vs-`[]` serialization fix).

Together: this PR is the transport, symfony/ai#2132 is the content, modelcontextprotocol/experimental-ext-skills#95 is the write-up.

wachterjohannes:
Updated to track the recent SEP-2640 redesign and to round out the index entries:

- **Index format realigned** (`2fb9722`): entries now mirror the `SKILL.md` frontmatter verbatim and carry a SHA-256 `digest`; dropped the `$schema` field and the `type`/`SkillType` enum (the `mcp-resource-template` type was removed from the index in the redesign).
- **Archives** (`634f81e`): skills can now also be served as one-shot packed resources via `addSkillsFromDirectory(..., archiveFormats: ['application/gzip'])` — a deterministic `.tar.gz` per skill, listed under the entry's `archives`, with the index `digest` matching the served bytes.

All green: phpstan (level 6), php-cs-fixer, 812 unit tests, inspector snapshot. Description updated accordingly.

Remaining SEP surface, as a follow-up: the optional `resources/directory/read` method (+ `directoryRead` capability), and ZIP archives.

chr-hertel:
Even without mate needing this, this could become interesting - keeping this open! I guess it needs a heavier rebase nowadays tho, since the extension part changed with v0.8.0

wachterjohannes:
@chr-hertel yeah i also think that this is something important for the future! will rebase that in the next days

PR #375, opened 2026-06-05
[Server] Allow pre-built instance handlers in element registration

### LINKED ISSUE #369: [Server] Passing tool handler using instance method throws "Invalid array handler"
**Describe the bug**

**To Reproduce**
```<?php

use Mcp\Server;

require_once __DIR__ . '/vendor/autoload.php';

class Foo
{
    public function bar()
    {
    }
}

$foo = new Foo();
$mcpServerInstance = Server::builder()
    ->addTool(
        handler: [$foo, 'bar'],
        name:    'foo',
    )
    ->build();
    
```
After running this I get

```
PHP Fatal error:  Uncaught Mcp\Exception\InvalidArgumentException: Invalid array handler format. Expected [ClassName::class, 'methodName']. in /app/vendor/mcp/sdk/src/Capability/Discovery/HandlerResolver.php:45
```
**Expected behavior**
Per documentation this should be possible

Image

valeriudev:
Registering an element with a **pre-built object instance** as the handler throws at build time:

```php
$foo = new Foo();
Server::builder()->addTool(handler: [$foo, 'bar'], name: 'foo')->build();
// Mcp\Exception\InvalidArgumentException: Invalid array handler format.
// Expected [ClassName::class, 'methodName'].  (HandlerResolver.php)
```

This is the only place that rejects an instance: the `Handler` type alias already declares `array{0: object|string, 1: string}`, `getHandlerDescription()` already formats object handlers, and the runtime invoker (`ReferenceHandler`) already dispatches them via its `is_callable` branch. `HandlerResolver::resolve()`'s guard simply required a string in slot 0, out of sync with that contract.

**Fix:** accept an object in slot 0 and normalize it to its class name for reflection — the same `is_object($h[0]) ? $h[0]::class : $h[0]` idiom `getHandlerDescription()` already uses. The guard can't be replaced by `is_callable()`, since `is_callable([Class::class, 'instanceMethod'])` is `false` and that lazy-instantiate form must keep working.

This unblocks handler objects with constructor dependencies that the container-less `new $className()` fallback cannot build.

**Tests:** added `testResolvesValidInstanceArrayHandler`; full unit suite, CS, and PHPStan are green.

Fixes #369

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to `mcp/sdk` will be documented in this file.
 
-0.6.1
+0.7.0
 -----
 
 * Allow `[$instance, 'methodName']` as an element handler in `Builder::addTool()`, `addResource()`, `addResourceTemplate()`, and `addPrompt()`. Unblocks handlers with constructor dependencies that the container-less `new $className()` fallback cannot build.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to `mcp/sdk` will be documented in this file.
 
-0.7.0
+0.6.1
 -----
 
 * Allow registering an element handler as a pre-built object instance (`[$instance, 'methodName']`) via `Builder::addTool()`, `addResource()`, `addResourceTemplate()`, and `addPrompt()`. `HandlerResolver` previously rejected instances with "Invalid array handler format" even though the `Handler` type already permitted them — this unblocks handler objects with constructor dependencies that the container-less fallback cannot build.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to `mcp/sdk` will be documented in this file.
 
+0.7.0
+-----
+
+* Allow registering an element handler as a pre-built object instance (`[$instance, 'methodName']`) via `Builder::addTool()`, `addResource()`, `addResourceTemplate()`, and `addPrompt()`. `HandlerResolver` previously rejected instances with "Invalid array handler format" even though the `Handler` type already permitted them — this unblocks handler objects with constructor dependencies that the container-less fallback cannot build.
+
 0.6.0
 -----
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 0.6.1
 -----
 
-* Allow registering an element handler as a pre-built object instance (`[$instance, 'methodName']`) via `Builder::addTool()`, `addResource()`, `addResourceTemplate()`, and `addPrompt()`. `HandlerResolver` previously rejected instances with "Invalid array handler format" even though the `Handler` type already permitted them — this unblocks handler objects with constructor dependencies that the container-less fallback cannot build.
+* Allow `[$instance, 'methodName']` as an element handler in `Builder::addTool()`, `addResource()`, `addResourceTemplate()`, and `addPrompt()`. Unblocks handlers with constructor dependencies that the container-less `new $className()` fallback cannot build.
 
 0.6.0
 -----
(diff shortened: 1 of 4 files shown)


src/Capability/Discovery/HandlerResolver.php
```diff
@@ -41,10 +45,11 @@ public static function resolve(\Closure|array|string $handler): \ReflectionMetho
         }
 
         if (\is_array($handler)) {
-            if (2 !== \count($handler) || !isset($handler[0]) || !isset($handler[1]) || !\is_string($handler[0]) || !\is_string($handler[1])) {
-                throw new InvalidArgumentException('Invalid array handler format. Expected [ClassName::class, \'methodName\'].');
+            if (2 !== \count($handler) || !isset($handler[0]) || !isset($handler[1]) || !(\is_string($handler[0]) || \is_object($handler[0])) || !\is_string($handler[1])) {
```
soyuka: `is_object($handler[0])` now accepts any object, including `Closure` and anonymous-class instances. A handler like `[$closure, 'method']` would pass this guard, then `$closure::class` yields `Closure`, and the user gets `Handler method "method" not found in class "Closure"` instead of the format error — which is misleading.

Two options:
- Explicitly reject `Closure` here (`!$handler[0] instanceof \Closure`) so the format-error path catches it.
- Or accept it knowingly and document that `[$closure, 'method']` is undefined behavior.

Also: the 4-clause negated condition is hard to scan. Consider extracting `$isHandler = \is_string($handler[0]) || \is_object($handler[0])` or splitting into early returns.

src/Capability/Discovery/HandlerResolver.php
soyuka, line 1: Pre-existing nit surfaced while reviewing this PR: at L78, the error message reads `'Handler method "%s::%s" must be abstract.'` but the branch *rejects* abstract methods — it should read `must not be abstract`. Worth a follow-up commit while this file is open.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to `mcp/sdk` will be documented in this file.
 
+0.6.1
+-----
+
+* Allow registering an element handler as a pre-built object instance (`[$instance, 'methodName']`) via `Builder::addTool()`, `addResource()`, `addResourceTemplate()`, and `addPrompt()`. `HandlerResolver` previously rejected instances with "Invalid array handler format" even though the `Handler` type already permitted them — this unblocks handler objects with constructor dependencies that the container-less fallback cannot build.
```
soyuka:
```suggestion
* Allow `[$instance, 'methodName']` as an element handler in `Builder::addTool()`, `addResource()`, `addResourceTemplate()`, and `addPrompt()`. Unblocks handlers with constructor dependencies that the container-less `new $className()` fallback cannot build.
```

src/Capability/Discovery/HandlerResolver.php
valeriudev: Went with option A — explicitly rejecting `\Closure` so `[$closure, 'method']` falls through to the format error instead of the misleading "method not found in class Closure". Also extracted the guard into a `$hasValidTarget` local to drop the 4-clause negation, and added a regression test (`testThrowsForClosureInArrayHandler`).

src/Capability/Discovery/HandlerResolver.php
valeriudev: Fixed. `must be abstract` → `must not be abstract`, with the assertion in testThrowsForAbstractMethodHandler updated to match

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to `mcp/sdk` will be documented in this file.
 
+0.6.1
```
chr-hertel: let's go with `0.7.0` here - it's rather a feature to me:

```suggestion
0.7.0
```

chr-hertel:
Thanks @valeriudev for working on this - we need this :) 👍

valeriudev:
@chr-hertel updated the version

Glad to contribute, I look forward to understanding the project more deeply and making bigger contributions over time.

chr-hertel (approved):

chr-hertel:
Bringing in an example for this with #390

PR #378, opened 2026-06-06
[Server] Always emit `items` for array tool parameter schemas

### LINKED ISSUE #151: Failed to validate tool : Error: tool parameters array type must have items.
**Describe the bug**
I have created a simple mysql db interaction tool and it have three parameters. but vscode throw the error when try to use the tool.

error:

> Failed to validate tool mcp_php-mysql-mcp_execute_select_query: Error: tool parameters array type must have items. Please open a Github issue for the MCP server or extension which provides this tool

 

**To Reproduce**
1.  setup the initial template
2.  create the initial class
3.  add this method to class

```php
<?php

  /**
     * Execute a custom SELECT query on the database.
     * 
     * @param string $query The string SELECT query to execute (Required, MUST be a SELECT statement, start from SELECT)
     * @param array $params Optional parameters for prepared statement (default: [])
     * @param int $limit Optional limit on number of rows to return (default: 100)
     * @return array Query results or error message
     */
    #[McpTool(name: 'execute_select_query')]
    public function executeSelectQuery(string $query, array $params = [], int $limit = 100): array
    {
        
        

        if(!is_array($params)) {
            return [
                'error' => 'Params must be an array.'
            ];
        }
        
        try {
            // Validate that query is a SELECT statement
            $trimmedQuery = trim($query);
            if (!preg_match('/^SELECT\s+/i', $trimmedQuery)) {
                $errorMsg = 'Only SELECT queries are allowed. Query must start with SELECT.';
                $this->logError('execute_select_query', $errorMsg, ['query' => $query]);
                return [
                    'error' => $errorMsg
                ];
            }
            
            // Check for dangerous keywords that could be used for data modification
            $dangerousPatterns = [
                '/\bINTO\s+OUTFILE\b/i',
                '/\bINTO\s+DUMPFILE\b/i',
                '/\bLOAD_FILE\b/i',
                '/\bINSERT\b/i',
                '/\bUPDATE\b/i',
                '/\bDELETE\b/i',
                '/\bDROP\b/i',
                '/\bCREATE\b/i',
                '/\bALTER\b/i',
                '/\bTRUNCATE\b/i',
                '/\bREPLACE\b/i',
                '/\bEXEC\b/i',
                '/\bEXECUTE\b/i'
            ];
            
            foreach ($dangerousPatterns as $pattern) {
                if (preg_match($pattern, $trimmedQuery)) {
                    $errorMsg = 'Query contains forbidden keywords. Only safe SELECT queries are allowed.';
                    $this->logError('execute_select_query', $errorMsg, ['query' => $query]);
                    return [
                        'error' => $errorMsg
                    ];
                }
            }
            
            // Apply limit if not already present in query
            if (!preg_match('/\bLIMIT\s+\d+/i', $trimmedQuery)) {
                $trimmedQuery = rtrim($trimmedQuery, ';') . " LIMIT {$limit}";
            }
            
            // Execute query
            if (!empty($params)) {
                $stmt = $this->connection->prepare($trimmedQuery);
                $stmt->execute($params);
            } else {
                $stmt = $this->connection->query($trimmedQuery);
            }
            
            $results = $stmt->fetchAll();
            
            return [
                'success' => true,
                'rows' => $results,
                'count' => count($results),
            ];
        } catch (PDOException $e) {
            $errorMsg = 'Query execution failed: ' . $e->getMessage();
            
            return [
                'error' => $errorMsg,
                'query' => $query
            ];
        }
    }

?>
```

4. add mcp server to vscode and start server
5.  use the tool in agent mode

**Expected behavior**
validated method that usable by AI agent

**Additional context**
If i use single parameter it worked.

valeriudev:
## Problem

VS Code / Copilot and other strict MCP clients reject tools whose input schema declares an `array` parameter without an `items` keyword:

> Failed to validate tool `<name>`: Error: tool parameters array type must have items. Please open a Github issue for the MCP server or extension which provides this tool

Any handler with an untyped `array` parameter (e.g. `array $params = []`, as in the report) generated `{"type": "array"}` with no `items`, so the tool was unusable in those clients.

## Root cause

`SchemaGenerator` only set `items` when it could infer an element type:

- Untyped arrays (`array`, `array<string, mixed>` maps, `mixed`) and untyped variadics emitted `type: array` with no `items` at all.
- A nullable typed array (`string[]|null`) lost its element type entirely, because the `|null` suffix wasn't stripped before the `T[]` inference ran — so it too fell back to no `items`.

## Fix

`src/Capability/Discovery/SchemaGenerator.php`, two parts:

1. **`ensureArrayItems()`** runs *after* all schema merges (and on the variadic path). When the schema is array-typed and still has no `items`, it defaults to the empty schema — `items: {}` (a `\stdClass`, so it serializes to `{}` and not `[]`), which matches any element. Running post-merge means a `#[Schema]` attribute that reshapes a parameter (e.g. to `object`) is respected and the invariant is only enforced on what is genuinely still an array.
2. **`inferArrayItemsType()`** strips a top-level nullable union (`string[]|null`, `null|int[]`) before inference, so nullable typed arrays recover their element type. Inner unions like `array<int|string>` are left untouched (the strip is anchored to the start/end only); array nullability is still handled separately via `allows_null`.

No public API symbol changes — this corrects derived schema output to be spec-compliant, so it is a bugfix, not a `[BC Break]`.

## Verification

- New unit tests `testRecoversItemsTypeForNullableTypedArrays` and `testUntypedVariadicStillDeclaresItems`; tightened the existing array-case assertions; new fixtures `nullableTypedArrays` / `untypedVariadic`.
- `make tests` (unit + inspector), `make phpstan`, `make cs`, `composer validate --strict`, and `make docs` all green.
- End-to-end: the committed `HttpComplexToolSchemaTest` snapshot now shows the nullable `string[]|null` attendee-emails parameter correctly carrying `items: {type: string}`.

## Note for merge

A parallel branch of mine also opens a `0.6.1` CHANGELOG section. If it lands first, I'll rebase and move this bullet under the existing `0.6.1` heading rather than re-adding it.

Fixes #151

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 -----
 
 * Allow `[$instance, 'methodName']` as an element handler in `Builder::addTool()`, `addResource()`, `addResourceTemplate()`, and `addPrompt()`. Unblocks handlers with constructor dependencies that the container-less `new $className()` fallback cannot build.
+* Always emit an `items` schema for array tool parameters: untyped arrays get `items: {}` and nullable typed arrays (e.g. `string[]|null`) keep their element type. Fixes strict clients rejecting tools with "array type must have items" (#151).
 
 0.6.0
 -----
--- a/src/Capability/Discovery/SchemaGenerator.php
+++ b/src/Capability/Discovery/SchemaGenerator.php
@@ -223,7 +223,7 @@ final class SchemaGenerator implements SchemaGeneratorInterface
     private function buildParameterSchema(array $paramInfo, ?array $methodLevelParamSchema): array
     {
         if ($paramInfo['is_variadic']) {
-            return $this->buildVariadicParameterSchema($paramInfo);
+            return $this->ensureArrayItems($this->buildVariadicParameterSchema($paramInfo));
         }
 
         $inferredSchema = $this->buildInferredParameterSchema($paramInfo);
@@ -240,7 +240,34 @@ final class SchemaGenerator implements SchemaGeneratorInterface
             $mergedSchema = array_merge($mergedSchema, $parameterLevelSchema);
         }
 
-        return $mergedSchema;
+        // Run after all merges so that when a Schema attribute reshapes the parameter
+        // (e.g. to `object`), the array `items` invariant is only enforced on what is
+        // genuinely still an array.
+        return $this->ensureArrayItems($mergedSchema);
+    }
+
+    /**
+     * Guarantees an array-typed schema always declares `items`.
+     *
+     * `items` is optional in JSON Schema, but some strict clients reject an array schema
+     * without it. When no element type could be inferred, default to the empty schema `{}`
+     * (matches anything) — represented as `new \stdClass()` so it serializes to `{}` rather
+     * than `[]`.
+     *
+     * @param array<string, mixed> $schema
+     *
+     * @return array<string, mixed>
+     */
+    private function ensureArrayItems(array $schema): array
+    {
+        $type = $schema['type'] ?? null;
+        $isArray = 'array' === $type || (\is_array($type) && \in_array('array', $type, true));
+
+        if ($isArray && !isset($schema['items'])) {
+            $schema['items'] = new \stdClass();
+        }
+
+        return $schema;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to `mcp/sdk` will be documented in this file.
 
+0.6.1
```
```suggestion
0.7.0
```

chr-hertel (dismissed):
Thanks @valeriudev - will merge it with CHANGELOG referencing v0.7.0 instead of v0.6.1

chr-hertel (approved):

PR #379, opened 2026-06-06
[Schema][Server] Preserve request id in JSON-RPC error responses instead of fabricating id:""

### LINKED ISSUE #333: [Server] stdio parse errors use `id:""` instead of preserving the request id or returning null
**Describe the bug**

When the stdio server receives JSON-RPC input that exceeds PHP's JSON nesting limit, it returns a JSON-RPC parse error with `id:""`.

That response is hard for clients to correlate with the original request. In the repro below, the malformed request had a numeric id such as `900512`, but the server response used an empty-string id like:

```json
{"jsonrpc":"2.0","id":"","error":{"code":-32700,"message":"Maximum stack depth exceeded"}}
```

The server process stayed alive and later `ping` requests succeeded, so this is not a crash report.

The issue is specifically `id:""`: the original request had a numeric id, but the server returned an empty-string id that was not present in the request. This makes the error hard for JSON-RPC clients to correlate.

MCP 2025-11-25 defines `JSONRPCErrorResponse.id` as optional, and MCP `RequestId` is `string | number`, so omitting `id` is valid if the original request id truly cannot be recovered after a parse failure. `id:null` is the JSON-RPC convention for an unknown id in a parse-error response, but `null` is not itself an MCP `RequestId`.

Preserving the original id when recoverable, or otherwise omitting the id / documenting the unknown-id behavior, would avoid fabricating an unrelated id.
- Environment:
  - Reproduced with stable release `v0.5.0` (`fb2c8c2ee4ab2791239c5f534bb07bfb7589d4e8`)
  - Also reproduced with a `main` snapshot from 2026-05-26 (`293a5880e44b36c2120b114d175b6ddc4d7d1b06`)
  - Transport: stdio HTTP server

**To Reproduce**

1. Start a PHP SDK MCP server over stdio. I used the discovery calculator example.
2. Complete a normal `initialize` request followed by `notifications/initialized`.
3. Send a JSON-RPC request with a numeric id and a deeply nested object value. The example below uses a nested `initialize.params.capabilities` object at depth 512.
4. Observe the JSON-RPC error response id.

**Expected behavior**

For JSON-RPC errors, the response `id` should match the request id when the request id is known or recoverable.

If the id cannot be determined after a parse failure, MCP 2025-11-25 allows `JSONRPCErrorResponse.id` to be omitted. JSON-RPC also conventionally uses `id:null` for an unknown id in parse-error responses, although `null` is not an MCP `RequestId`.

What I would not expect is `id:""`, because that introduces a request id value that was not present in the request and is difficult for JSON-RPC clients to correlate.**Logs**

```text
-> init payload depth=512:
   {"jsonrpc":"2.0","id":900512,"method":"initialize","params":{"protocolVersion":"2025-11-25","capabilities":{... deeply nested object ...}}}

<- {"jsonrpc":"2.0","id":"","error":{"code":-32700,"message":"Maximum stack depth exceeded"}}

-> health after deep payload 512:
   {"jsonrpc":"2.0","id":7,"method":"ping","params":{}}

<- {"jsonrpc":"2.0","id":7,"result":{}}
```

The same `id:""` parse-error behavior appeared for deeper inputs as well, including depths 1024, 2048, 4096, 8192, and 10000. The server process stayed alive.

valeriudev:
## The bug

When the server receives input it cannot turn into a valid message, the JSON-RPC error response carries a fabricated empty-string `id` that was never in the request:

```json
{"jsonrpc":"2.0","id":"","error":{"code":-32700,...}}
```

The reporter sent an `initialize` request with a real numeric id (`900512`) nested past PHP's `json_decode()` depth limit; the server replied with `id:""`. An empty string is not a valid JSON-RPC / MCP `RequestId`, so this breaks client correlation.

## Root cause

Two paths both collapsed to `""`:

- `Error::forParseError()` and `Error::forInvalidRequest()` defaulted `$id` to `''`, and `Error::$id` was typed `string|int` (no `null`). On an unrecoverable parse failure (`\JsonException` in `Protocol::processInput()`) the factory was called with no id, so `""` reached the wire.
- On an invalid-but-parseable message, `MessageFactory` threw `InvalidInputMessageException` without carrying the decoded `id`, so `Protocol::handleInvalidMessage()` also fell back to `""` — even though the real id was recoverable from the decoded payload.

## The fix

- Unrecoverable parse errors (`-32700`) now return `id: null`, per JSON-RPC 2.0 (id is `null` when it cannot be determined).
- Invalid-but-parseable requests (`-32600`) now preserve the original request `id`.
- `Error::$id` / `Error::getId()` and the `$id` parameter of `Error::forParseError()` / `Error::forInvalidRequest()` are widened to `string|int|null`. `jsonSerialize()` is unchanged — it now emits `"id":null` instead of `"id":""`.
- The recoverable id is threaded from `MessageFactory` through `InvalidInputMessageException` via new `getRequestId()`/`setRequestId()`. The id is only attached when the decoded value is a valid `string|int` (mirroring `Request`/`Error::fromArray` validation); `id:0` is preserved (type-checked, not truthiness), while `null`/`true`/array/missing stay `null`.

No `[BC Break]`: the type is widened, not narrowed, and the only behavioral change is replacing an invalid fabricated id with a spec-compliant one.

## Tests

Two regression tests in `tests/Unit/Server/ProtocolTest.php`:

- `testParseErrorDoesNotFabricateEmptyStringId` — feeds JSON nested 600 deep so `json_decode()` throws, asserts `error.code === -32700`, `id !== ''`, and that `id` is null or omitted.
- `testInvalidMessagePreservesRecoverableId` — feeds valid JSON with id `42` but no `method`/`result`/`error`, asserts `error.code === -32600` and `id === 42`.

Both were red for the right reason (got `""`) before the fix and are green after. Full unit suite (766 tests), phpstan (level 6), and php-cs-fixer all pass.

Fixes #333


src/Schema/JsonRpc/Error.php
soyuka, line 1: 🟡 **`for*()` factories left behind (L89–L111).** `forMethodNotFound`, `forInvalidParams`, `forInternalError`, `forServerError`, and `forResourceNotFound` still default `$id = ''`. Any caller that forgets to pass the id reintroduces the exact bug this PR fixes. Suggest widening them to `string|int|null` with `null` default for consistency, even if no current call site hits that path.

🟡 **`fromArray()` no longer round-trips its own output (L60–L65).** The ctor and `jsonSerialize()` now accept/emit `id: null`, but `fromArray()` still throws on null (`isset($data['id'])` is false for null, then the type guard rejects it). Encoding a parse-error response and decoding it back via `fromArray()` blows up. Either allow `null` here or document the asymmetry.

🔵 **`@phpstan-type ErrorData` is stale (L19–L25).** Still declares `id: string|int`; widen to `string|int|null` to match the ctor / `getId()` / `jsonSerialize()` shape.

src/JsonRpc/MessageFactory.php
```diff
@@ -111,6 +111,9 @@ public function create(string $input): array
             try {
                 $messages[] = $this->createMessage($message);
             } catch (InvalidInputMessageException $e) {
+                if (\is_array($message) && isset($message['id']) && (\is_string($message['id']) || \is_int($message['id']))) {
```
soyuka, line 153: ❓ JSON-RPC 2.0 allows `id: null` in requests. Here `isset()` drops it, so a null-id invalid request falls back to the `null` default in `forInvalidRequest()` — same outcome, fine. Intentional? If so, worth a one-line comment so the next reader doesn't "fix" it by switching to `array_key_exists`.

tests/Unit/Server/ProtocolTest.php
```diff
@@ -297,6 +297,45 @@ public function testInvalidJsonReturnsParseError(): void
...
+        $this->assertSame(Error::PARSE_ERROR, $decoded['error']['code']);
+        // The original id is genuinely unrecoverable after a parse failure: the response
+        // must be null or omit the id, never a fabricated empty string.
+        $this->assertNotSame('', $decoded['id'] ?? null, 'Parse error must not fabricate an empty-string id');
+        // isset() is false for both an absent key and a null value, so this asserts the id is
+        // null or omitted (never a fabricated empty string).
+        $this->assertFalse(
+            isset($decoded['id']),
+            'Unrecoverable parse error id should be null or omitted',
+        );
```
soyuka: 🔵 nit: assertion intent and actual behavior don't quite line up. `jsonSerialize()` always emits the `id` key, so the response is *always* `id: null`, never omitted — the "or omitted" branch can never trigger. Tighter:

```php
$this->assertArrayHasKey('id', $decoded);
$this->assertNull($decoded['id'], 'Unrecoverable parse error must emit id: null');
```

That also catches a future regression where someone drops the key entirely (which would break clients that strictly validate the JSON-RPC shape).

tests/Unit/Server/ProtocolTest.php
```diff
@@ -349,6 +388,55 @@ public function testInvalidMessageStructureReturnsError(): void
...
+            $sessionId
+        );
+
+        $outgoing = $protocol->consumeOutgoingMessages($sessionId);
+        $this->assertCount(1, $outgoing);
+
+        $message = json_decode($outgoing[0]['message'], true);
+        $this->assertArrayHasKey('error', $message);
+        $this->assertEquals(Error::INVALID_REQUEST, $message['error']['code']);
+        $this->assertSame(42, $message['id'], 'Invalid-but-parseable message must preserve its recoverable id, not return ""');
```
soyuka: 🔵 nit: PR description specifically calls out that `id: 0` and string ids are preserved (type-checked, not truthiness), but only int `42` is exercised here. Worth adding a data provider or two extra cases for `id: 0` (the truthiness trap) and a string id like `"req-1"` — both are cheap to assert and lock down the two explicit promises.

soyuka (commented):
Automatic review worth a look.

src/JsonRpc/MessageFactory.php
valeriudev: Intentional - isset + the type guard. We only recover valid `string|int` ids; a null or non-scalar id stays at the exception's null default. `array_key_exists` would converge to the same via the type guard. Leaving a comment in the code.

chr-hertel (changes requested):
Correct me if I'm wrong, but I think you missed `Mcp\Client\State\ClientState::storeResponse` here.

Sorry about PHPStan not catching that - somehow expected with this kind of refactoring tbh.

valeriudev:
> Correct me if I'm wrong, but I think you missed `Mcp\Client\State\ClientState::storeResponse` here.
> 
> Sorry about PHPStan not catching that - somehow expected with this kind of refactoring tbh.

You're right, fixed - thanks. A null-id error can't be matched to any pending request, so handleResponse() now logs and drops it before it reaches the state. Added a test in tests/Unit/Client/ProtocolTest.php

chr-hertel:
Hey @valeriudev — thanks for tracking this down, the empty-string `id` bug was a real problem.

There's some overlap with #442, which also fixes #333 and just merged. The two took different stances on what an unrecoverable id should serialize as: #442 omits the `id` key, this PR emitted `id: null`. I checked the actual MCP JSON schema (`RequestId` in `modelcontextprotocol/modelcontextprotocol`) across every revision including `2026-07-28`, and `RequestId` never permits `null` — it's always `type: ["string", "integer"]`. As of `2025-11-25`, `id` was made *optional* on `JSONRPCErrorResponse` rather than nullable, which is exactly the omit approach.

Since #442 already landed with that semantics, I rebased this branch onto current `main`:
- Dropped the `Error.php`/`fromArray()`/`jsonSerialize()` changes — already done by #442, so they were pure conflicts.
- Kept the part that's genuinely additive and not covered by #442: recovering the real request id for an *invalid-but-parseable* message (valid JSON, bad/missing `method`) instead of answering it id-less. That's `InvalidInputMessageException::getRequestId()`/`setRequestId()` threaded from `MessageFactory` through `Protocol::handleInvalidMessage()`.
- Updated the parse-error test assertion to require the id be *omitted*, not `null`.
- Kept your authorship on every commit.

Full unit suite, phpstan, and cs-fixer all pass on the rebased branch, force-pushed to this PR. Let me know if you want any of that reworked — happy to iterate.

chr-hertel (approved):

Issue #319, opened 2026-05-19
[Client] Implement OAuth 2.0 Authorization Code flow with PKCE (RFC 6749 + RFC 7636)


soyuka:
## Context

Primary user-facing flow. Required for any interactive MCP client to obtain tokens after PRM/AS discovery.

## Scope

- `Mcp\Client\Auth\Grant\AuthorizationCodeGrant`:
  - Generate PKCE `code_verifier` + `code_challenge` (S256).
  - Build authorize URL with `client_id`, `redirect_uri`, `response_type=code`, `code_challenge`, `code_challenge_method`, `scope`, `state`, `resource` (audience-binding RFC 8707).
  - Pluggable user-agent dispatcher: callback hook so library users can open a browser / present URL in CLI.
  - Local loopback redirect listener (default) or custom redirect handler.
  - Exchange code → tokens at `token_endpoint`; persist via `TokenStorageInterface`.
- Verify `state` round-trip; reject mismatched.

## Conformance scenarios unblocked

`auth/basic-cimd` and prerequisite for all scope/refresh/cross-app scenarios.

## Dependencies

Blocked by: #315, #316, #317, #318. Pairs with #319 (token endpoint auth methods).

## Acceptance

- Unit tests for PKCE generation + state validation.
- Conformance: `auth/basic-cimd` passes.

cc @soyuka

chr-hertel:
Cross-reference: this work is a foundation for two 2026-07-28 SEPs:

- [SEP-2468](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2468) — clients must validate the `iss` parameter in the authorization response (RFC 9207) to mitigate mix-up attacks. Tracked separately as #360.
- [SEP-2207](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2207) — when the AS is OIDC-flavored, clients should request the `offline_access` scope to obtain refresh tokens. Tracked separately as #363.

See umbrella #338 for the full MCP Spec [2026-07-28](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/) authorization milestone.

PR #221, opened 2026-01-12
[Server] Oauth2 based on middleware

sveneld:
## Motivation and Context
Implementation of OAuth based on middlewares from https://github.com/modelcontextprotocol/php-sdk/pull/218 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [X] I have added appropriate error handling
- [X] I have added or updated documentation as needed

## Additional context

Fix #214

PR #382, opened 2026-06-15
[Server] scope OAuth to resource server, not authorization server

soyuka:
## Summary

We keep getting PRs that push the SDK toward becoming a full OAuth 2.1 **authorization server** — minting its own tokens, registering clients, running login/consent (i.e. being an IdP). The most explicit is #373 (~3,400 lines / 50+ files), but the direction has crept in incrementally.

This PR draws the line authoritatively and removes the surface signal that invites it. **No behavior change** — ADR + docblock + docs only.

## Why a docs/ADR change instead of code

A scope doc alone won't stop these PRs. The real driver is the *API surface*: the SDK already exposes the authorization-server endpoints in proxy form (`OAuthProxyMiddleware` answers `/authorize`, `/token`, `/.well-known/oauth-authorization-server`; DCR landed in #269). From a contributor's seat the SDK looks ~60% of an AS, so #373 reads as "finishing the job."

So this does two things:
1. **Authoritative ruling (ADR 0001)** — a single link to cite when declining AS/IdP PRs.
2. **Reframe the proxy surface** — make `OAuthProxyMiddleware`'s delegation-only intent unmistakable, so the leap to "issue tokens here" is no longer implied.

## OAuth's three roles — the SDK implements one and a half

| Role | Status | Scope |
|------|--------|-------|
| Resource Server (validate tokens, PRM, `WWW-Authenticate`) | shipped | in scope |
| Delegation/proxy to an upstream AS | shipped (`OAuthProxyMiddleware`) | in scope — delegation only |
| Authorization Server / IdP (issue tokens, consent) | absent | **out of scope** (this is #373) |

## Changes

- **`adr/0001-oauth-authorization-server-out-of-scope.md`** — the ruling, rationale (security liability, RFC footprint, mature libs/IdPs already exist), boundary statement, and "what to do instead" (front with Keycloak/Auth0/Entra/Okta, or run `league/oauth2-server` in your own app behind the existing proxy + validator seams). References #373 as the example and #221/#269 as the surface origin.
- **`adr/README.md`** — new ADR index.
- **`src/Server/Transport/Http/Middleware/OAuthProxyMiddleware.php`** — class docblock only: leads with delegation intent, explicit non-goals (never mints/signs/stores/rotates tokens).
- **`docs/authorization.md`** — new "Scope: what this SDK does and does not do" section (three-role table + ADR link); proxy entry reframed as delegation.

## Out of scope

- Client-side OAuth (token acquisition, #319) — a separate genuine gap, not addressed here.
- No class renames / signature / behavior changes.

`php -l` passes on the edited middleware. No markdown linter is configured in the repo.

### DIFF
--- a/adr/0001-oauth-authorization-server-out-of-scope.md
+++ b/adr/0001-oauth-authorization-server-out-of-scope.md
@@ -1,0 +1,97 @@
+# 0001 — The MCP server is an OAuth Resource Server, not an Authorization Server
+
+- Status: Accepted
+- Date: 2026-06-15
+
+## Context
+
+OAuth 2.1 defines three distinct roles:
+
+| Role | What it does | Status in this SDK | Scope |
+|------|--------------|--------------------|-------|
+| Resource Server | Validates incoming bearer tokens, serves Protected Resource Metadata (RFC 9728), emits `WWW-Authenticate` | Shipped (`AuthorizationMiddleware`, `JwtTokenValidator`, `ProtectedResourceMetadata`) | **IN scope** |
+| Delegation / proxy to an upstream AS | Forwards `/authorize` and `/token` to your existing IdP | Shipped (`OAuthProxyMiddleware`) | **IN scope — delegation ONLY** |
+| Authorization Server / Identity Provider (IdP) | Mints its own tokens, registers clients, runs login and consent | Absent | **OUT of scope** |
+
+The SDK repeatedly receives pull requests that move it toward becoming a full OAuth 2.1
+**authorization server** — asking the MCP server to mint its own access and refresh tokens,
+register clients, and run login and consent flows, that is, to become an Identity Provider.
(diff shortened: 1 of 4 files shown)


wachterjohannes:
👍 Agree with this ADR for the SDK — Resource Server (+ delegation), not an IdP. I'll close #373 against it.

The "what to do instead" path — run `league/oauth2-server` behind the validator/proxy seams — is right, but it shouldn't be left to every app to wire up from scratch. That reusable layer belongs in `symfony/mcp-bundle` (tracking in symfony/ai#2134): the **Protected Resource Metadata builder** as a framework-agnostic service, plus an **opt-in, league-backed AS** — including the **RFC 7591 DCR** piece league lacks but MCP clients require. "Use league yourself" leaves every Symfony app re-solving DCR + consent wiring.

chr-hertel (approved):
Thanks @soyuka - let's try to adopt that ADR practice - i like :)

PR #383, opened 2026-06-15
[Server] Harden JSON-RPC input parsing against malformed input and resource exhaustion

soyuka:
## Problem

`MessageFactory::create()` decided single-vs-batch by inspecting the raw first byte (`$input[0] === '{'`). That misclassifies valid-but-non-object JSON:

- a scalar payload (`5`, `"x"`, `true`) falls through and the subsequent `foreach` runs over a non-array — a PHP warning and a silently-empty result (or a `TypeError` on a batch element);
- a leading-whitespace object (`"  {...}"`) is treated as a batch and the loop iterates the object's *values* as if they were messages.

There was also no bound on batch size, so a single small request could expand into arbitrarily many operations (amplification), and no bound on the HTTP request body size.

## Changes

- **Type-based message detection.** Determine single message vs batch from the decoded JSON type (object → single, list array → batch) instead of the first byte. Scalars and empty payloads are returned as `InvalidInputMessageException` entries — consistent with the existing per-message error contract — rather than crashing or no-oping.
- **Batch size cap.** New `maxBatchSize` (default 100); oversized batches are rejected before any message is constructed.
- **Non-object batch elements** are reported as `InvalidInputMessageException` instead of reaching a `TypeError`.
- **Bounded HTTP body.** `StreamableHttpTransport` gains `maxBodyBytes` (default 4 MiB) and returns `413` for oversized bodies. Unknown-size/chunked bodies are read incrementally and stopped at the cap so they cannot exhaust memory.

Both limits are constructor-configurable. `MessageFactory::create()`'s contract is unchanged: it still throws only `\JsonException` for unparseable JSON; every other invalidity is surfaced as an `InvalidInputMessageException` in the returned array, so callers need no changes.

## Tests

- `MessageFactoryTest`: scalar/string rejected, empty batch rejected, batch element must be object, leading-whitespace object parsed as a single message, batch size over/within cap, non-positive `maxBatchSize` throws.
- `StreamableHttpTransportTest`: body over cap → 413, body within cap not rejected, non-positive `maxBodyBytes` throws.

Full suite green (874 tests), PHPStan and PHP-CS-Fixer clean. Backward compatible — new constructor args are optional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ All notable changes to `mcp/sdk` will be documented in this file.
 * [BC Break] `ResourceDefinition::__construct()` signature changed — `$title` parameter added between `$name` and `$description`. Callers using positional arguments must switch to named arguments.
 * [BC Break] `ResourceTemplate::__construct()` signature changed — `$title` parameter added between `$name` and `$description`. Callers using positional arguments must switch to named arguments.
 * [BC Break] `McpResource` and `McpResourceTemplate` attribute signatures changed — `$title` parameter added between `$name` and `$description`. Callers using positional arguments must switch to named arguments.
+* Harden JSON-RPC input parsing: single-message vs batch is now decided from the decoded JSON type (object → single, list array → batch) instead of the raw first byte. Scalars, empty payloads, and non-object batch elements are surfaced as `InvalidInputMessageException` entries instead of triggering warnings or a `TypeError`.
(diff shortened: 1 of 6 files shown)


chr-hertel (dismissed):
docs and CHANGELOG would be nice, other then that LGTM, thanks! 👍

chr-hertel (approved):

PR #384, opened 2026-06-19
[Server] Reject malformed MCP session ID headers

### LINKED ISSUE #380: [Server] Malformed or duplicate `Mcp-Session-Id` headers cause HTTP 500 responses
**Describe the bug**

The PHP SDK Streamable HTTP server returns HTTP 500 for malformed or duplicate Mcp-Session-Id inputs. Since these inputs come from client-supplied request metadata, I would expect them to be rejected as client/transport errors, rather than surfacing as internal server errors.

Tested with:

- PHP SDK: reproduced with stable release `v0.6.0` (`433c84b58af346dd32f15f9909679e96a46ebe23`)
- Transport: Streamable HTTP

**To Reproduce**

1. Start a PHP SDK Streamable HTTP server.
2. Complete a normal initialization flow.
3. Send a valid JSON-RPC request, such as `tools/list`, with a malformed `Mcp-Session-Id`.
4. Repeat with a wrong-format session id and with duplicate `Mcp-Session-Id` headers.

Representative request:

```text
POST /mcp HTTP/1.1
Content-Type: application/json
Accept: application/json, text/event-stream
Mcp-Session-Id: {"not":"a-token"}

{"jsonrpc":"2.0","id":"session-id-repro","method":"tools/list","params":{}}
```

For the duplicate case, send two `Mcp-Session-Id` headers on the same request.

**Expected behavior**

The MCP specification explicitly uses `400 Bad Request` for missing required session ids and `404 Not Found` for requests containing a terminated session id. It does not appear to explicitly define the status code for malformed or duplicate `Mcp-Session-Id` headers, but those inputs are client-supplied request metadata, so I would expect a `4xx` rejection rather than `500 Internal Server Error`.

**Logs**

| Case | HTTP status |
|---|---|
| malformed `Mcp-Session-Id` (e.g. JSON object) | 500 |
| wrong-format `Mcp-Session-Id` (e.g. non-ASCII) | 500 |
| duplicate `Mcp-Session-Id` headers | 500 |
| stale but well-formed `Mcp-Session-Id` | 404 |
| omitted `Mcp-Session-Id` | 400 |

The last two rows show that some session-id validation paths already return 4xx responses. The 500 responses appear to be specific to malformed or duplicate session-id header inputs.

jstar0:
## Summary

Fixes #380.

Malformed or repeated `Mcp-Session-Id` request headers currently reach UUID parsing inside `StreamableHttpTransport`, so client-supplied bad session metadata can surface as an internal server error.

## Changes

- Reject repeated `Mcp-Session-Id` headers with `400 Bad Request`.
- Reject malformed single `Mcp-Session-Id` values with `400 Bad Request`.
- Keep valid, missing, and existing stale-session handling on the existing paths.
- Add regression coverage for malformed and duplicate session header inputs.

## Verification

Focused regressions were also run for the malformed and duplicate session-id header cases.

```bash
make unit-tests
make phpstan
git diff --check
```

## Risk

This is limited to invalid Streamable HTTP session-id request metadata. Valid session ids continue to be parsed as UUIDs, missing session ids still flow through the existing request handling, and well-formed but stale session ids still return the existing not-found response.

### DIFF
--- a/src/Server/Transport/StreamableHttpTransport.php
+++ b/src/Server/Transport/StreamableHttpTransport.php
@@ -24,7 +24,6 @@ use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\StreamFactoryInterface;
 use Psr\Http\Server\MiddlewareInterface;
 use Psr\Log\LoggerInterface;
-use Symfony\Component\Uid\Exception\InvalidArgumentException as UidInvalidArgumentException;
 use Symfony\Component\Uid\Uuid;
 
 /**
@@ -304,7 +303,8 @@ class StreamableHttpTransport extends BaseTransport
 
         try {
             $this->sessionId = $sessionIdString ? Uuid::fromString($sessionIdString) : null;
-        } catch (UidInvalidArgumentException) {
+            // Symfony UID 5.4/6.4 throw the global parent; newer versions throw a namespaced subclass.
+        } catch (\InvalidArgumentException) {
             return $this->createErrorResponse(Error::forInvalidRequest(self::SESSION_HEADER.' header must be a valid UUID.'), 400);
         }
 
--- a/src/Server/Transport/StreamableHttpTransport.php
+++ b/src/Server/Transport/StreamableHttpTransport.php
@@ -294,8 +294,18 @@ class StreamableHttpTransport extends BaseTransport
     private function handleRequest(ServerRequestInterface $request): ResponseInterface
     {
         $this->request = $request;
-        $sessionIdString = $request->getHeaderLine(self::SESSION_HEADER);
-        $this->sessionId = $sessionIdString ? Uuid::fromString($sessionIdString) : null;
+        $sessionIdHeaders = $request->getHeader(self::SESSION_HEADER);
+        if (\count($sessionIdHeaders) > 1) {
+            return $this->createErrorResponse(Error::forInvalidRequest(self::SESSION_HEADER.' header must not be repeated.'), 400);
+        }
+
+        $sessionIdString = $sessionIdHeaders[0] ?? '';
+
+        try {
+            $this->sessionId = $sessionIdString ? Uuid::fromString($sessionIdString) : null;
+        } catch (\InvalidArgumentException) {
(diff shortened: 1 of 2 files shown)


src/Server/Transport/StreamableHttpTransport.php
```diff
@@ -294,8 +294,18 @@ private static function normalizeMiddleware(iterable $middleware): array
...
+        $sessionIdHeaders = $request->getHeader(self::SESSION_HEADER);
+        if (\count($sessionIdHeaders) > 1) {
+            return $this->createErrorResponse(Error::forInvalidRequest(self::SESSION_HEADER.' header must not be repeated.'), 400);
+        }
+
+        $sessionIdString = $sessionIdHeaders[0] ?? '';
+
+        try {
+            $this->sessionId = $sessionIdString ? Uuid::fromString($sessionIdString) : null;
+        } catch (\InvalidArgumentException) {
```
chr-hertel: Let's be specific please and catch the `Symfony\Component\Uid\Exception\InvalidArgumentException` instead, that's what we're expecting.

chr-hertel (changes requested):
Thanks for working on the patch @jstar0! Just one minor thing and we can merge it in 👍

src/Server/Transport/StreamableHttpTransport.php
jstar0: Narrowed the catch to Symfony\Component\Uid\Exception\InvalidArgumentException in d07b7ba. The focused session-header regressions, full transport test, full unit suite, style check, and PHPStan all pass.

jstar0: The supported dependency matrix showed that the namespaced exception is not available on Symfony UID 5.4/6.4: PHPStan reported the caught class as missing, and those jobs still throw the global InvalidArgumentException from Uuid::fromString(). I restored the common parent catch and documented the version split in bc1d3bc. Newer UID exceptions subclass that parent, and the catch remains scoped directly around Uuid::fromString(), so this covers all supported Symfony versions without broadening the surrounding request handling.

chr-hertel: oh wow, wasn't aware of https://github.com/symfony/symfony/pull/60226

chr-hertel (approved):

PR #386, opened 2026-06-24
[Server] Add outgoing (elicitation, sampling) request and client response events

omouren:
Introduce `ServerRequestEvent` and `ClientResponseEvent` to provide full observability over the server's interaction with the client on handshake-era connections. This includes requests sent by the server (e.g. `elicitation/create`, `sampling/createMessage`) and the client's replies to those requests.

Additionally, update `ResponseEvent` and `ErrorEvent` to be dispatched when a suspended Fiber completes. This ensures that deferred responses from long-running, asynchronous operations are also observable via the existing event mechanisms.

On the modern protocol revision (`2026-07-28`), there is no server-initiated JSON-RPC over HTTP, so `ServerRequestEvent` / `ClientResponseEvent` are not dispatched there, they would misrepresent the wire. Instead, `StatelessProtocol` dispatches the same lifecycle events as the handshake path:

- **`RequestEvent`** : including multi-round-trip retries, when the client sends `inputResponses`, they are already lifted onto the session as `InputContext`
- **`ResponseEvent`** : an elicitation ask is a successful result (`InputRequiredResult`, `resultType: input_required`)
- **`ErrorEvent`** : handler-related failures

Documentation is updated in `docs/advanced/events.md`.

## Motivation and Context

We wanted to be able to log every step of our MCP usage, including elicitation, across both protocol eras.

## How Has This Been Tested?

- Unit tests added/extended in `ProtocolTest`, `StatelessProtocolTest`, and `MessageFactoryTest`
- Full unit suite passes locally (1596 tests)
- Manual testing on handshake-era connections (elicitation events) + MRTR

## Breaking Changes

- **[BC Break]** `TransportInterface` gains `setFiberTerminationHandler(callable $handler): void`. Custom transport implementations must implement it, SDK transports (`StdioTransport`, `StreamableHttpTransport`) already do via `ManagesTransportCallbacks`.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist


src/JsonRpc/MessageFactory.php
```diff
@@ -118,6 +118,16 @@ public function create(string $input): array
         return $messages;
     }
 
+    /**
+     * @param array<string, mixed> $data
+     *
+     * @throws InvalidInputMessageException
+     */
+    public function createFromArray(array $data): MessageInterface
```
chr-hertel: i agree that for a additional public method `createFromArray` is better than `createMessage`, but i think we can skip the private method, inline here and adopt the new one in `create`

src/Event/OutgoingRequestEvent.php
```diff
@@ -0,0 +1,50 @@
...
+
+use Mcp\Schema\JsonRpc\Request;
+use Mcp\Server\Session\SessionInterface;
+
+/**
+ * Event dispatched when the server sends a request to the client (e.g. elicitation/create, sampling/create).
+ *
+ * @author Olivier Mouren <mouren.olivier@gmail.com>
+ */
+final class OutgoingRequestEvent
```
chr-hertel: would prefer to name it `ServerRequestEvent`

src/Server/Protocol.php
```diff
@@ -540,6 +553,57 @@ public function handleFiberYield(mixed $yieldedValue, ?Uuid $sessionId): void
...
+
+    private function resolveFiberParentRequest(mixed $data): ?Request
+    {
+        if (!\is_array($data)) {
+            return null;
+        }
+
+        try {
+            $message = $this->messageFactory->createFromArray($data);
+        } catch (\Throwable) {
```
chr-hertel: why catch `Throwable` - there's only `InvalidInputMessageException` thrown. always prefer to be defensive and only catch what we want to compensate for.

in this case i'm not even sure we need to catch sth at all - what's your reasoning here?

src/Server/Protocol.php
```diff
@@ -540,6 +553,57 @@ public function handleFiberYield(mixed $yieldedValue, ?Uuid $sessionId): void
...
+            $responseEvent = $this->dispatchEvent(new ResponseEvent($finalResult, $parentRequest, $session));
+            $finalResult = $responseEvent->getResponse();
+        } else {
+            $errorEvent = $this->dispatchEvent(new ErrorEvent($finalResult, $parentRequest, $session, null));
+            $finalResult = $errorEvent->getError();
+        }
+
+        $session->save();
+
+        return $finalResult;
```
chr-hertel: this part feels a bit clunky since we're having that kind of early return on `!$parentRequest` but there's nothing else unconditional in the method left to execute - i would just merge that:


```suggestion
        if (null !== $parentRequest) {
            if ($finalResult instanceof Response) {
                $responseEvent = $this->dispatchEvent(new ResponseEvent($finalResult, $parentRequest, $session));
                $finalResult = $responseEvent->getResponse();
            } else {
                $errorEvent = $this->dispatchEvent(new ErrorEvent($finalResult, $parentRequest, $session, null));
                $finalResult = $errorEvent->getError();
            }
        }

        $session->save();

        return $finalResult;
```

chr-hertel (changes requested):
Thanks @omouren - looks like a valuable addition to me, left some comments and didn't test yet.

omouren:
Thanks for the review @chr-hertel 👍

chr-hertel:
Hey @omouren this needs a rebase - and more importantly, please double-check the handling in context of spec version 2026-07-28 with StatelessProtocol

omouren:
Hello @chr-hertel
I’ve rebased onto main and resolved the merge conflict.

I haven’t looked at spec revision 2026-07-28 / StatelessProtocol yet, and I’m not entirely sure how to test the new spec for now. I retested with the previous spec and the events still work as expected there.

omouren:
Hello @chr-hertel 

Follow-up on 2026-07-28 / StatelessProtocol.

I did not map elicitation onto ServerRequestEvent / ClientResponseEvent as there is no server-initiated JSON-RPC on that revision, so those events would lie about the wire. They stay for the handshake-era.

I updated the StatelessProtocol to dispatch the same lifecycle events as the handshake path:
- RequestEvent (after InputContext is lifted)
- ResponseEvent / ErrorEvent from finalize()
- An ask is a ResponseEvent + InputRequiredResult
- The answer is the next RequestEvent + InputContext

src/Event/ClientResponseEvent.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        return $this->response;
+    }
+
+    public function getSession(): SessionInterface
+    {
+        return $this->session;
+    }
+
+    public function getId(): string|int
```
chr-hertel: in case `$response` is an `Error` instance, the ID could also be `null`:

```suggestion
    public function getId(): string|int|null
```

unfortunately our PHPStan setup is not on the proper level to catch that.

docs/advanced/events.md
```diff
@@ -72,6 +81,26 @@ The SDK dispatches 4 broad event types at the protocol level, allowing you to ob
...
+- `getMethod(): string` - Convenience method to get the request method
+
+### ClientResponseEvent
+
+**Dispatched**: When the server receives a client response to a prior outgoing request.
+
+**Properties**:
+- `getResponse(): Response|Error` - The client's reply
+- `getSession(): SessionInterface` - The current session
+- `getId(): string|int` - The JSON-RPC message ID
```
chr-hertel: same here

src/Server/Protocol.php
```diff
@@ -645,6 +658,49 @@ public function handleFiberYield(mixed $yieldedValue, ?Uuid $sessionId): void
...
+            if ($finalResult instanceof Response) {
+                $responseEvent = $this->dispatchEvent(new ResponseEvent($finalResult, $parentRequest, $session));
+                $finalResult = $responseEvent->getResponse();
+            } else {
+                $errorEvent = $this->dispatchEvent(new ErrorEvent($finalResult, $parentRequest, $session, null));
+                $finalResult = $errorEvent->getError();
+            }
+        }
+
+        $session->save();
```
chr-hertel, line 688: should be moved into the `if (null !== $parentRequest)`, no?

or maybe just invert that to keep the if small
```php
if (null === $parentRequest) {
    return $finalResult;
}
```

src/Server/Transport/TransportInterface.php
```diff
@@ -118,6 +118,15 @@ public function setResponseFinder(callable $finder): void;
...
     public function setFiberYieldHandler(callable $handler): void;
 
+    /**
+     * Set a handler invoked when a suspended Fiber completes.
+     *
+     * The transport calls this before sending the Fiber's final result to the client.
+     *
+     * @param callable(FiberReturn, Uuid): FiberReturn $handler
+     */
+    public function setFiberTerminationHandler(callable $handler): void;
```
chr-hertel, line 128: that's def a BC break and deserves a CHANGELOG entry please 🙏

docs/advanced/events.md
```diff
@@ -72,6 +81,26 @@ The SDK dispatches 4 broad event types at the protocol level, allowing you to ob
 - `getSession(): SessionInterface` - The current session
 - `getMethod(): string` - Convenience method to get the notification method
 
+### ServerRequestEvent
+
+**Dispatched**: When the server sends a request to the client (e.g. `elicitation/create`, `sampling/create`).
```
chr-hertel:
```suggestion
**Dispatched**: When the server sends a request to the client (e.g. `elicitation/create`, `sampling/createMessage`).
```

src/Server/Protocol.php
omouren: @chr-hertel 
save() has to stay even when $parentRequest is null. pull() already forgets SESSION_FIBER_PARENT_REQUEST. Without a save that deletion never hits the session store, so the next createWithId() would see the parent request again.

omouren:
Thanks for the review @chr-hertel, all done, except for the `null === $parentRequest`, you can see my comment above.

src/Server/Stateless/StatelessProtocol.php
```diff
@@ -692,20 +750,12 @@ private function streamFrames(\Generator $run, RequestMeta $meta, string $method
...
-        if (!$result instanceof Error && null !== $capabilityError = $this->checkInputRequests($result->result, $meta, $method, $id)) {
-            yield $capabilityError->message?->jsonSerialize();
-
-            return;
-        }
-
-        yield $result instanceof Error
-            ? $result->jsonSerialize()
-            : ['jsonrpc' => '2.0', 'id' => $id, 'result' => $this->codec->encodeResult($method, (array) $result->result->jsonSerialize(), $cacheable)];
+        yield json_decode($this->finalize($result, $request, $session, $meta, $method, $id, $cacheable)->toJson(), true, flags: \JSON_THROW_ON_ERROR);
```
chr-hertel: with this encode/decode extra turn we're replacing some `{}` with `[]` - build a reproducer and mitigation here: https://github.com/chr-hertel/php-sdk/compare/0b36721...pr-386-stream-frame-encoding

chr-hertel (changes requested):
was unable to patch this into your branch - most likely the last round. in general it's helpful to allow edits from maintainers 🙏

src/Server/Stateless/StatelessProtocol.php
omouren: Thanks, I merged your branch 👍

PR #393, opened 2026-07-13
[Schema] Fix Implementation::fromArray rejecting falsy-but-valid name/version

### LINKED ISSUE #392: Implementation::fromArray rejects clientInfo version "0" (empty() on a falsy string)
## Summary

`Mcp\Schema\Implementation::fromArray()` validates the version with `empty($data['version'])`. In PHP, `empty('0')` is true, so a client sending `clientInfo: {"name": "probe", "version": "0"}` is rejected even though "0" is a valid version string.

## Observed behavior (SDK 0.6, StreamableHttpTransport)

The thrown `InvalidArgumentException` makes the initialize message fail to parse as an `InitializeRequest`, so `Protocol::resolveSession()` falls through to the non-initialize branch and the client receives a misleading error instead of anything pointing at clientInfo:

```json
{"jsonrpc":"2.0","id":"","error":{"code":-32600,"message":"A valid session id is REQUIRED for non-initialize requests."}}
```

## Repro

POST an initialize request with `"clientInfo": {"name": "x", "version": "0"}` to any server using `StreamableHttpTransport`; the same payload with `"version": "1.0"` succeeds. Same applies to `name` since it uses the same check.

## Suggested fix

Replace the `empty()` checks in `Implementation::fromArray()` with explicit ones, e.g. `!isset($data['version']) || !is_string($data['version']) || $data['version'] === ''`.

DevkumarPatel:
Fixes #392.

## Problem

`Implementation::fromArray()` validates the required `name` and `version` fields with `empty()`. Because `empty('0') === true` in PHP, a perfectly valid `clientInfo`/`serverInfo` carrying `"version": "0"` (or `"name": "0"`) is rejected with an `InvalidArgumentException`.

During the `initialize` handshake this is especially confusing: the exception aborts parsing of the `InitializeRequest`, and `Protocol::resolveSession()` then reports the misleading error:

> A valid session id is REQUIRED for non-initialize requests.

…which points nowhere near the actual cause.

## Change

Replace the `empty()` guards with explicit checks:

```php
if (!isset($data['name']) || !\is_string($data['name']) || '' === $data['name']) {
if (!isset($data['version']) || !\is_string($data['version']) || '' === $data['version']) {
```

This keeps rejecting the cases that should fail — missing, non-string, and genuinely empty (`""`) — while accepting falsy-but-valid strings like `"0"`. Behavior is strictly widened; nothing previously valid is now rejected.

## Tests

Adds `tests/Unit/Schema/ImplementationTest.php` (there was no existing test for this class), covering:

- **Regression:** `"version": "0"` and `"name": "0"` are accepted.
- Rejection of missing / empty-string / non-string `name` and `version`.
- Constructor defaults, minimal & full `fromArray`, non-array `icons`, and `jsonSerialize` round-trip.

```
OK (14 tests, 34 assertions)
```

`vendor/bin/phpstan` reports no errors on the changed file.

## Scope note

The same `empty()`-on-required-string pattern exists in several sibling schema DTOs (`Icon`, `Prompt`, `PromptArgument`, `ResourceDefinition`, `ResourceTemplate`, `Root`, `Tool`) — e.g. a tool literally named `"0"` would hit the identical bug. I've kept this PR scoped to the reported issue; happy to follow up with a sweep (single PR or per-DTO) if you'd prefer.

### DIFF
--- a/src/Schema/Implementation.php
+++ b/src/Schema/Implementation.php
@@ -45,10 +45,10 @@ class Implementation implements \JsonSerializable
      */
     public static function fromArray(array $data): self
     {
-        if (empty($data['name']) || !\is_string($data['name'])) {
+        if (!isset($data['name']) || !\is_string($data['name']) || '' === $data['name']) {
             throw new InvalidArgumentException('Invalid or missing "name" in Implementation data.');
         }
-        if (empty($data['version']) || !\is_string($data['version'])) {
+        if (!isset($data['version']) || !\is_string($data['version']) || '' === $data['version']) {
             throw new InvalidArgumentException('Invalid or missing "version" in Implementation data.');
         }
 
--- a/tests/Unit/Schema/ImplementationTest.php
+++ b/tests/Unit/Schema/ImplementationTest.php
@@ -92,6 +92,7 @@ final class ImplementationTest extends TestCase
         $this->expectException(InvalidArgumentException::class);
         $this->expectExceptionMessage('Invalid or missing "name" in Implementation data.');
 
+        /* @phpstan-ignore argument.type */
         Implementation::fromArray(['version' => '1.0.0']);
     }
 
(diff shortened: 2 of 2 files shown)


chr-hertel (approved):

chr-hertel:
Thanks @DevkumarPatel for that patch - and follow-up for the broader scope very much welcome! :+1:

PR #395, opened 2026-07-14
[Client][Server] Add Roots support

wachterjohannes:
Adds support for the MCP **Roots** capability, letting a client expose `file://` "workspace folders" that the server is allowed to operate on.

### Client side
- `ListRootsRequestHandler` + `RootsCallbackInterface` — answer server `roots/list` requests from a user-provided callback (mirrors the existing `SamplingRequestHandler` / `SamplingCallbackInterface` pattern).
- `Client::sendRootsListChanged()` — send `notifications/roots/list_changed`. Guarded: it throws unless the client advertised the `roots.listChanged` capability, so a misconfigured client can't emit an undeclared notification.
- `Mcp\Exception\RootsException` — thrown from a roots callback, its message is forwarded to the server (same contract as `SamplingException`); any other throwable is logged and returned as a generic internal error.

### Server side
- `ClientGateway::listRoots()` — request the client's roots (returns `ListRootsResult`, throws `ClientException` on an error response).
- `ClientGateway::supportsRoots()` — capability-gating helper alongside `supportsElicitation()`.

### Schema
- `ListRootsResult::fromArray()` — parse the client's response; `Root` continues to enforce the `file://` URI scheme required by the spec.

### Docs / examples / tests
- README + `docs/client.md` sections and a runnable `examples/client/stdio_roots.php`.
- Unit tests covering the handler, result parsing (incl. non-`file://` rejection), capability serialization/round-trip, `sendRootsListChanged()` gating, and the `ClientGateway` roots methods.
- CHANGELOG entry under `0.7.0`.

### Conformance notes
- Root URIs are restricted to `file://` per the current spec.
- `roots` capability presence gates `roots/list`; `roots.listChanged` gates the `list_changed` notification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,11 @@ All notable changes to `mcp/sdk` will be documented in this file.
 * Add `maxBodyBytes` (default 4 MiB) to `StreamableHttpTransport` — POST bodies exceeding the cap are rejected with `413`. Unknown-size/chunked bodies are read incrementally and stopped at the cap so they cannot exhaust memory.
 * Reject malformed `Mcp-Session-Id` headers with a `400` response: a repeated header or a value that is not a valid UUID is now rejected up front instead of surfacing as an uncaught `Uuid::fromString()` error.
 * Extract RFC 9728 metadata serving into `ProtectedResourceMetadataHandler`, a transport-neutral PSR-15 `RequestHandlerInterface` that can be mounted directly as a Symfony/Laravel controller; `ProtectedResourceMetadataMiddleware` now delegates to it (no BC break).
+* Add client-side `roots/list` handler (`ListRootsRequestHandler` + `RootsCallbackInterface`) and `Client::sendRootsListChanged()`, plus server-side `ClientGateway::listRoots()` / `supportsRoots()` and `ListRootsResult::fromArray()`.
 
 0.6.0
 -----
 
-* Add client-side `roots/list` handler (`ListRootsRequestHandler` + `RootsCallbackInterface`) and `Client::sendRootsListChanged()`, plus server-side `ClientGateway::listRoots()` / `supportsRoots()` and `ListRootsResult::fromArray()`.
 * Add `Builder::add(Tool|ResourceDefinition|ResourceTemplate|Prompt $definition, ElementHandlerInterface $handler)` for explicit registration of elements whose schema is only known at runtime.
 * Add handler interfaces `ToolHandlerInterface`, `ResourceHandlerInterface`, `ResourceTemplateHandlerInterface`, `PromptHandlerInterface`, and the `ElementHandlerInterface` marker.
 * [BC Break] Renamed `Mcp\Schema\Resource` to `Mcp\Schema\ResourceDefinition`. No alias.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 0.6.0
 -----
 
+* Add client-side `roots/list` handler (`ListRootsRequestHandler` + `RootsCallbackInterface`) and `Client::sendRootsListChanged()`, plus server-side `ClientGateway::listRoots()` / `supportsRoots()` and `ListRootsResult::fromArray()`.
 * Add `Builder::add(Tool|ResourceDefinition|ResourceTemplate|Prompt $definition, ElementHandlerInterface $handler)` for explicit registration of elements whose schema is only known at runtime.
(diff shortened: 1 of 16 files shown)


src/Schema/Result/ListRootsResult.php
```diff
@@ -33,6 +34,28 @@ public function __construct(
...
+    public static function fromArray(array $data): self
+    {
+        if (!isset($data['roots']) || !\is_array($data['roots'])) {
+            throw new InvalidArgumentException('Missing or invalid "roots" in ListRootsResult data.');
+        }
+
+        $roots = array_map(
+            static fn (array $root): Root => Root::fromArray($root),
+            array_values($data['roots']),
+        );
```
chr-hertel: flaky in case `roots` doesn't carry an array as element => `TypeError`

chr-hertel (changes requested):

src/Schema/Result/ListRootsResult.php
wachterjohannes: @chr-hertel fixed

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 * Add `maxBodyBytes` (default 4 MiB) to `StreamableHttpTransport` — POST bodies exceeding the cap are rejected with `413`. Unknown-size/chunked bodies are read incrementally and stopped at the cap so they cannot exhaust memory.
 * Reject malformed `Mcp-Session-Id` headers with a `400` response: a repeated header or a value that is not a valid UUID is now rejected up front instead of surfacing as an uncaught `Uuid::fromString()` error.
 * Extract RFC 9728 metadata serving into `ProtectedResourceMetadataHandler`, a transport-neutral PSR-15 `RequestHandlerInterface` that can be mounted directly as a Symfony/Laravel controller; `ProtectedResourceMetadataMiddleware` now delegates to it (no BC break).
+* Add client-side `roots/list` handler (`ListRootsRequestHandler` + `RootsCallbackInterface`) and `Client::sendRootsListChanged()`, plus server-side `ClientGateway::listRoots()` / `supportsRoots()` and `ListRootsResult::fromArray()`.
```
chr-hertel: Needs to move to a 0.8.0 section please

src/Client.php
```diff
@@ -244,6 +246,22 @@ public function setLoggingLevel(LoggingLevel $level): void
...
+     *
+     * @throws RuntimeException if the client did not advertise the `roots.listChanged` capability
+     */
+    public function sendRootsListChanged(): void
+    {
+        if (true !== $this->config->capabilities->rootsListChanged) {
+            throw new RuntimeException('Cannot send a "roots/list_changed" notification without advertising the "roots.listChanged" capability. Build the client with new ClientCapabilities(roots: true, rootsListChanged: true).');
+        }
+
+        $this->protocol->sendNotification(new RootsListChangedNotification());
```
chr-hertel, line 267: i get why you're not using `sendRequest` - it's just a notification, but is this sufficient - we're skipping `isConnected` check here

examples/client/stdio_roots.php
chr-hertel, line 1: i don't think we need the try-catch block here

and, not sure, but can we demo it the other way around so that the list roots request actually is executed?

CHANGELOG.md
wachterjohannes: Moved to a new 0.8.0 section in 4c21047.

src/Client.php
wachterjohannes: Good catch — it was worse than just skipping the check: `Protocol::sendNotification()` sends via `$this->transport?->send()`, so on a disconnected client the notification was silently dropped instead of failing. It now throws a `ConnectionException` like `sendRequest()` does (capability check stays first), with a test for it. 5d278a9

examples/client/stdio_roots.php
wachterjohannes: Both done in 5d278a9. try/catch is gone, and the demo is turned around: the client-communication server now has an `inspect_workspace_roots` tool using `ClientGateway::supportsRoots()`/`listRoots()`, and the example calls it — so the server really issues `roots/list` and the client's handler answers. Output when run:

```
Calling 'inspect_workspace_roots'...
[ROOTS] Server requested the client's list of roots

Result:
{
    "status": "ok",
    "message": "Client exposed 2 root(s).",
    "roots": [
        {"uri": "file:///home/user/projects/app", "name": "Application"},
        {"uri": "file:///home/user/projects/library", "name": "Library"}
    ]
}
```

The `sendRootsListChanged()` notification is kept at the end as a secondary demo.

chr-hertel (dismissed):

chr-hertel:
Thanks @wachterjohannes! Will patch the CHANGELOG before merging.

chr-hertel (approved):

PR #400, opened 2026-07-19
[Schema] Add resource_link content block

### LINKED ISSUE #399: Missing resource_link content block: tool results cannot reference resources without embedding them
The spec's `resource_link` content type (introduced in protocol revision 2025-06-18, still present in 2025-11-25 — both of which this SDK advertises in `ProtocolVersion`) has no corresponding schema class: `src/Schema/Content/` contains `TextContent`, `ImageContent`, `AudioContent`, `EmbeddedResource`, and the `ResourceContents` variants, but grepping the whole `src/` tree for `resource_link`/`ResourceLink` at v0.7.0 comes up empty.

**Why it matters:** `resource_link` is the lightweight way for a tool result to reference a resource *without embedding its bytes*. The only current alternative, `EmbeddedResource`, inlines the full `ResourceContents` — the wrong tool when a result references many or large resources. The concrete case we hit: a Pimcore DAM MCP server (pimcore/data-hub-simple-rest#287) whose search/tree tools return hits referencing image/video assets that are separately readable via registered resource templates. Per spec, each hit should carry a `resource_link` block with the template-constructed URI; embedding blobs in a 25-hit search result is not viable, so tool output is stuck smuggling URIs through text/structured content until the type exists.

**Proposal:** add `Schema/Content/ResourceLink` (`type: "resource_link"` with `uri`, `name`, plus the optional `title`/`description`/`mimeType`/`size`/`annotations` fields from the spec schema), accepted anywhere tool-result `content` blocks are (de)serialized.

Happy to open a PR for it if that's welcome — it looks self-contained (new schema class + registration in the content deserialization map).

cancan101:
## Summary

Implements the spec's `resource_link` content type (protocol revision 2025-06-18, still present in 2025-11-25), which was missing from the SDK: `src/Schema/Content/` had `TextContent`, `ImageContent`, `AudioContent`, and `EmbeddedResource`, but no way to reference a resource by URI/name without embedding its full contents.

Fixes #399

## Spec revision implemented

Per the MCP schema, `ResourceLink extends Resource` (`BaseMetadata` + `Icons`), i.e.:

```ts
interface ResourceLink extends Resource {
  type: "resource_link";
}
interface Resource extends BaseMetadata, Icons {
  uri: string;
  description?: string;
  mimeType?: string;
  annotations?: Annotations;
  size?: number;
  _meta?: { [key: string]: unknown };
}
```

`Mcp\Schema\Content\ResourceLink` mirrors this exactly: required `uri` + `name`, optional `title`, `description`, `mimeType`, `annotations`, `size`, `icons`, `_meta` — following the same constructor/`fromArray`/`jsonSerialize` conventions as the sibling `EmbeddedResource` and `Mcp\Schema\ResourceDefinition` classes.

## Registration points touched

`resource_link` is now accepted everywhere `resource` (`EmbeddedResource`) content blocks are (de)serialized by `type`, matching the spec's `ContentBlock` union (`TextContent | ImageContent | AudioContent | ResourceLink | EmbeddedResource`):

- `Mcp\Schema\Result\CallToolResult::fromArray()`
- `Mcp\Schema\Content\PromptMessage` (constructor type union, `fromArray()`, phpstan types)
- `Mcp\Capability\Formatter\PromptResultFormatter` (both the typed-array path and the direct-`Content`-instance path)

`SamplingMessage` / `CreateSamplingMessageResult` are intentionally left untouched — the spec's `SamplingMessageContentBlock` union does not include `ResourceLink`.

Also added a doc example (`docs/mcp-elements.md`) and a CHANGELOG entry.

## Test coverage

- `tests/Unit/Schema/Content/ResourceLinkTest.php`: construction, optional-field defaults, `jsonSerialize`/`fromArray` round-trip, invalid-type/missing-field/invalid-`_meta` rejection.
- `tests/Unit/Schema/Content/PromptMessageTest.php`: `resource_link` deserialization, serialization, round-trip, and constructor acceptance.
- `tests/Unit/Schema/Result/CallToolResultTest.php`: `resource_link` deserialization alongside other content types in a single result, serialization, and a full round-trip mixing all five content types.
- `tests/Unit/Capability/Formatter/PromptResultFormatterTest.php`: `resource_link` through both the `PromptMessage`-instance path and the raw-array `role`/`content` path.

Full unit suite (841 tests / 2557 assertions), `php-cs-fixer --dry-run`, `phpstan analyse`, `composer validate --strict`, and `make docs` all pass locally.

## Downstream use case

pimcore/data-hub-simple-rest#287 — a Pimcore DAM MCP server whose search/tree tools return hits referencing image/video assets that are separately readable via registered resource templates. Per spec, each hit should carry a `resource_link` block with the template-constructed URI; embedding blobs in a 25-hit search result isn't viable, so tool output was stuck smuggling URIs through text/structured content until this type existed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to `mcp/sdk` will be documented in this file.
 
+0.8.0
+-----
+
+* Add `Mcp\Schema\Content\ResourceLink` for the spec's `resource_link` content block (protocol revision 2025-06-18+), letting tool results and prompt messages reference a resource by URI/name without embedding its contents. Accepted anywhere `resource` (`EmbeddedResource`) content is (de)serialized: `CallToolResult::fromArray()`, `PromptMessage::fromArray()`, and `PromptResultFormatter`.
+
 0.7.0
 -----
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 * Prompt generators returning content as typed arrays (`['type' => 'text', ...]` etc.) no longer lose the optional fields: `annotations` on every content type, and `_meta` and an explicit `mimeType` on embedded resource contents, now carry through to the resulting `PromptMessage` instead of being silently dropped. A missing resource `mimeType` still defaults to `text/plain`/`application/octet-stream` as before.
 * Add `annotations` support to `ImageContent` (constructor, `fromArray()`, `fromFile()`, `fromString()`, `jsonSerialize()`), matching `TextContent` and `AudioContent`.
 * Add client-side `roots/list` handler (`ListRootsRequestHandler` + `RootsCallbackInterface`) and `Client::sendRootsListChanged()`, plus server-side `ClientGateway::listRoots()` / `supportsRoots()` and `ListRootsResult::fromArray()`.
+* Add `Mcp\Schema\Content\ResourceLink` for the spec's `resource_link` content block (protocol revision 2025-06-18+), letting tool results and prompt messages reference a resource by URI/name without embedding its contents. Accepted anywhere `resource` (`EmbeddedResource`) content is (de)serialized: `CallToolResult::fromArray()`, `PromptMessage::fromArray()`, and `PromptResultFormatter`.
 
 0.7.0
 -----
--- a/docs/mcp-elements.md
+++ b/docs/mcp-elements.md
@@ -107,7 +107,7 @@ public function returnVoid(): void { /* no return */ }           // Empty conten
 For fine control over output formatting:
 
 ```php
-use Mcp\Schema\Content\{TextContent, ImageContent, AudioContent, EmbeddedResource};
+use Mcp\Schema\Content\{TextContent, ImageContent, AudioContent, ResourceLink, EmbeddedResource};
 
 public function getFormattedCode(): TextContent
 {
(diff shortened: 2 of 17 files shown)


chr-hertel:
Hi @cancan101, thanks for working on it and happy to hear Pimcore is adopting MCP :sunglasses: :+1: 

one thing to add, before i jump into review, can you find a meaningful addition to one of our examples here?

getting stuff into the examples also brings in functional test coverage via inspector :)

cancan101:
Good call — pushed a commit that extends the `discovery-userprofile` example: `lookup_user` looks up a user by ID and returns a `resource_link` pointing at the existing `user://{userId}/profile` resource template instead of embedding it, mirroring the search-hit-referencing-a-resource use case from the PR description.

This is now exercised by that example's inspector snapshot tests (`tools_list`, `tools_call-lookup_user`, `tools_call-lookup_user_not_found`), and it actually caught a real bug: `ToolReference::extractStructuredContent()` was passing the raw `[TextContent, ResourceLink]` return straight through as `structuredContent`, which serializes to a JSON list — the inspector's MCP client rejected the tool call outright since the spec requires `structuredContent` to be an object. Fixed by skipping structured-content extraction whenever the raw result contains a `Content` instance, since those are already reflected in `content`. Added a unit test for that in `RegistryTest`.

Ran the full suite locally (Node 20+ for inspector, since 18 can't `npx` it): unit suite (842 tests), the full `tests/Inspector` suite, `php-cs-fixer`, `phpstan`, and `composer validate --strict` all green.

chr-hertel (dismissed):
Thanks @cancan101 - will be part of v0.8.0

chr-hertel (approved):

PR #407, opened 2026-07-28
[Schema] Always emit {} for empty tool schema properties

### LINKED ISSUE #405: Tool serializes empty inputSchema.properties as [] instead of {}
## Environment
- `mcp/sdk` v0.7.0
- PHP 8.4

## Summary

A `Tool` whose `inputSchema.properties` is an empty PHP array serializes to `"properties":[]`. JSON Schema requires an object there, so strict clients reject the whole `tools/list` payload — e.g. OpenAI-compatible clients answer `400 invalid_function_parameters: "[] is not of type 'object'"`. One parameterless tool then breaks every request, not just calls to that tool.

The SDK already knows empty `properties` must become `{}` — it does so in `SchemaGenerator::buildSchemaFromParameters()` and in `Tool::fromArray()` (via `normalizeSchemaProperties()`, PR #223). But this normalization lives only on those **input** boundaries. `Tool::__construct()` and `Tool::jsonSerialize()` do **not** enforce it, so any `Tool` built another way — or whose schema survived a `json_decode($json, true)` round-trip — reaches the wire unnormalized.

## Reproduction

Direct construction (bypasses `fromArray`):

```php
use Mcp\Schema\Tool;

$tool = new Tool(
    name: 'no_params',
    title: null,
    inputSchema: ['type' => 'object', 'properties' => []],
    description: null,
    annotations: null,
);

echo json_encode($tool);
// {"name":"no_params","inputSchema":{"type":"object","properties":[]}}   <-- invalid
```

The round-trip that bites real servers (an empty JSON object decodes to an empty PHP array, then re-serializes as `[]`):

```php
$schema = json_decode('{"type":"object","properties":{}}', true); // ['type'=>'object','properties'=>[]]
$tool = new Tool('t', null, $schema, null, null);

echo json_encode($tool); // "properties":[]  — the {} was silently lost
```

**Expected:** `"properties":{}` in every case.
**Actual:** `"properties":[]` whenever the schema wasn't produced by `SchemaGenerator` or passed through `Tool::fromArray()`.

## Additional facet — normalization is also shallow

Even `SchemaGenerator` and `Tool::fromArray()` only normalize the **top-level** `properties`. A nested object parameter with no members, or an empty `outputSchema.properties`, still serializes as `[]`:

```php
$tool = Tool::fromArray([
    'name' => 't',
    'inputSchema' => [
        'type' => 'object',
        'properties' => ['filter' => ['type' => 'object', 'properties' => []]], // nested, not normalized
    ],
]);
echo json_encode($tool); // ...{"filter":{"type":"object","properties":[]}}  <-- still invalid
```

## Suggested fix

Make the object invariant total rather than boundary-only:
- Normalize in `Tool::__construct()` (and/or `Tool::jsonSerialize()`) so a `Tool` can never hold or emit `properties: []`, regardless of how it was built.
- Make the walk recursive so nested object schemas and `outputSchema` are covered too, not just the top-level `properties`.

This mirrors the precedent in PR #378 ("Always emit `items` for array tool parameter schemas"), which fixed the same class of strict-client rejection (VS Code / Copilot) in the SDK rather than leaving each consumer to patch it at the transport layer. Downstream servers currently have to re-parse and rewrite every `tools/list` response to work around this.

BrocksiNet:
## Summary
- Move empty-`properties` normalization into `Tool::__construct()` so every construction path (not only `fromArray` / `SchemaGenerator`) emits JSON Schema objects as `{}` instead of `[]`.
- Make the walk recursive across nested property schemas, `items`, `additionalProperties`, and combinators, and apply it to `outputSchema` as well.
- Add unit coverage for direct construction, `json_decode` round-trips, nested objects, array item schemas, and empty `outputSchema.properties`.

Fixes #405

## Test plan
- [x] `make unit-tests` (834 tests)
- [x] `make phpstan`
- [x] `make cs`
- [x] `make inspector-tests` (95 tests, 7 skipped)
- [x] `composer validate --strict`
- [ ] Confirm a parameterless tool serializes `"properties":{}` in `tools/list` against a strict client (e.g. OpenAI-compatible)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 0.8.0
 -----
 
-* Always emit `{}` for an empty tool schema `properties` map: `Tool` normalizes it in the constructor, recursively and for both `inputSchema` and `outputSchema`, so `tools/list` never serializes an invalid `properties: []`.
+* Always emit `{}` for empty tool schemas: `Tool` recursively normalizes every empty sub-schema — `properties`, `items`, `additionalProperties`, `$defs`, combinators and the other draft-07 to 2020-12 schema keywords — in the constructor, for both `inputSchema` and `outputSchema`, so an object position is never serialized as `[]`.
 
 0.7.0
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to `mcp/sdk` will be documented in this file.
 
+0.8.0
+-----
+
+* Always emit `{}` for an empty tool schema `properties` map: `Tool` normalizes it in the constructor, recursively and for both `inputSchema` and `outputSchema`, so `tools/list` never serializes an invalid `properties: []`.
+
 0.7.0
 -----
 
--- a/src/Schema/Tool.php
+++ b/src/Schema/Tool.php
@@ -28,7 +28,7 @@ use Mcp\Exception\InvalidArgumentException;
  *     type: 'object',
  *     properties?: array<string, mixed>|\stdClass,
  *     required?: string[]|null,
- *     additionalProperties?: bool|array<string, mixed>,
+ *     additionalProperties?: bool|array<string, mixed>|\stdClass,
  *     description?: string
  * }
  * @phpstan-type ToolData array{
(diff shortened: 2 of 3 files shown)


chr-hertel (dismissed):
Thanks @BrocksiNet - will extend the list of supported keywords and rebase for v0.8.0 while merging. Thanks for the patch!

chr-hertel (approved):

PR #408, opened 2026-08-07
docs: fix guide fragment links

### LINKED ISSUE #232: [Docs] Links within markdown guides are broken
With the current config the cross-links in markdown files are broken after build with phpDocumentor.
Either it's on me to understand too little about that or we have three options here:
1. switch from MD to RST for proper menu structure - would also bring the guides to left instead of top menu
2. hardlink to html files, which would lead to broken links on github MD preview
3. switch from phpDocumentor to something else

wWzZb:
## What

- Replace three broken cross-page fragment links in `docs/mcp-elements.md` with canonical published guide URLs.
- Keep the same target sections while making the links work in both GitHub Markdown and generated documentation.

## Why

phpDocumentor does not resolve Markdown paths that include fragments, so the generated links disappear. Canonical absolute URLs preserve the fragments without changing the documentation layout.

## Impact

Documentation-only change; no runtime behavior is affected.

## Checks

- `make docs`

Closes #232

### DIFF
--- a/docs/mcp-elements.md
+++ b/docs/mcp-elements.md
@@ -40,10 +40,10 @@ Each capability can be registered using two methods:
 - **Resource Templates**: Same `uriTemplate`  
 - **Prompts**: Same `name`
 
-For manual registration details, see [Server Builder Manual Registration](server-builder.md#manual-capability-registration).
+For manual registration details, see [Server Builder Manual Registration](https://php.sdk.modelcontextprotocol.io/server-builder.html#manual-capability-registration).
 
 For runtime, config-driven elements whose shape is not known at compile time, see
-[Explicit element registration](server-builder.md#explicit-element-registration) in the Server Builder docs.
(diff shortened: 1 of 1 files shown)


chr-hertel:
Unfortunately absolute links break the experience in GitHub - I think unless the upstream issue is not solved and we don't switch to RST, we'd need to switch the tooling :thinking:

wWzZb:
Agreed — absolute published URLs are not the right fix because they degrade the GitHub browsing experience.

I reproduced the underlying `page.md#fragment` failure on the latest phpDocumentor/guides `main` using this SDKs actual `mcp-elements.md` → `server-builder.md#...` links. The upstream fix is phpDocumentor/guides#1300. I refreshed it onto current `main`, confirmed PHP 8.5 lowest dependencies and the full locked suite pass, and added regression coverage for the PHP SDK link shape. Since I cannot update the contributor branch directly, the clean mergeable refresh is at https://github.com/jonathanhefner/phpDocumentor-guides/pull/2.

I am closing this workaround in favor of the upstream fix. Issue #232 should remain open until that fix is merged and available to the docs build.

PR #409, opened 2026-08-07
[Schema][Server] feat: support sampling with tools

### LINKED ISSUE #155: [Server][Schema] Implement SEP-1577: Sampling With Tools
This is a tracking issue for implementation of [SEP-1577](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1577).

### Summary

This extension adds `tools` and `toolChoice` parameters to `sampling/createMessage`, enabling MCP servers to run their own agentic loops using the client's tokens under user supervision. The PHP SDK needs to implement:

- New `tools` and `toolChoice` parameters in `CreateMessageRequest`
- New content types: `ToolUseContent` and `ToolResultContent` in `SamplingMessage`
- Updated `CreateMessageResult` with array content support and explicit `stopReason` values
- New client capability: `sampling.tools`
- Soft-deprecation of `includeContext` (now fenced by `sampling.context` capability)

This reduces complexity for client implementations while unlocking agentic tool-calling patterns in sampling requests.

### Related Issues & PRs

- **Implementation PRs**: n/a
- **Related PRs**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1796
- **Related Issues**: n/a

wWzZb:
## What

- Add `tools` and `toolChoice` to sampling requests and wire them through `ClientGateway::sample()`.
- Add sampling tool-use and tool-result content blocks, array content support, message metadata, and standard stop-reason values with open-string compatibility.
- Expose `sampling.context` and `sampling.tools` client capabilities.
- Add round-trip and validation coverage for the new schema types.

## Why

The SDK did not yet model the SEP-1577 protocol additions, so servers could not express tool-enabled sampling loops through the typed API.

## Impact

Existing single-content sampling calls remain supported. The new constructor parameters are optional and appended to preserve current call sites.

## Checks

- `make cs`
- `make phpstan`
- `make tests` — 938 tests, 2786 assertions, 7 existing skips
- `make docs`

Closes #155


src/Schema/Result/CreateSamplingMessageResult.php
```diff
@@ -28,17 +30,28 @@
...
+     * @param SamplingStopReason|string|null                                                                                  $stopReason the reason why sampling stopped, if known
+     * @param ?array<string, mixed>                                                                                           $meta       optional message metadata
      */
     public function __construct(
         public readonly Role $role,
-        public readonly TextContent|ImageContent|AudioContent $content,
+        public readonly TextContent|ImageContent|AudioContent|ToolUseContent|array $content,
         public readonly string $model,
-        public readonly ?string $stopReason = null,
+        public readonly SamplingStopReason|string|null $stopReason = null,
```
chr-hertel: if we adopt the enum here, we can basically drop `string`, right?


```suggestion
        public readonly ?SamplingStopReason $stopReason = null,
```

src/Schema/Content/SamplingMessage.php
```diff
@@ -19,17 +19,41 @@
...
 class SamplingMessage extends Content
 {
+    /**
+     * @param TextContent|ImageContent|AudioContent|ToolUseContent|ToolResultContent|list<TextContent|ImageContent|AudioContent|ToolUseContent|ToolResultContent> $content
+     * @param ?array<string, mixed>                                                                                                                               $meta
+     */
     public function __construct(
         public readonly Role $role,
-        public readonly TextContent|ImageContent|AudioContent $content,
+        public readonly TextContent|ImageContent|AudioContent|ToolUseContent|ToolResultContent|array $content,
```
chr-hertel, line 37: please try to introduce a 
```php
@phpstan-type SamplingContent = TextContent|ImageContent|AudioContent|ToolUseContent|ToolResultContent
```

so we could maybe shorten this with:

```suggestion
        public readonly SamplingContent|SamplingContent[],
```

src/Schema/ClientCapabilities.php
```diff
@@ -57,8 +59,17 @@ public static function fromArray(array $data): self
...
         if (isset($data['sampling'])) {
             $sampling = true;
+            if (\is_array($data['sampling'])) {
+                $samplingContext = isset($data['sampling']['context']);
+                $samplingTools = isset($data['sampling']['tools']);
+            } elseif (\is_object($data['sampling'])) {
+                $samplingContext = property_exists($data['sampling'], 'context');
+                $samplingTools = property_exists($data['sampling'], 'tools');
+            }
         }
```
chr-hertel, line 73: please add tests for those code branches

chr-hertel (changes requested):
Thanks @wWzZb for working on this - please extend the CHANGELOG.md for upcomgin release `0.8.0`, extend docs and example where meaningful and double-check if there's still missing unit tests here and there - i at least caught one place where i think it makes sense to add some.

Thanks again, great addition for feature parity with the spec! 👍

src/Schema/Result/CreateSamplingMessageResult.php
wWzZb: I kept the `string` arm intentionally. The current MCP schema defines `stopReason` as an open string so clients can return provider-specific reasons, while `SamplingStopReason` documents the standard values. `fromArray()` now maps known values to the enum and preserves unknown values; the provider-specific path is covered by a unit test. See https://modelcontextprotocol.io/specification/2025-11-25/schema#create-message-result.

src/Schema/Content/SamplingMessage.php
wWzZb: Added a `SamplingContent` PHPStan type alias and used it for the constructor parameter and hydration return type in 991fe4f.

src/Schema/ClientCapabilities.php
wWzZb: Added explicit coverage for object-form sampling capabilities in 991fe4f. The existing round-trip test covers the array-form branch, so both accepted input shapes are now exercised.

wWzZb:
Addressed the review feedback in 991fe4f:

- added the sampling content PHPStan alias;
- covered both array- and object-form sampling capability hydration;
- added the upcoming 0.8.0 changelog entry;
- documented tool-enabled sampling with a callback example;
- strengthened the provider-specific `stopReason` round-trip test.

I kept `SamplingStopReason|string|null` because the MCP schema explicitly defines `stopReason` as an open string for provider-specific values, and replied with the spec reference in that thread.

Local checks: `make cs`, `make phpstan`, `make tests` (937 tests, 2785 assertions, 7 skips), and `make docs`.

wWzZb:
Follow-up spec audit in 0b1cb6d:

- tightened `ToolResultContent` so its constructor only accepts standard tool-result content blocks;
- added regression coverage for rejecting non-standard nested message content;
- updated the server-side sampling guide to the 2025-11-25 specification and documented `tools`, `toolChoice`, and the capability requirements;
- documented the soft-deprecation rules for `includeContext`.

Checks: `make cs`, `make phpstan`, `make tests` (938 tests, 2786 assertions, 7 skips), and `make docs`.

src/Schema/Result/CreateSamplingMessageResult.php
chr-hertel: good point - i don't think adopting the enum is not worth it then. this way the enum adds some kind of fuzziness => we might upcast your value from string to enum - or not. users might check against string and later we add another enum case and have a BC break.

chr-hertel: reverted for now to get this merged - feel free to open the discussion again if i missed something :+1:

chr-hertel (approved):

chr-hertel:
Thanks @wWzZb!

PR #411, opened 2026-08-10
fix: normalize empty tool schema properties

### LINKED ISSUE #405: Tool serializes empty inputSchema.properties as [] instead of {}

Hanshuk:
## Summary

Fixes #405.

PHP serializes empty arrays as `[]`, which causes an empty JSON Schema `properties` object to be serialized as an array instead of the required JSON object `{}`.

This can cause strict MCP clients to reject otherwise valid tool definitions.

### Changes

* Recursively normalize empty `properties` arrays to `\stdClass`.
* Apply schema normalization during `Tool::jsonSerialize()` so directly constructed `Tool` instances are handled correctly despite readonly schema properties.
* Apply the same normalization to both `inputSchema` and `outputSchema`.
* Preserve legitimate JSON Schema arrays such as `required` and `enum`.
* Retain the existing `Tool::fromArray()` normalization behavior.

### Test coverage

Added regression coverage for:

* Direct `Tool` construction with empty `properties`.
* `Tool::fromArray()` with empty `properties`.
* JSON decode → Tool construction → serialization.
* Nested object schemas.
* Existing non-empty `properties`.
* Legitimate `required` and `enum` arrays.
* `outputSchema` with empty `properties`.

### Validation

* `git diff --check` passes.
* PHPUnit could not be executed locally because PHP/Composer are unavailable in the development environment.

The implementation is intentionally limited to the schema normalization and regression tests required for #405.

### DIFF
--- a/src/Schema/Tool.php
+++ b/src/Schema/Tool.php
@@ -21,8 +21,8 @@ use Mcp\Exception\InvalidArgumentException;
  *
  * @phpstan-type ToolInputSchema array{
  *     type: 'object',
- *     properties: array<string, mixed>,
- *     required: string[]|null
+ *     properties?: array<string, mixed>,
+ *     required?: string[]|null
  * }
  * @phpstan-type ToolOutputSchema array{
  *     type: 'object',
--- a/src/Schema/Tool.php
+++ b/src/Schema/Tool.php
@@ -169,7 +169,7 @@ class Tool implements \JsonSerializable
         }
 
         foreach ($schema as $key => $value) {
-            if ($key !== 'properties' && \is_array($value)) {
+            if ('properties' !== $key && \is_array($value)) {
                 $schema[$key] = self::normalizeSchemaProperties($value);
             }
         }
--- a/src/Schema/Tool.php
+++ b/src/Schema/Tool.php
@@ -127,7 +127,7 @@ class Tool implements \JsonSerializable
         if (null !== $this->title) {
             $data['title'] = $this->title;
         }
-        $data['inputSchema'] = $this->inputSchema;
+        $data['inputSchema'] = self::normalizeSchemaProperties($this->inputSchema);
         if (null !== $this->description) {
             $data['description'] = $this->description;
         }
@@ -141,14 +141,14 @@ class Tool implements \JsonSerializable
             $data['_meta'] = $this->meta;
         }
         if (null !== $this->outputSchema) {
-            $data['outputSchema'] = $this->outputSchema;
+            $data['outputSchema'] = self::normalizeSchemaProperties($this->outputSchema);
         }
 
         return $data;
     }
 
     /**
-     * Normalize schema properties: convert an empty properties array to stdClass.
+     * Normalize schema properties: convert an empty properties array to stdClass recursively.
      *
      * @param array<string, mixed> $schema
      *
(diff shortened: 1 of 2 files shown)


chr-hertel:
Hi @Hanshuk, thanks for this, but basically a duplicate of #407, right?
If you have some feedback for that PR to improve it, please share it there.

chr-hertel:
Closed in favor of #407 - please open a new one, if I'm missing something. Thanks anyhow, @Hanshuk!

PR #422, opened 2026-08-15
[Schema] Support PHPStan/Psalm integer range types

### LINKED ISSUE #397: Schema generator - ability to use phpstan/psalm number intervals
**Describe the bug**
When you use typehint on property, then use `@param` to narrow it, it will translate as object in schema generator.

When you use `@param int<0, max> $offset` in your typehint, Schema generator will translate it to object, even if property itself is integer on language level

**To Reproduce**
```
/**
  * @param int<0, max> $offset
  */
public function sendMeData(
      #[Schema(minimum: 0)]
      int $offset,
) {
}
```

**Expected behavior**
Offset parameter is generated as type: integer into Schema, but it's `object`

**Additional context**
You can override it by adding `type: 'integer'` into Schema , but it's third time you need to do it for one value. When there is `int` defined on language level, there should be no way to change it by phpdoc.

phpdoc typehint is usefull for phpStan here.

wWzZb:
## Summary

- Treat PHPStan/Psalm `int<...>` PHPDoc ranges as the base JSON Schema `integer` type.
- Keep `minimum` and `maximum` under explicit `#[Schema]` control rather than deriving them from PHPDoc.
- Add regression coverage for numeric, `min`, and `max` interval bounds.

The generator gives a non-generic PHPDoc type precedence over the native `int` type. Since `int<...>` was not recognized by the mapper, it fell through to the `object` fallback.

Fixes #397

## Test plan

- [x] `vendor/bin/phpunit` (1,254 tests, 3,502 assertions, 7 skipped)
- [x] `vendor/bin/phpstan analyse --memory-limit=-1`
- [x] `vendor/bin/php-cs-fixer fix --dry-run`
- [x] `composer validate --strict`
- [x] Targeted regression test with `phpdocumentor/reflection-docblock` 5.6.0

### DIFF
--- a/src/Capability/Discovery/SchemaGenerator.php
+++ b/src/Capability/Discovery/SchemaGenerator.php
@@ -677,7 +677,12 @@ final class SchemaGenerator implements SchemaGeneratorInterface
             return ['array'];
         }
 
-        // PRIORITY 3: Handle unions (recursive)
+        // PRIORITY 3: Treat PHPStan/Psalm integer ranges as their base type
+        if (preg_match('/^int\s*<\s*[^<>]+\s*>$/i', $normalizedType)) {
+            return ['integer'];
+        }
+
+        // PRIORITY 4: Handle unions (recursive)
         if (str_contains($normalizedType, '|')) {
             $types = explode('|', $normalizedType);
             $jsonTypes = [];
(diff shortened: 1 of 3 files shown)


chr-hertel (approved):
Thank you @wWzZb!

Issue #359, opened 2026-05-26
[Schema] SEP-2164: Switch resource-not-found error code to -32602 with protocol-version-aware fallback


chr-hertel:
Implements [SEP-2164](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2164) for the MCP Spec [2026-07-28 release](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/).

Tracked by umbrella #337.

## Spec summary

> Resource-not-found now uses JSON-RPC `-32602` (Invalid Params) instead of the MCP-custom `-32002`, aligning with JSON-RPC.

## PHP SDK changes

- `src/Schema/JsonRpc/Error.php:37` defines `RESOURCE_NOT_FOUND = -32002`. Change to `-32602`, or deprecate the constant and add `RESOURCE_NOT_FOUND_INVALID_PARAMS = -32602`.
- `Error::forResourceNotFound()` (line ~109) is called from `ReadResourceHandler`, `ResourceSubscribeHandler`, `ResourceUnsubscribeHandler`, `GetPromptHandler`, `CompletionCompleteHandler` — all flip automatically once the constant changes.
- **Negotiate by protocol version:** pre-`2026-07-28` clients still expect `-32002`. Add version-aware dispatch in `Error::forResourceNotFound()` reading session/request protocol version.
- Update conformance & unit tests asserting the old code.

## Related

- Umbrella #337

Issue #337, opened 2026-05-26
[2026-07-28] Schema modernization & error code standardization


chr-hertel:
Tracking issue for the MCP Spec [2026-07-28 release](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/) — **Schema & errors** milestone.

## SEPs covered

| SEP | Title | Spec PR |
|---|---|---|
| SEP-2106 | Tools `inputSchema`/`outputSchema` conform to JSON Schema 2020-12 | [#2106](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2106) |
| SEP-2164 | Standardize resource-not-found error code to `-32602` | [#2164](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2164) |

## Sub-issues

**SEP-2106 (JSON Schema 2020-12):**
- [ ] #356 — Vocabulary support (oneOf/anyOf/allOf/if-then-else/`$ref`)
- [ ] #357 — Allow non-object `outputSchema` and `structuredContent`
- [ ] #358 — Harden `SchemaValidator` against external `$ref` SSRF and composition DoS

**SEP-2164:**
- [ ] #359 — Switch resource-not-found error code to `-32602` with protocol-version-aware fallback

## Notes

- Both SEPs are **merged**.
- SEP-2106 is the largest code-level lift this cycle for the schema layer: full JSON Schema 2020-12 vocabulary (composition, conditionals, `$ref`), plus non-object `outputSchema` / `structuredContent`. Also requires hardening against external `$ref` SSRF and composition DoS.
- SEP-2164 is mechanically a one-constant flip in `src/Schema/JsonRpc/Error.php` with a protocol-version-aware fallback.

PR #423, opened 2026-08-15
[Server] Return invalid-params for unknown tool in tools/call

latent-9:
## Summary

`CallToolHandler` returns `-32601` (Method not found) when a `tools/call` request names a tool that is not registered. The tool name is a parameter of a method that does exist, so the correct JSON-RPC error for this case is `-32602` (Invalid params).

## Details

`tools/call` is a valid, supported method. When its `name` argument does not resolve to a registered tool, the method was found but a parameter was invalid. `-32601` is reserved for an unknown method (an unsupported JSON-RPC method name), which is not what happened here.

The same handler already treats bad tool input as invalid params: when the arguments fail schema validation it returns `Error::forInvalidParams(...)` (`-32602`). An unknown tool name belongs to the same class of error and should use the same code.

The reference TypeScript schema classifies an unknown tool name under `InvalidParamsError`, and the other SDKs have converged on `-32602` for the unknown-name case (for example modelcontextprotocol/ruby-sdk#517 for `prompts/get`).

## Impact

A client that receives `-32601` for `tools/call` can reasonably conclude that the server does not implement `tools/call` at all and stop issuing tool calls, when in fact only the specific tool name was wrong. `-32602` correctly signals that the request was understood but a parameter was invalid, so the client can recover by choosing a valid tool.

## Scope

This is limited to the tool-not-found path and is independent of #337 and #359 (resource and prompt not-found moving from `-32002` to `-32602`). No protocol-version gate is needed: `-32602` is a standard JSON-RPC code valid in every MCP revision, and `-32601` was never the specified value for an unknown tool name.

## Changes

- `CallToolHandler`: return `Error::forInvalidParams(...)` instead of `Error::METHOD_NOT_FOUND` when a tool is not found.
- Update the corresponding unit test to assert `Error::INVALID_PARAMS`.


chr-hertel:
Solved already with #447 - thanks anyhow @latent-9!

PR #424, opened 2026-08-16
[Server] Reject JSON-RPC batch requests

ez-lbz:
MCP no longer supports JSON-RPC batches: a POST body must be a single JSON-RPC message. `MessageFactory` currently accepts top-level arrays and hydrates each entry, so a batch is processed as a set of messages instead of being refused outright.

This change rejects any top-level array as invalid input before any entry is hydrated, returning a single `InvalidInputMessageException` (the existing per-message error contract) instead of processing the batch. The now-dead `maxBatchSize` cap, its constructor parameter, and `DEFAULT_MAX_BATCH_SIZE` are removed.

Tests:
- `MessageFactoryTest`: batch payloads (valid, mixed, and error-containing) are asserted to be rejected wholesale; obsolete `maxBatchSize` tests removed.
- `MalformedInputTest`: a batch payload is asserted to be rejected without hydrating any entry.
- `Server\ProtocolTest`: batch input is asserted to produce a single Invalid Request error and to never hydrate batch entries.

### DIFF
--- a/src/JsonRpc/MessageFactory.php
+++ b/src/JsonRpc/MessageFactory.php
@@ -120,7 +120,7 @@ final class MessageFactory
         // A list is a batch array. MCP removed JSON-RPC batches from the
         // protocol, so the whole payload is invalid rather than a set of messages.
         if (array_is_list($data)) {
-            return [new InvalidInputMessageException('JSON-RPC batch requests are not supported; send a single JSON-RPC message.')];
+            return [new InvalidInputMessageException('JSON-RPC batch requests are not supported anymore since specification release 2025-06-18; send a single JSON-RPC message.')];
         }
 
         try {
--- a/src/JsonRpc/MessageFactory.php
+++ b/src/JsonRpc/MessageFactory.php
@@ -69,23 +69,11 @@ final class MessageFactory
     ];
 
     /**
-     * Upper bound on the number of messages accepted in a single batch, guarding
-     * against amplification where one small request expands into many operations.
-     */
-    public const DEFAULT_MAX_BATCH_SIZE = 100;
-
-    /**
      * @param list<class-string<Request>|class-string<Notification>> $registeredMessages
-     * @param int                                                    $maxBatchSize       Maximum number of messages accepted in a single JSON-RPC batch
      */
     public function __construct(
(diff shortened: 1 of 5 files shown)


src/JsonRpc/MessageFactory.php
chr-hertel, line 104: this can be simplified even further now:

```suggestion
    * @throws InvalidInputMessageException When the input data is not a valid message
    * @throws \JsonException When the input string is not valid JSON
    */
public function create(string $input): MessageInterface
```

chr-hertel (changes requested):
Hi @ez-lbz, thanks for bringing this up - this will simplify that part message handling quite a bit :)

src/JsonRpc/MessageFactory.php
```diff
@@ -115,44 +105,29 @@ public function create(string $input): array
...
+        // protocol, so the whole payload is invalid rather than a set of messages.
         if (array_is_list($data)) {
-            if (\count($data) > $this->maxBatchSize) {
-                return [new InvalidInputMessageException(\sprintf('JSON-RPC batch size %d exceeds the maximum allowed batch size of %d.', \count($data), $this->maxBatchSize))];
-            }
-
-            $batch = $data;
-        } else {
-            $batch = [$data];
+            return [new InvalidInputMessageException('JSON-RPC batch requests are not supported; send a single JSON-RPC message.')];
```
chr-hertel:
```suggestion
            return [new InvalidInputMessageException('JSON-RPC batch requests are not supported anymore since specification release 2025-06-18; send a single JSON-RPC message.')];
```

ez-lbz:
Thanks for the review! Applied the suggestion to reference the spec release that removed batches in the error message (commit 752c430). The `maxBatchSize` parameter and cap were already removed as dead code in the original change.

chr-hertel:
Putting this on hold for now - the spec version `2025-06-18` removed batch support but we currently still support that version. Will check back, in my understanding we only need to support the current and two older versions, but need to check back.

Anyhow, when we tackle this, we should go even one step further and change the method profile to `Mcp\JsonRpc\MessageFactory::create(string $input): MessageInterface`

PR #430, opened 2026-08-17
[Schema] Drop SDK-only name pattern on resources and resource templates

chr-hertel:
`ResourceDefinition` and `ResourceTemplate` rejected any `name` not matching `[a-zA-Z0-9_-]+`. That rule is ours, not the spec's, and since both classes also parse what a peer sent (`fromArray()`), it made the client choke on conformant servers over a display string.

**Spec evidence**

- `schema/2025-11-25/schema.json`: `name` on `Resource`, `ResourceTemplate` (via `BaseMetadata`) is a plain `{"type": "string"}` — no `pattern`, `minLength` or `format`. There is no `pattern` keyword anywhere in the schema.
- `schema/2025-11-25/schema.ts` `BaseMetadata.name`: *"Intended for programmatic or logical use, but used as a display name in past specs or fallback (if title isn't present)."* — nothing about allowed characters.
- The spec's own `resources/list` / `resources/templates/list` examples use names our regex rejected: `"main.rs"`, `"README.md"`, `"Project Files"` ([server/resources.mdx](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/2025-11-25/server/resources.mdx)).

URI / URI-template validation is unchanged. Tests cover the spec example names; CHANGELOG notes the (loosening) BC break.

**"Isn't there a name format rule somewhere?"** — yes, but only for tools: [SEP-986](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/986) (`2025-11-25`, [tools.mdx › Tool Names](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool-names)) says tool names *SHOULD* be `[A-Za-z0-9_.-]{1,128}`. It is SHOULD-level, scoped to tools, and no revision from `2024-11-05` through `2026-07-28`/draft carries anything comparable for resource, resource template or prompt names. The RFC references in `resources.mdx` (RFC 3986 / RFC 6570) concern `uri` / `uriTemplate`, which this PR keeps validating.


src/Schema/ResourceDefinition.php
```diff
@@ -68,9 +63,9 @@ public function __construct(
         public readonly ?array $icons = null,
         public readonly ?array $meta = null,
     ) {
-        if (!preg_match(self::RESOURCE_NAME_PATTERN, $name)) {
-            throw new InvalidArgumentException(\sprintf('Invalid resource name "%s": must contain only alphanumeric characters, underscores, and hyphens.', $name));
-        }
+        // `name` is deliberately unconstrained: the specification puts no
+        // pattern on it, and this class also parses what a server sent, so a
+        // rule of our own would reject conformant peers over a display string.
```
chr-hertel:
```suggestion
```

src/Schema/ResourceTemplate.php
```diff
@@ -61,9 +56,8 @@ public function __construct(
         public readonly ?Annotations $annotations = null,
         public readonly ?array $meta = null,
     ) {
-        if (!preg_match(self::RESOURCE_NAME_PATTERN, $name)) {
-            throw new InvalidArgumentException(\sprintf('Invalid resource name "%s": must contain only alphanumeric characters, underscores, and hyphens.', $name));
-        }
+        // See {@see ResourceDefinition}: the specification puts no pattern on
+        // `name`, so neither does this.
```
chr-hertel:
```suggestion
```

PR #433, opened 2026-08-18
[Server] Log expected tool failures at debug level

### LINKED ISSUE #157: [Server] Lower log level on failing tool calls
Hi @chr-hertel , sorry, but I didn't have time to finish the PR.

At a quick glance it looks like a sufficient solution, the only question is whether it's really necessary to log errors from ToolCallException. In my experience, they were never needed, and in the worst case you can log directly in the McpTool method.

If I take an example where the AI/human calls divideNumbers with $b as 0 and McpTool responds that you can't divide by zero, I would expect it to handle that and provide a different number or ask for a different one. Otherwise, I would end up with Sentry full of logs that I can’t and don’t want to deal with, because it’s not an application error.

Logging is probably fine during debugging, but I would keep it at most at debug/info severity.

Thanks for mention.

_Originally posted by @MartkCz in https://github.com/modelcontextprotocol/php-sdk/issues/73#issuecomment-3527374587_

ousamabenyounes:
## Summary

- log expected `ToolCallException` failures at debug severity instead of error severity
- keep unexpected exceptions at error severity
- preserve and verify the existing log message and context

Fixes #157

## Test verification (RED → GREEN)

Before the production change, the focused regression test failed with:

```text
Expected level: debug
Actual level: error
FAILURES! Tests: 1, Assertions: 1, Failures: 1.
```

After the change:

```text
OK (26 tests, 125 assertions)
```

Full local CI replay:

```text
PHP CS Fixer: 0 files can be fixed
PHPStan: No errors
PHPUnit: OK (1247 tests, 3396 assertions)
phpDocumentor: complete
php-diff-coverage: PASS 1/1 changed executable lines covered
```


chr-hertel (approved):
Thanks @ousamabenyounes! 👍

PR #438, opened 2026-08-18
[Schema] Serialize a settingless extension as an object

chr-hertel:
An extension declaring support with no settings was advertised as `[]`, an empty JSON array, which the wire schema rejects — `extensions` maps an id to a settings *object*.

Both `ServerCapabilities` and `ClientCapabilities` now cast each entry, so it goes out as `{}`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ All notable changes to `mcp/sdk` will be documented in this file.
 0.8.0
 -----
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 0.8.0
 -----
 
+* Fix an extension with no settings being advertised as `[]` rather than `{}` in `ServerCapabilities` and `ClientCapabilities`, which the wire schema rejects.
```
```suggestion
```

Issue #352, opened 2026-05-26
[Server][Capability][Examples] SEP-1865 MCP Apps: Link tools to UI templates via attribute metadata + example/inspector tests


chr-hertel:
Implements the tool↔UI linkage and example coverage for [SEP-1865](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1865) (MCP Apps) for the MCP Spec [2026-07-28 release](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/).

Tracked by umbrella #335.

## PHP SDK changes

- Extend `#[McpTool]` (or add a sibling attribute) to declare which `ui://` template a tool's result should render in.
- Update the tool result formatter in `src/Capability/Formatter/` to emit the linkage metadata in `tools/call` responses per the spec wire format.
- Add `examples/server/apps/` walkthrough: a tool that returns structured data plus a `ui://` template that renders it.
- Inspector tests verifying the wire format end-to-end.

## Related

- Umbrella #335
- Depends on SEP-1865 schema + SEP-1865 builder sub-issues

chr-hertel:
Covered by #281 and the remainder by #429 (auto-closes on merge). Gap analysis in the PR description.

PR #439, opened 2026-08-18
[Examples][Tests] Make the MCP Apps example work over HTTP, and pin it

chr-hertel:
It was the only example without a session store, so every request under `php -S` or PHP-FPM — each its own process — was answered "Session not found or has expired". It works over stdio and nowhere else. It now uses a `FileSessionStore` like the others.

Inspector snapshots pin what the extension actually puts on the wire: the `ui://` resource's `text/html;profile=mcp-app` MIME type and `ui` marker, and the `_meta.ui` on the tool that links the two.

Closes the test half of #352.


examples/server/mcp-apps/server.php
```diff
@@ -18,12 +18,17 @@
...
+use Mcp\Server\Session\FileSessionStore;
 
 logger()->info('Starting MCP Apps Example Server...');
 
 $server = Server::builder()
     ->setServerInfo('MCP Apps Weather Example', '1.0.0')
     ->setLogger(logger())
+    // Every request is its own PHP process under `php -S` and php-fpm alike, so
+    // the handshake era needs somewhere outside memory to keep the session.
+    // Without this the example works over stdio and nowhere else.
```
chr-hertel:
```suggestion
```

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 0.8.0
 -----
 
+* Fix the MCP Apps example being unusable over HTTP. It was the only example without a session store, so every request under `php -S` or PHP-FPM — each its own process — was answered "Session not found or has expired". It now uses a `FileSessionStore` like the others, and is covered by Inspector snapshot tests that pin what the extension actually puts on the wire: the `ui://` resource's `text/html;profile=mcp-app` MIME type and `ui` marker, and the `_meta.ui` on the tool that links the two.
```
```suggestion
```

PR #464, opened 2026-08-19
Fail loudly when discovery is configured but symfony/finder is missing

### LINKED ISSUE #398: Builder::build() silently skips configured file-based discovery when symfony/finder is missing — should fail loudly
Calling `setDiscovery()` is an explicit declaration that the server's tools come from attribute discovery. But since `symfony/finder` is only a `suggest`ed dependency, a host application can easily end up without it — and when that happens, `build()` does this:

```php
if (null !== $this->discoveryBasePath) {
    if (null !== $this->discoverer || class_exists(Finder::class)) {
        // ... discovery runs
    } else {
        $logger->warning('File-based discovery requires symfony/finder...');
    }
}
```

The result is the worst failure mode available: the server builds successfully, `initialize` succeeds, and `tools/list` returns an empty array. The operator sees a healthy server; the symptom surfaces far from the cause, as confused MCP clients with no tools. The only breadcrumb is a single warning log line.

This also contradicts the SDK's own `Discoverer::__construct()`, which already throws `RuntimeException('File-based discovery requires symfony/finder. ...')` for exactly this situation — the builder's `class_exists` pre-check just routes around that guard, downgrading a configured-but-impossible feature from an error to a whisper.

**Proposal:** when `discoveryBasePath` is set, no custom discoverer was supplied, and `Finder` is unavailable, `build()` should throw (the Discoverer's existing message is perfect) instead of warning-and-skipping. This costs nothing for explicit-registration users, client-only users, or anyone without `setDiscovery()` — it only converts a silent production mystery into an immediate, actionable boot error for people who asked for discovery and can't have it.

Observed on v0.7.0. Context: we hit this failure mode while integrating the SDK into a Symfony bundle (pimcore/data-hub-simple-rest#312) and worked around it by requiring `symfony/finder` in the bundle directly — which remains the right consumer-side fix, but doesn't help the next integrator who doesn't know about the silent path.

Faneraiy14:
## Summary

`Builder::build()` routed around `Discoverer::__construct()`'s own `class_exists(Finder::class)` guard: instead of throwing, it logged a single warning and silently skipped the `DiscoveryLoader`. The result is the worst failure mode available — the server builds successfully, `initialize` succeeds, and `tools/list` quietly returns an empty array, with no actionable signal beyond one log line the operator may never see.

`setDiscovery()` is an explicit request for file-based discovery; if that request can't be honored, `build()` should fail immediately with the same `RuntimeException` and message `Discoverer` already throws for the identical condition, rather than downgrading a configured-but-impossible feature to a whisper.

Fixes #398

## Test plan

- [x] `vendor/bin/phpunit --testsuite=unit` — 1513 tests, 3919 assertions, all passing
- [x] `vendor/bin/phpunit --testsuite=integration` — 64 tests, 145 assertions, all passing
- [x] `vendor/bin/phpstan analyse` (full repo, level 6 per `phpstan.dist.neon`) — no errors

### DIFF
--- a/src/Server/Builder.php
+++ b/src/Server/Builder.php
@@ -29,6 +29,7 @@ use Mcp\Capability\Registry\ReferenceHandlerInterface;
 use Mcp\Capability\RegistryInterface;
 use Mcp\Exception\InvalidArgumentException;
 use Mcp\Exception\LogicException;
+use Mcp\Exception\RuntimeException;
 use Mcp\JsonRpc\MessageFactory;
 use Mcp\Schema\Annotations;
 use Mcp\Schema\Enum\ProtocolVersion;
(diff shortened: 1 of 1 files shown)


chr-hertel:
Hi there @Faneraiy14, thanks but this is already addressed with #432 - which is even on hold until behavior is settled as wanted or not.

PR #465, opened 2026-08-19
[Server][Capability] Cast variadic tool parameters element-by-element

Faneraiy14:
## Summary

`SchemaGenerator::buildVariadicParameterSchema()` already advertises a variadic parameter (e.g. `int ...$scores`) as a JSON `"array"` schema, so a spec-conformant client sends an array for it. But `ReferenceHandler::prepareArguments()` never special-cased `ReflectionParameter::isVariadic()` — it fell through to the regular single-value branch and passed the whole array to `castArgumentType()`, which tried to cast it as one scalar and failed (e.g. `"Cannot cast value to integer"` for `int ...$scores` given `[1, 2, 3]`).

Any tool handler declaring a variadic parameter was therefore unreachable through the advertised schema — confirmed by direct reproduction against `ReferenceHandler::handle()` before this change:

```
EXCEPTION: Mcp\Exception\RegistryException: Cannot cast value to integer. Expected integer representation.
```

This resolves the `// TODO: Handle variadic parameters.` left in `prepareArguments()`.

## Fix

Casts each array element individually via the existing `castArgumentType()` (unchanged — it already inspects the variadic parameter's *element* type, not `array`, per PHP reflection semantics) and appends to the result in order. Variadic is always the last parameter, so this preserves correct positional ordering. A non-array value for a variadic argument now fails clearly with an invalid-params error instead of the confusing cast message.

## Test plan

- [x] New unit tests: array argument cast element-by-element, omitted argument treated as zero elements, non-array argument rejected with a clear error
- [x] `vendor/bin/phpunit --testsuite=unit` — 1516 tests, 3923 assertions, all passing
- [x] `vendor/bin/phpstan analyse` (full repo, level 6 per `phpstan.dist.neon`) — no errors

### DIFF
--- a/src/Capability/Registry/ReferenceHandler.php
+++ b/src/Capability/Registry/ReferenceHandler.php
@@ -99,9 +99,32 @@ final class ReferenceHandler implements ReferenceHandlerInterface
         $finalArgs = [];
 
         foreach ($reflection->getParameters() as $parameter) {
-            // TODO: Handle variadic parameters.
             $paramName = $parameter->getName();
             $paramPosition = $parameter->getPosition();
+
+            if ($parameter->isVariadic()) {
+                // SchemaGenerator advertises variadic parameters as a JSON "array"
+                // schema (see buildVariadicParameterSchema()), so the incoming value
+                // here is an array whose elements each need casting to the variadic's
+                // element type. Falling through to castArgumentType() below would try
+                // to cast the whole array as a single scalar and fail (e.g. "Cannot
+                // cast value to integer" for `int ...$extra`). Variadic is always the
+                // last parameter, so appending here preserves correct final ordering.
+                $values = $arguments[$paramName] ?? [];
+                if (!\is_array($values)) {
+                    throw RegistryException::invalidParams(\sprintf('Parameter `%s` must be an array of values.', $paramName));
+                }
+                foreach (array_values($values) as $value) {
+                    try {
+                        $finalArgs[] = $this->castArgumentType($value, $parameter);
+                    } catch (InvalidArgumentException $e) {
+                        throw RegistryException::invalidParams($e->getMessage(), $e);
+                    } catch (\Throwable $e) {
+                        throw RegistryException::internalError("Error processing parameter `{$paramName}`: {$e->getMessage()}", $e);
+                    }
+                }
+                continue;
+            }
 
             // Check if parameter is a special injectable type
             $type = $parameter->getType();
--- a/src/Capability/Registry/ReferenceHandler.php
+++ b/src/Capability/Registry/ReferenceHandler.php
@@ -103,24 +103,23 @@ final class ReferenceHandler implements ReferenceHandlerInterface
             $paramPosition = $parameter->getPosition();
 
             if ($parameter->isVariadic()) {
-                // SchemaGenerator advertises variadic parameters as a JSON "array"
-                // schema (see buildVariadicParameterSchema()), so the incoming value
-                // here is an array whose elements each need casting to the variadic's
-                // element type. Falling through to castArgumentType() below would try
-                // to cast the whole array as a single scalar and fail (e.g. "Cannot
-                // cast value to integer" for `int ...$extra`). Variadic is always the
-                // last parameter, so appending here preserves correct final ordering.
+                // Each element is cast individually below; a non-array value is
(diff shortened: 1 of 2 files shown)


src/Capability/Registry/ReferenceHandler.php
```diff
@@ -99,10 +99,33 @@ private function prepareArguments(\ReflectionFunctionAbstract $reflection, array
...
             $paramPosition = $parameter->getPosition();
 
+            if ($parameter->isVariadic()) {
+                // SchemaGenerator advertises variadic parameters as a JSON "array"
+                // schema (see buildVariadicParameterSchema()), so the incoming value
+                // here is an array whose elements each need casting to the variadic's
+                // element type. Falling through to castArgumentType() below would try
+                // to cast the whole array as a single scalar and fail (e.g. "Cannot
+                // cast value to integer" for `int ...$extra`). Variadic is always the
+                // last parameter, so appending here preserves correct final ordering.
```
chr-hertel: Seven lines of rationale is heavier than the rest of this file, and the PR description already carries it. Two lines would do.

src/Capability/Registry/ReferenceHandler.php
```diff
@@ -99,10 +99,33 @@ private function prepareArguments(\ReflectionFunctionAbstract $reflection, array
...
+                $values = $arguments[$paramName] ?? [];
+                if (!\is_array($values)) {
```
chr-hertel, line 111: This regresses behaviour that works on `main` today: `['topics' => 'php']` for `string ...$topics` currently yields `['php']`.

It matters for prompts. `prompts/get` arguments are `Record<string, string>` by protocol — the Inspector sends `userIds=["101","102"]` through as a *literal string* — so a variadic prompt parameter fails here where it works now.

Wrapping a non-array in a one-element list instead of throwing keeps both cases working.

src/Capability/Registry/ReferenceHandler.php
```diff
@@ -99,10 +99,33 @@ private function prepareArguments(\ReflectionFunctionAbstract $reflection, array
...
+                    throw RegistryException::invalidParams(\sprintf('Parameter `%s` must be an array of values.', $paramName));
+                }
+                foreach (array_values($values) as $value) {
```
chr-hertel: `array_values()` silently accepts a JSON object and flattens it into positional args, even though the advertised schema is `array`. `array_is_list()` plus a reject would be tighter.

src/Capability/Registry/ReferenceHandler.php
```diff
@@ -99,10 +99,33 @@ private function prepareArguments(\ReflectionFunctionAbstract $reflection, array
...
+                    try {
+                        $finalArgs[] = $this->castArgumentType($value, $parameter);
+                    } catch (InvalidArgumentException $e) {
+                        throw RegistryException::invalidParams($e->getMessage(), $e);
```
chr-hertel: Drops which element failed. `Cannot cast value to integer` on a 20-item array is hard to act on — worth including the parameter name and the index.

chr-hertel (commented):
Thanks for tracking this down, @Faneraiy14 — the diagnosis is right, including the bit that trips most people up: `getType()` on `int ...$scores` returns `int`, not `array`, so `castArgumentType()` genuinely needs no changes. The ordering argument holds as well.

Four notes inline. The one on the non-array rejection is a real regression, the rest are polish.

One more thing that can't be anchored to a line: please prefix the title with the component to match the repo convention — `[Capability] Cast variadic tool parameters element-by-element`.

Nothing in the repo exercises variadics end-to-end right now, which is exactly why the prompt case below stays invisible. I'll follow up with an example and docs once this lands.

Faneraiy14:
good catch on the prompt-argument case, missed that entirely.

fixed all 4:
- non-array now wraps as single-element instead of throwing (the regression)
- non-list array (object) gets rejected explicitly instead of array_values() silently flattening it
- cast error now says which index failed (`scores[1]: ...`)
- trimmed the rationale comment down

tests updated to match - replaced the old "throws on non-array" test since that's no longer the behavior, added tests for the wrap case, the object-rejection case, and the indexed error message. 1518 unit tests + phpstan level 6 still green.

renamed the title too.

chr-hertel (approved):
Thanks @Faneraiy14 - just added a small additional test, but will merge in a second 👍

PR #477, opened 2026-08-23
[Server] Keep one unregistrable element from taking the whole registry down

### LINKED ISSUE #476: One placeholder-less resource template makes the whole server unserviceable
Reported by Mariano Damian Ferro Villanueva via email, opening it here on their behalf.

## Problem

A `#[McpResourceTemplate]` whose `uriTemplate` contains no placeholder takes the whole server down, not just that one template.

```php
#[McpResourceTemplate(
    uriTemplate: 'data://tags',
    name: 'all_tags',
    title: 'All Tags',
    description: 'All Tags',
    mimeType: 'application/json'
)]
public function tag_all(?int $paged = 1): array
{
    // ...
}
```

The handshake still succeeds, and then every request fails, including ones that have nothing to do with resources:

```
tools/list -> {"jsonrpc":"2.0","id":2,"error":{"code":-32602,"message":"Error registering manual resource template 'data://tags': Invalid URI template : \"data://tags\" must be a valid URI template with at least one placeholder."}}
tools/call -> {"jsonrpc":"2.0","id":3,"error":{"code":-32602,"message":"Error registering manual resource template 'data://tags': Invalid URI template : \"data://tags\" must be a valid URI template with at least one placeholder."}}
```

Reproduced on `main` against `Server::builder()` with the Streamable HTTP transport, on both the handshake and the 2026-07-28 lifecycle.

## Cause

1. `ResourceTemplate::__construct()` requires at least one placeholder and throws ([src/Schema/ResourceTemplate.php#L59-L61](https://github.com/modelcontextprotocol/php-sdk/blob/main/src/Schema/ResourceTemplate.php#L59-L61)).
2. `ReflectedElementLoader` wraps that into a `ConfigurationException` ([ReflectedElementLoader.php#L214-L219](https://github.com/modelcontextprotocol/php-sdk/blob/main/src/Capability/Registry/Loader/ReflectedElementLoader.php#L214-L219)), which aborts `Registry::load()` before any element is registered.
3. `Builder::$lazyLoading` defaults to `true`, so that load runs on the first read during request handling. `Registry::load()` sets `loaded` only on success, so every subsequent request retries and fails the same way.

So one misconfigured element is enough to make a server serve nothing, and the client is told `-32602` (Invalid params) for what is a server-side configuration error it cannot do anything about.

The original report saw it as a 500 with `Cannot modify header information - headers already sent (output started at /vendor/symfony/http-foundation/Response.php:393)`, which is how it surfaces once the response is already being written. I could not reproduce that part on `main`, so treat it as a symptom of the surrounding stack rather than part of this issue.

## Secondary issue: inconsistent handling

The same mistake behaves differently depending on the registration path:

- `ReflectedElementLoader` throws `ConfigurationException`, a hard failure taking down the registry.
- `Discoverer::processFile()` catches `\Throwable`, logs it and continues, so an attribute-discovered template with the same mistake is silently dropped.

## Suggested fix

Validate the URI template in `Builder::addResourceTemplate()`, where the developer wrote it, instead of at the first read of the registry. PR follows.

Worth considering separately: a `ConfigurationException` reaching a client as `-32602` is misleading (it is caught by `catch (\InvalidArgumentException)` in `Protocol`), and a single failing element aborting the whole registry load is a large blast radius for any other configuration mistake.

Line references are against `main`.

wachterjohannes:
Fixes #476.

A placeholder-less resource template made the server serve nothing. `ReflectedElementLoader` wrapped the failure in a `ConfigurationException` and rethrew it, which aborted `Registry::load()` before any element was registered. Loading is lazy by default, so that ran while a request was being served: `tools/list` and `tools/call` answered `-32602` with the template's message, and since `load()` marks itself loaded only on success, every following request repeated it.

The loader now logs the failure and skips that element, the way `Discoverer` already treats an attribute it cannot process (`Discoverer.php:300-310`). The bad template is the only thing missing; the rest of the registry stays servable, and a server-side configuration error stops being reported to the client as invalid params on unrelated calls. All four element types carried the same rethrow, so all four change.

The `ResourceTemplate` message now also says what to do about it: a URI without a placeholder addresses a single resource, so it belongs in `addResource()`.

### Changed from the first revision

The first take validated the URI template in `Builder::addResourceTemplate()`. @chr-hertel [pushed back on that](https://github.com/modelcontextprotocol/php-sdk/pull/477#discussion_r3839852554), rightly: every schema invariant would end up mirrored in the builder alongside the value object that already owns it, and the `ResourceTemplate::isValidUriTemplate()` accessor existed only to make that mirroring possible. Both are gone. Nothing in `Builder` changed, and this revision fixes the reported path — the `#[McpResourceTemplate]` blast radius — rather than one doorway into it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 * Add `annotations` to `ImageContent`.
 * Fix empty tool/resource schemas serializing as `[]` instead of `{}`.
 * Fix `PromptResultFormatter` dropping `annotations`, `_meta` and `mimeType` for plain-array content.
+* Fix one unregistrable element aborting the whole registry load: `ReflectedElementLoader` now logs and skips it instead of rethrowing a `ConfigurationException`, which under lazy loading answered every request with the failed element's message.
 
 0.7.0
 -----
--- a/docs/servers/registration.md
+++ b/docs/servers/registration.md
@@ -136,7 +136,9 @@ $server = Server::builder()
 #### Parameters
 
 - `handler` (callable|string): The resource template handler
-- `uriTemplate` (string): The resource URI template
+- `uriTemplate` (string): The resource URI template. It must carry at least one `{placeholder}`, which is what lets it
+  address more than one resource. A URI without one is a single resource: register it with `addResource()` instead.
+  A template that does not qualify is logged and skipped when the registry loads; the rest of the server keeps working.
 - `name` (string|null): Optional resource template name
 - `title` (string|null): Optional human-readable title for display in UI
 - `description` (string|null): Optional resource template description
--- a/src/Capability/Registry/Loader/ReflectedElementLoader.php
+++ b/src/Capability/Registry/Loader/ReflectedElementLoader.php
@@ -21,7 +21,6 @@ use Mcp\Capability\Discovery\SchemaGenerator;
 use Mcp\Capability\Discovery\SchemaGeneratorInterface;
 use Mcp\Capability\Registry\ElementReference;
 use Mcp\Capability\RegistryInterface;
-use Mcp\Exception\ConfigurationException;
 use Mcp\Schema\Annotations;
 use Mcp\Schema\Icon;
 use Mcp\Schema\Prompt;
@@ -35,6 +34,11 @@ use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 
 /**
+ * An element that cannot be registered is logged and skipped, never rethrown: loading is lazy by
+ * default and therefore runs while a request is being served, where aborting takes every other
+ * element down with it and answers unrelated calls with the failed element's message. This matches
+ * how `Discoverer` already treats a bad attribute.
+ *
  * @author Antoine Bluchet <soyuka@gmail.com>
  *
  * @phpstan-import-type Handler from ElementReference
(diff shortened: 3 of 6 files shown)


src/Server/Builder.php
```diff
@@ -743,6 +749,10 @@ public function addResourceTemplate(
         ?Annotations $annotations = null,
         ?array $meta = null,
     ): self {
+        if (!ResourceTemplate::isValidUriTemplate($uriTemplate)) {
+            throw new InvalidArgumentException(\sprintf('Invalid URI template "%s" for resource template handler %s: a template needs a scheme and at least one placeholder, e.g. "user://{userId}/profile". Use addResource() for a URI that addresses a single resource.', $uriTemplate, $this->describeHandler($handler)));
+        }
```
chr-hertel: bringing the validation logic into `Builder` is not a real option to me. even tho you tried to mitigate with that new `isValidUriTemplate` method, this still is a bit leaky and wouldn't scale nice - think of bringing all those validation+exception paths to the `Builder` ... 😬

wachterjohannes: Agreed, and dropped — `Builder` is untouched now, and `ResourceTemplate::isValidUriTemplate()` is gone with it.

Chasing your point further: the validation never belonged at registration at all, because the reported case is `#[McpResourceTemplate]`, which never goes through `addResourceTemplate()`. The real defect is that `ReflectedElementLoader` rethrows a failed element as a `ConfigurationException` and takes `Registry::load()` down with it, before anything is registered — under lazy loading that happens mid-request, so one bad template answers `tools/list` and `tools/call` with the template's message on every request.

`Discoverer` already has the answer (`Discoverer.php:300-310`): catch per element, log, move on. The loader now does the same for all four element types. One bad element costs you that element.

PR #490, opened 2026-08-29
[Server][Capability] Do not announce an externally loaded registry as changed

chr-hertel:
`Registry` suppresses its `*ListChangedEvent`s while loading, but the `loading` guard is only set by `Registry::load()`. `Builder::resolve()` loads a caller-supplied registry by calling `$chainLoader->load($registry)` directly, which bypasses it — so with a notification bus configured, every `build()` publishes one `list_changed` per element for a registry that did not change.

```php
$registry = new Registry(new PublishingEventDispatcher($bus));
$tools(Server::builder()->setRegistry($registry)->setNotificationBus($bus))->build();
echo $bus->cursor();   // 3, and 0 for a registry the Builder constructs itself
```

`InMemoryNotificationBus` mostly gets away with it. `Psr16NotificationBus` does not: under PHP-FPM the server is built per request and the bus is shared, so every request broadcasts its whole element list to every open `subscriptions/listen` stream and consumes the 256-entry backlog — after which a lagging reader silently skips real notifications.

Splits the guarded body of `load()` into `loadFrom(LoaderInterface $loader)` and routes the custom-registry branch through it.

`loadFrom()` takes only the `loading` guard; `loaded` stays with `load()`. The loader it runs is the caller's, so marking the registry loaded there would retire a constructor loader that never ran, and make a second `loadFrom()` — one registry handed to two builders — a silent no-op.

`RegistryInterface` declares no `load()`, so this stays on the concrete `Registry`; anything else keeps the path it has today. No BC break.

Found in `symfony/mcp-bundle`, which always supplies its own registry (symfony/ai#2458).

### DIFF
--- a/src/Capability/Registry.php
+++ b/src/Capability/Registry.php
@@ -90,7 +90,6 @@ final class Registry implements RegistryInterface
 
         $this->loadFrom($this->loader);
 
-        // Only on success: a failure propagates, so it is retried on the next read.
         $this->loaded = true;
     }
 
--- a/src/Capability/Registry.php
+++ b/src/Capability/Registry.php
@@ -88,10 +88,28 @@ final class Registry implements RegistryInterface
             return;
         }
 
+        $this->loadFrom($this->loader);
+
+        // Only on success: a failure propagates, so it is retried on the next read.
+        $this->loaded = true;
+    }
+
+    /**
+     * Runs $loader with the change events its registrations would dispatch suppressed, since they
+     * describe the registry filling up rather than changing.
+     *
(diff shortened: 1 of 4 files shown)


src/Capability/Registry.php
```diff
@@ -88,10 +88,28 @@ public function load(): void
             return;
         }
 
+        $this->loadFrom($this->loader);
+
+        // Only on success: a failure propagates, so it is retried on the next read.
```
chr-hertel:
```suggestion
```

PR #497, opened 2026-09-03
[Client] Expose session ID from HttpTransport

### LINKED ISSUE #496: [Client] Expose the Mcp-Session-Id from HttpTransport for cross-request session resumption
Getting the session ID can be helpful to allow resuming previous sessions.

`HttpTransport` captures the `Mcp-Session-Id` response header into a private property and there is no public accessor on the transport or on `Client`:

https://github.com/modelcontextprotocol/php-sdk/blob/v0.8.1/src/Client/Transport/HttpTransport.php#L41

While the BaseTransport has a `protected $sessionId`, the HttpTransport rescopes it to `private`.

### Proposal

Add a public getter, e.g.:

```php
public function getSessionId(): ?string;
```

on `HttpTransport` (optionally surfaced via `Client`), so callers can persist the ID after a request and feed it back via the existing `$headers` mechanism in a later PHP request.

RobLoach:
Adds `HttpTransport::getSessionId()` so request-scoped callers can persist the server-minted session ID and re-inject it later via the constructor's `$headers` parameter. Fixes #496.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to `mcp/sdk` will be documented in this file.
 
+0.9.0
+-----
+
+* Add `HttpTransport::getSessionId()` to read the server-minted `Mcp-Session-Id`: a request-scoped caller can persist it and pass it back through the constructor's `$headers` on a later transport. Always `null` on `2026-07-28`, which removed protocol-level sessions.
+
 0.8.0
 -----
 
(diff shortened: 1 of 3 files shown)


chr-hertel (dismissed):
Fair with that disclaimer in the comment - won't be there for long with modern era discontinuing the session, but makes sense for now.

Thanks @RobLoach!

PR #499, opened 2026-09-05
[Capability] Remove CompletionProvider's dead providerClass argument

chr-hertel:
`#[CompletionProvider(providerClass: Foo::class)]` has never worked in any release. The constructor's exactly-one-source check was `array_filter([$provider, $values, $enum])` — `$providerClass` was never counted, so the form always threw `Only one of provider, values, or enum can be set.` during `newInstance()`. Discovery catches that and drops the element, so the whole prompt/template silently disappeared. `Discoverer::getCompletionProviders()` then stored `$attributeInstance->provider` in the `providerClass` branch, always `null` there.

So rather than making it work: `provider:` already accepts a `class-string<ProviderInterface>` and resolves it through the container identically, which makes `providerClass` a redundant spelling that nothing can depend on — every version that shipped it threw on construction. Removing it also unlocks the positional form `#[CompletionProvider(Foo::class)]`, which until now hit `providerClass` and threw.

[BC Break] in the changelog, though the practical risk is nil: no working code could have set the property.

Covered from both ends. `DiscoveryTest` pins the positional class-string surviving discovery unresolved, and a new integration test drives a real client against a real server over stdio — the fixture's provider takes a scalar the auto-wiring container cannot supply, so it can only come from `Container::set()`, and completions arriving at all prove the container was asked to build it.

Found by Copilot's review on #492.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 0.9.0
 -----
 
+* [BC Break] Remove the `providerClass` argument of `#[CompletionProvider]`: constructing it always threw, so it never worked. Use `provider:`, which takes the same class-string and is now the first positional argument.
 * Add `HttpTransport::getSessionId()` to read the server-minted `Mcp-Session-Id`: a request-scoped caller can persist it and pass it back through the constructor's `$headers` on a later transport. Always `null` on `2026-07-28`, which removed protocol-level sessions.
 
 0.8.0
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to `mcp/sdk` will be documented in this file.
 0.9.0
 -----
 
+* [BC Break] Remove the `providerClass` argument of `#[CompletionProvider]`: constructing it always threw, so it never worked. Use `provider:`, which takes the same class-string and is now the first positional argument.
```
```suggestion
* [BC Break] Remove the `providerClass` argument of `#[CompletionProvider]`. Use `provider:`, which takes the same class-string and is now the first positional argument.
```

PR #506, opened 2026-09-12
[Server] fix OIDC discovery for issuers with trailing slash

karedum:
Preserve the original OIDC issuer for issuer validation and use a separate value without the trailing slash when constructing discovery URLs.

## Motivation and Context
OidcDiscovery::fetchMetadata() currently removes the trailing slash from the configured issuer before performing discovery. The normalized value is also used for issuer validation.

This causes discovery to fail when the authorization server uses a trailing slash as part of its canonical issuer, such as Authentik.

OpenID Connect Discovery requires a terminating slash to be removed when constructing the discovery URL, while issuer validation must preserve the issuer identifier.

## How Has This Been Tested?
Tested against an Authentik OAuth2/OpenID Connect provider whose issuer contains a trailing slash.

The existing test suite also passes locally.

## Breaking Changes
None expected. Discovery URLs are constructed as before; the change only preserves the originally configured issuer for exact issuer validation.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/Server/Transport/Http/OAuth/OidcDiscovery.php
+++ b/src/Server/Transport/Http/OAuth/OidcDiscovery.php
@@ -146,8 +146,8 @@ class OidcDiscovery implements OidcDiscoveryInterface
      */
     private function fetchMetadata(string $issuer): array
     {
-        $issuer = rtrim($issuer, '/');
-        $parsed = parse_url($issuer);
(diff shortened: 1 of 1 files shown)


chr-hertel (approved):
Thanks @karedum!

Issue #275, opened 2026-04-02
[Streamable HTTP][Server] Concurrent requests in the same session can overwrite queued responses and cause stale/unknown message IDs


tebaly:
**Describe the bug**

Concurrent HTTP requests within the same MCP session can corrupt session-backed protocol state in the PHP SDK.

In `Streamable HTTP` mode, the server stores MCP protocol state (including outgoing messages and pending requests) inside a shared session payload. The SDK mutates that payload using a `read -> modify -> write whole session` pattern without any locking or atomic merge semantics.

When multiple requests are processed concurrently for the same `Mcp-Session-Id` (for example, several parallel `tools/call` requests from Cursor IDE), one request can overwrite session changes made by another request. In practice, this appears to cause lost outgoing responses / queue entries and leads the client to report stale or unknown message IDs, while the server has actually already produced the response.

This looks like an SDK-level concurrency bug rather than an application bug.

**To Reproduce**

1. Run an MCP server using the PHP SDK over `Streamable HTTP`.
2. Configure it with a shared session store (for example PSR-16 cache via Symfony MCP Bundle, but the same read/modify/write pattern appears to affect file storage too).
3. Use a client that sends multiple requests concurrently within the same MCP session.
4. Trigger 2-3 parallel `tools/call` requests (for example multiple `get_record`-style read calls).
5. Observe that some calls succeed, while another response is effectively lost from the session-backed outgoing queue.
6. On the client side, this can surface as:
   - hanging tool calls,
   - `Received a response for an unknown message ID`,
   - stale responses after reconnect.

**Expected behavior**

Concurrent requests for the same MCP session should not overwrite each other's protocol state.

At minimum, the SDK should guarantee safe mutation of session-backed MCP state (`outgoing_queue`, pending request/response maps, counters, etc.) when multiple HTTP requests for the same session are processed in parallel.

Possible valid fixes could include:
- locking per MCP session during request processing,
- atomic session mutation,
- splitting queue state out of the monolithic session blob,
- or another concurrency-safe design.

**Logs**

Client-side symptoms observed in Cursor IDE:

```text
Ignoring stale response (unknown message ID): Received a response for an unknown message ID: {"jsonrpc":"2.0","id":10,"result":{...}}
```

We also observed cases where multiple parallel `get_record` calls were started, two completed successfully, and one response payload appeared only as a stale/unknown-message-id response on the client side.

Relevant SDK code showing the read/modify/write pattern:

[...]

PR #508, opened 2026-09-14
[Server] Answer a JSON POST with only its own responses

### LINKED ISSUE #467: StreamableHttpTransport returns a JSON array in the response body when parallel POST requests arrive concurrently (PHP-FPM)
### Summary

  When a client sends multiple tool calls as parallel POST requests to a Streamable HTTP MCP server running under PHP-FPM,
  StreamableHttpTransport::createJsonResponse() returns a JSON array of all queued responses in a single HTTP response body. This violates the MCP spec and causes spec-compliant clients to fail with a ValidationError.

  ### Environment

  - mcp/sdk: v0.6.0 (latest)
  - symfony/mcp-bundle: v0.10.0
  - PHP: 8.4, running under PHP-FPM (multiple concurrent worker processes)
  - MCP client: Python mcp SDK (mcp/client/streamable_http.py), protocol version 2025-03-26

###  Steps to reproduce

  1. Start a Symfony MCP server using StreamableHttpTransport under PHP-FPM
  2. Connect an MCP client and initialise a session
  3. Have the client dispatch multiple (8 tools in my case) parallel tool calls in a single agent turn
  4. Observe the server response

  ### What happens

  PHP-FPM spins up 8 worker processes, one per request. All 8 workers process their requests concurrently and each calls queueOutgoing(), which appends to the shared cache-backed session queue. There is then a race on consumeOutgoingMessages():

```php
  // Protocol.php
  public function consumeOutgoingMessages(Uuid $sessionId): array
  {
      $session = $this->sessionManager->createWithId($sessionId);
      $queue   = $session->get(self::SESSION_OUTGOING_QUEUE, []);
      $session->set(self::SESSION_OUTGOING_QUEUE, []);  // atomically clears the entire queue
      $session->save();
      return $queue;
  }
```
  Whichever worker wins the race reads all 8 responses from the queue. That response then hits this branch in StreamableHttpTransport:
```php
  // StreamableHttpTransport.php, line 151
  $responseBody = 1 === \count($messages)
      ? $messages[0]
      : '['.implode(',', $messages).']';  // ← returns a JSON array
```
  One HTTP response body becomes:
```json
  [                                                                                                                                                             {"jsonrpc":"2.0","id":2,"result":{...}},
    {"jsonrpc":"2.0","id":3,"result":{...}},
    {"jsonrpc":"2.0","id":4,"result":{...}},                                                                                                                    ...7 more entries...
  ]
```
  The remaining 7 workers drain an empty queue and return 202 No Content.

  The Python MCP client then calls:
```
  # mcp/client/streamable_http.py, line 385
  message = JSONRPCMessage.model_validate_json(content)
```

  and crashes with:

```
  pydantic_core._pydantic_core.ValidationError: 4 validation errors for JSONRPCMessage
  JSONRPCRequest
    Input should be an object [type=model_type, input_value=[{'jsonrpc': '2.0', 'id': ...}], input_type=list]
  JSONRPCNotification
    Input should be an object [type=model_type, ...input_type=list]
  JSONRPCResponse
    Input should be an object [type=model_type, ...input_type=list]
  JSONRPCError
    Input should be an object [type=model_type, ...input_type=list]
```

  The 7 requests that got 202 No Content receive no response at all, causing the agent to time out waiting for results.

  ### What the MCP spec requires

  The [MCP Streamable HTTP spec ](https://spec.modelcontextprotocol.io/specification/basic/transports/#streamable-http)is explicit on this point:

  ▎ "If the server responds directly in the HTTP response body, it MUST use Content-Type: application/json and the body MUST be a single JSON-RPC message
  ▎ object."

  For parallel tool calls the spec defines two compliant patterns:

  Option A — each POST returns its own single response synchronously:
  Client                        Server
```
    |--- POST /mcp (id=2) ---->|---> 200 {"jsonrpc":"2.0","id":2,"result":{...}}
    |--- POST /mcp (id=3) ---->|---> 200 {"jsonrpc":"2.0","id":3,"result":{...}}
    |--- POST /mcp (id=4) ---->|---> 200 {"jsonrpc":"2.0","id":4,"result":{...}}
```

  Option B — POSTs return 202, responses delivered over SSE GET stream:
  Client                        Server
```
    |--- GET /mcp (SSE) ------>|   (persistent stream)
    |--- POST /mcp (id=2) ---->|---> 202 Accepted
    |--- POST /mcp (id=3) ---->|---> 202 Accepted
    |--- POST /mcp (id=4) ---->|---> 202 Accepted
    |<-- SSE data: id=2 -------|
    |<-- SSE data: id=3 -------|
    |<-- SSE data: id=4 -------|
```

  Returning a JSON array is not a valid option under any revision of the spec.

 ###  Root cause

  The design of the shared session outgoing queue assumes a single-process, sequential request model (PHP CLI / built-in server). Under PHP-FPM, multiple worker processes share the same session cache and race to drain the queue, so a worker can inadvertently collect and return responses that belong to other concurrent requests.

  The array branch on line 151 of StreamableHttpTransport was written as a defensive fallback for a scenario that should never happen in single-process mode, but it surfaces as a reliable bug under any multi-process deployment.

  ### Suggested fix

  createJsonResponse() should never return more than one response. If the queue happens to contain multiple messages (due to a race), only the message matching the current request's ID should be returned; the rest should remain in the queue for the SSE GET stream to deliver.

  A minimal change that prevents the spec violation:

```php
  // StreamableHttpTransport.php
  protected function createJsonResponse(): ResponseInterface
  {
      $outgoingMessages = $this->getOutgoingMessages($this->sessionId);

      if (empty($outgoingMessages)) {
          return $this->responseFactory->createResponse(202)
              ->withHeader('Content-Type', 'application/json');
      }

      $messages = array_column($outgoingMessages, 'message');

      // NEVER return an array — the MCP spec requires a single JSON-RPC object
      // per HTTP response. Return only the first message; others will be delivered
      // via the SSE GET stream.
      $responseBody = $messages[0];

      // ...
  }
```

  A more complete fix would make the queue per-request-ID rather than a shared FIFO, so each worker only ever sees its own response.

vbcherepanov:
Fixes #467.

With PHP-FPM, requests of the same session run in parallel and put their responses into the same session queue. `createJsonResponse()` took the whole queue, so one request got all responses as a JSON array and the others got `202` with nothing.

Now the transport collects the ids from its own POST body and takes only those responses. The rest stays in the queue. Batches are still answered with an array, single messages with one object, and an invalid message without a usable id still gets its error.

`Protocol::consumeOutgoingMessages()` gets the ids as an optional second argument. Without it nothing changes, so the SSE path and `StdioTransport` work as before.

The reproducer is `testConcurrentPostsSharingASessionEachReceiveTheirOwnResponse`: the second POST is handled completely between the first one queueing its response and reading the queue. It fails on `main`.

Not part of this PR:
- the lost update on the session data itself (#275). The queue is read and written like before, and nothing is written if there is nothing to take.
- notifications queued outside of an SSE stream (e.g. resource subscriptions) are not added to a JSON response anymore, since they could only go there as an invalid array. They stay in the queue.


michalcharvat:
Independent confirmation on v0.6.0 (Apache + mod_php), driving a real MCP server rather than a unit test: 20 concurrent `tools/list` + `resources/list` pairs, each on a freshly initialized session, both requests released from a thread barrier.

| build | pairs | correct | cross-delivery | response lost (empty `202`) | HTTP 500 |
| --- | --- | --- | --- | --- | --- |
| v0.6.0 | 40 | 27 | 11 | 2 | 0 |
| v0.6.0 + this PR | 80 | 70 | 0 | 6 | 4 |
| v0.6.0 + this PR + #500 | 60 | 59 | 0 | 1 | 0 |

This PR removes the cross-delivery completely — 0 in 140 pairs after it, against 11 in 40 before. Both shapes described in #467 are gone: the JSON array carrying another request's response, and the `202` the other request got instead of its own.

Two notes from the same runs:

1. The HTTP 500s that surface once cross-delivery is gone are `JsonException: Control character error` from `Session::readData()` on a half-written session file — #498, fixed by #500. With both PRs applied they disappear. This PR does not cause them, it just stops masking them.
2. One lost response survives both (1 in 60 pairs), which is the lost update of #275 that this PR correctly says it does not address.

Real-client impact, for the record: Claude Code issues `tools/list` and `resources/list` concurrently on every connect, so a plain connection hits this. On a hit it cancels the orphaned request after 30 s with `notifications/cancelled … "Request timed out"` and silently ends up with no resources.

The patch was applied by hand onto v0.6.0 (the files have moved since), so line numbers differ but the logic is unchanged.

michalcharvat:
**Correction to my numbers above, and a data point on v0.8.1.**

The residual-loss figure in my comment — "one lost response survives both (1 in 60 pairs)" — came from too small a sample. Larger runs put it at **6 in 180 pairs** on v0.6.0 with both PRs applied, so roughly **1 in 30**. That residual is still the lost update of https://github.com/modelcontextprotocol/php-sdk/issues/275 and still outside this PR's scope, so nothing here changes; I am correcting it because "1 in 60" understates how often a real client ends up waiting on a response that was silently dropped.

The cross-delivery result did not move, and it is the one I would stand behind: **0 in 180 pairs**.

**v0.8.1 still has the bug.** Same harness against stock v0.8.1: 65 % of pairs correct, versus 67 % on stock v0.6.0. So the v0.8.0 refactor did not incidentally fix this and the PR is still needed against current releases — worth saying explicitly, because the files have moved enough that four hunks had to be finished by hand when porting it there.

| base | pairs | correct | cross-delivery | response lost |
|---|---|---|---|---|
| stock v0.6.0 | 40 | 67 % | 11 | 2 |
| v0.6.0 + 508 + 500 | 180 | 97 % | 0 | 6 |
| stock v0.8.1 | 60 | 65 % | many | — |
| v0.8.1 + 508 + 500 | 180 | 91 % | 0 | 17 |

One caveat on that 17, since it would be easy to misread as "v0.8.1 loses more than v0.6.0": those runs shared a busy machine and a single run contributed 5 of them. I would not claim a difference between the two bases without re-measuring on an idle box. The zero cross-delivery is the robust part of both patched rows.

Same setup as before — Apache + mod_php, 20 concurrent `tools/list` + `resources/list` pairs per run, each pair on a freshly initialized session, both requests released from a thread barrier.

